PR #32, opened 2021-03-01
chore: assure release workflow is triggered

ssbarnea:


joshuawilson (approved):

joshuawilson:
This looks like it should work, but we won't know until you try it. 

You might try testing in a fork.

PR #53, opened 2021-03-26
chore: fix pre-commit eslint dependencies

ssbarnea:


package.json
```diff
@@ -171,6 +171,7 @@
     "@commitlint/cli": "latest",
     "@commitlint/config-conventional": "latest",
     "husky": "latest",
+    "typescript-eslint": "^0.0.1-alpha.0",
```
joshuawilson, line 174: Why is this added to the Dependencies?

ssbarnea: That is indeed a mistake, I am still trying to make pre-commit.ci happy....

PR #56, opened 2021-04-04
docs: added a hint on how to install `ansible-lint` to README.md

stopendy:
## Updates
- Added some tips regarding how to install anisble-lint in venv
- No feature update

## Comments
It's my first pull request to GitHub ever.  
Any kinds of advice on GitHub, Pull Request, and even Markdown and English grammar is thankfully taken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,10 +17,10 @@ This extension also installs YAML extension in order to enable schema verificati
 
 ## Known Issues
 ### Depends on ansible-lint
-For the  moment you need to install `ansible-lint` yourself.  
+For the  moment you need to install `ansible-lint` yourself.
 If you don't want to mess up the system-wide python packages list, we have a hint for you.
 
-First, make your own Python virtual environment to install `ansible-lint`.  
+First, make your own Python virtual environment to install `ansible-lint`.
 For example, execute the following command to make a virtual environment called "system".
 
 ```sh
@@ -34,12 +34,12 @@ $ source ~/venv/system
 (system)$ pip install ansible-lint
 ```
 
-Open VS code, and select `File > Preferences > Settings`.  
-Search for "ansible.validate.executablePath" in the search bar on top.  
+Open VS code, and select `File > Preferences > Settings`.
+Search for "ansible.validate.executablePath" in the search bar on top.
 Click on "Edit in settings.json" to open `settings.json` for users.
 
-Then add the following line to register the `ansible-lint` execution path in VS Code (\*).  
-This setting is necessary because `ansible-lint` is not in regular PATH environment variable.  
+Then add the following line to register the `ansible-lint` execution path in VS Code (\*).
+This setting is necessary because `ansible-lint` is not in regular PATH environment variable.
 (\*) The path to `ansible-lint` must be changed depending on your environment.
 
 ```json
--- a/README.md
+++ b/README.md
@@ -16,8 +16,35 @@ easier to write Ansible playbooks, roles, collections, modules and plugins.
 This extension also installs YAML extension in order to enable schema verification and code-completion.
 
 ## Known Issues
+### Depends on ansible-lint
+For the  moment you need to install `ansible-lint` yourself.  
+If you don't want to mess up the system-wide python packages list, we have a hint for you.
 
-* For the  moment you need to install `ansible-lint` yourself
+First, make your own Python virtual environment to install `ansible-lint`.  
+For example, execute the following command to make a virtual environment called "system".
+
+```sh
+$ python3 -m venv ~/venv/system
+```
+
+Get into the "system" virtual environment, and install `ansible-lint`.
+
+```sh
+$ source ~/venv/system
+(system)$ pip install ansible-lint
+```
+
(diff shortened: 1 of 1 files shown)


joshuawilson:
Will these directions work on all 3 major OS; Windows, Mac, and Linux?

stopendy:
@joshuawilson 
Thank you for confirmation. Theoretically this procedure should work on 3 major platform. I’ve already confirmed on Fedora 33 + Python 3.9. 

I’ll try to set up Windows and Mac to test if the instruction also works on them by Sunday 11, April.

stopendy:
@joshuawilson 
Thanks to your advice, I noticed that the path expression `~/venv` and `source` command won't work on Windows. `ansible-lint` might not be able to be installed on Windows so WSL-like trick could be necessary. Also, older version of Mac might not have Python 3 installed, so additional information might be necessary. Plus, I also noticed that this procedure was not working even on Linux. I should have tested `ansible-lint` feature more. I am sorry for spending your time.

Now I guess this pull request might as well not be merged, and closed immediately. If the workaround written below would be a good solution for someone, I am happy to keep testing on Windows and Mac, and write the instruction on the Discussion of this repository, not in README.md.

The detail follows.

## Why It's Not Working
By configuring the absolute path to `ansible-lint` in VS Code, Ansible Language extension detects `ansible-lint` executable, and no warning is displayed on VS Code. That's how I thought it's working.

However when I manually execute `/home/user-name/venv/system/ansible-lint playbook.yml`, then the following error was issued, even after I installed `ansible-base` in the `system` venv.

```
FileNotFoundError: [Errno 2] No such file or directory: 'ansible'
```

Finally executing `ansible-lint` outside the venv seems not supported. `ansible-lint` is successfully executed under venv, but `ansible` executable is called outside the venv, so VS Code claimed that `ansible` was missing.

## Workaround
When I launched VS Code under `system` venv as below, `ansible-lint` with Ansible Language Extension successfully worked. This could be because the environment variable `PATH` is exported to VS Code process, and the child process `ansible-lint`, and the grandchild process `ansible` could be executed under `system` venv.

In this case, no modification of `ansible.validate.executablePath` on VS Code is needed, because all the program runs in `system` venv. However I don't think running VS Code under venv every time is not elegant. This instruction is not be worth written on README.md.

The workaround procedure follows.

### Setup

```sh
$ python3 -m venv ~/venv/system
$ source ~/venv/system/bin/activate
(system)$ pip install ansible-lint ansible-base
(system)$ deactivate
```

### Launch VS Code Under Venv

```sh
source ~/venv/system/bin/activate; code; deactivate
```

joshuawilson:
I'm glad I could get you to think about it. I'm not saying you should close it. It is up to you if you want to put more time into it.

stopendy:
Thank you for your thoughtful comment. I considered how much it would affect to users if I put more time into it, and decided to close this pull request.

This trick solves only corner case for a limited time until `ansible-lint` is integrated to the extension. Also merging this kind of topic into README might be a bloat. I guess I'd better make an effort to contribute in another way.

stopendy:
I came up with 2 other approaches. Workaround 2 might not work on Windows because it makes use of symbolic links. Workaround 3 might technically work, but there are a variety of how-to depending on the operating system and desktop environments.

The combination of Workaround 1 and Workaround 2 might be worth looking, but I'll have to confirm if there is Workaround 2 would work on Windows and Mac.

## Workaround 2
\* This workaround is tested only on Linux.

Another way for VS Code to call `ansible-lint`, `ansible`, and `ansible-playbook` is to include those command in venv into PATH environment variable.

Assume that PATH is like following:

```sh
echo $PATH
/home/user-name/.local/bin:/home/user-name/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin
```

Here the first path `/home/user-name/.local/bin` is good to tweek, because this path will never affect system-wide. 

Like "Workaround 1", create a venv and install ansible-lint and ansible.

```sh
$ python3 -m venv ~/venv/system
$ source ~/venv/system/bin/activate
(system)$ pip install ansible-lint ansible
(system)$ deactivate
```

Then add 3 commands to `/home/user-name/.local/bin`.

```sh
ln -s ~/venv/system/bin/ansible-lint ~/.local/bin/ansible-lint
ln -s ~/venv/system/bin/ansible ~/.local/bin/ansible
ln -s ~/venv/system/bin/ansible-playbook ~/.local/bin/ansible-playbook
```
Now `ansible-lint`, and `ansible` and `ansible-playbook` which are internally called by `ansible-lint` are added to PATH.

Open VS Code normally and Ansible Language Extension should be able to run ansible-lint with no problem.

## Workaround 3
Set login script to add `~/venv/system/bin/` to PATH after setting up "system" venv as a replacement to creating symbolic links in Workaround 2 could be another way. However the way login script is set depends on which OS or desktop environment we use. Covering every single environment is not practical.

stopendy:
I fixed the commit locally based on workaround 2. Since I'm unable to reopen this ticket, I opend another pull request [#67](67).

PR #63, opened 2021-04-15
build: 0.3.1 release

ssbarnea:


joshuawilson:
I'm thought the version was in the package.lock too. It is set to 0.0.4.

ssbarnea:
We clearly need CI for that, I failed to update it twice. I am not sure what is its purpose but it should not be allowed to get out of sync.

PR #73, opened 2021-04-19
build: assure package lock file is updated

ssbarnea:


joshuawilson (approved):

JPinkney (approved):

.pre-commit-config.yaml
```diff
@@ -40,3 +40,11 @@ repos:
...
       files: \.[jt]sx?$  # *.js, *.jsx, *.ts and *.tsx
       types: [file]
+- repo: local
+  hooks:
+    - id: package-lock.json
+      name: package-lock.json must be in sync with package.json
+      entry: npm ci
+      always_run: true
+      language: node
+      pass_filenames: false
```
asottile, line 50: fwiw this check adds 6-10 seconds to *any* commit during the pre-commit check --  and I'm on a fast network, on a slower network it's way worse! -- pre-commit is not a good fit for this

ssbarnea: I am not aware of any pre-commit user that installs the git hook and as long it runs under 30s, the performance is acceptable for me.

Yep, i may move it outside if it performance gets annoying. Still, even if moved outside it would still need to run on CI.

If that was to be a python project I would have did a `tox -e deps` that would use pip-compile but there is no tox here.

TBW, thanks for taking the time to look at it. I am aware that many use-cases of pre-commit are bit outside of instant-like checks (mypy and pylink would be two other slow ones). That is why I try to put the slow checks at the end.

asottile: since it's already running to perform your tests, you can replace it with a simple `git diff --exit-code || ...`

PR #80, opened 2021-04-27
chore: assure commitlint runs

ssbarnea:


joshuawilson:
Why was `yarn.lock` added?

ssbarnea:
> Why was `yarn.lock` added?

That is a very good question, likely accident. Is related to the question I asked earlier on Slack, how to prevent cross npm/yarn contamination? Is there a way to ensure only one is used?

joshuawilson:
> > Why was `yarn.lock` added?
> 
> That is a very good question, likely accident. Is related to the question I asked earlier on Slack, how to prevent cross npm/yarn contamination? Is there a way to ensure only one is used?

I do not know of a way to prevent it. You could add yarn.lock to .gitignore to avoid committing it.

PR #104, opened 2021-05-13
feat: auto add custom YAML tags

ssbarnea:
Alter setttings.json and add tags used by Ansible when they are
missing.


joshuawilson (approved):

FlorianLaunay:
A bit late but work fine for me :wink:

PR #112, opened 2021-05-15
fix: lower version of vscode required

ssbarnea:
To keep extension compatible with Eclipse Theia, we lower
the version of vscode engine we require.

Related: https://github.com/eclipse-theia/theia-blueprint/issues/96


ssbarnea:
Warning: while the extension installs with this change it fails to activate because YAML extensions requires a newer vscode version.  This needs further investigation before merging.

ssbarnea:
Closing as this will not make hte extension compatible with Theia. YAML extensions requires 1.52,

PR #113, opened 2021-05-19
fix: correct line number processing

ssbarnea:
Fixes bug which prevented logging of errors on line 0.

### DIFF
--- a/src/features/validationProvider.ts
+++ b/src/features/validationProvider.ts
@@ -243,14 +243,16 @@ export default class AnsibleValidationProvider {
 			let diagnostics: DiagnosticsDictionary = {};
 			let processLine = (line: string) => {
 				let matches = line.match(AnsibleValidationProvider.matchExpression);
-				this.output.appendLine(`Found:\n${matches}`);
+				this.output.appendLine(`Found: ${matches}\n`);
 				if (matches) {
 					let message = matches.groups?.message ?? "unknown";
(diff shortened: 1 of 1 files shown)


src/features/validationProvider.ts
```diff
@@ -243,14 +243,16 @@ export default class AnsibleValidationProvider {
 			let diagnostics: DiagnosticsDictionary = {};
 			let processLine = (line: string) => {
 				let matches = line.match(AnsibleValidationProvider.matchExpression);
-				this.output.appendLine(`Found:\n${matches}`);
+				this.output.appendLine(`Found: ${matches}\n`);
 				if (matches) {
 					let message = matches.groups?.message ?? "unknown";
-					let line = parseInt(matches.groups?.line ?? "1") - 1;
+					let line = parseInt(matches.groups?.line ?? "0");
```
joshuawilson, line 249: Do you have a test for this?

ssbarnea: In fact I have no tests and this will soon come to hunt me. If one of you could add even a sample test I could start adding tests. The reason for not adding tests is because I not dot know how to write them for extensions.

Anyway, this morning I proved that there is no bug in the extension and was caused by https://github.com/ansible-community/ansible-lint/pull/1556 -- the linter will be released soon.

JPinkney: IIRC the last time I had to setup tests I followed https://code.visualstudio.com/api/working-with-extensions/testing-extension

PR #137, opened 2021-07-30
feat: add context menus and commands for running playbooks

### LINKED ISSUE #132: [vscode client] Add support for running ansible-playbook and ansible-navigator from IDE browser
To provide entry for running playbooks and drop-down option can be supported to trigger either `ansible-playbook` or `anisble-navigator` in the terminal from IDE browser window.
In the dropdown menu add `ansible-playbook` and `ansible-navigator` options and trigger the run for playbook within the terminal

webknjaz:
In particular, the commands can be invoked via:
* pallete
* file explorer context menu
* open editor context menu

The commands allow choosing between running:
* `ansible-navigator run`
* `ansible-playbook`

Resolves #132

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,7 @@
 {
   "activationEvents": [
+    "onCommand:extension.ansible-navigator.run",
+    "onCommand:extension.ansible-playbook.run",
     "onLanguage:ansible"
   ],
   "badges": [
@@ -30,6 +32,18 @@
     }
   ],
   "contributes": {
+    "commands": [
+      {
+        "category": "%commands.category.ansible-playbook%",
+        "command": "extension.ansible-playbook.run",
+        "title": "%commands.title.ansible-playbook%"
+      },
+      {
+        "category": "%commands.category.ansible-playbook%",
+        "command": "extension.ansible-navigator.run",
+        "title": "%commands.title.ansible-navigator%"
+      }
+    ],
     "configuration": {
       "type": "object",
       "title": "Ansible",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,8 @@
 {
   "activationEvents": [
     "onLanguage:yaml",
+    "onCommand:extension.ansible.navigate",
```
ssbarnea, line 4: My impression was that we would use "navigator" everywhere and not "navigate". 

@ganeshrn WDYT?

ganeshrn: Ye agreed the name should be `navigator` since the config file is for `ansible-navigator` tool

ssbarnea:
@webknjaz please rebase.

package.json
webknjaz: But since it's a command/action/function/callback, it should be a verb. The command is `ansible-navigator run` after all, with "run".

ganeshrn: `ansible.ansibleLint.path` is used for ansible-lint executable path and `ansible.ansible.path` for ansible excutable. 
I think the naming convention should follow a uniform pattern. How about `ansible.ansibleNavigator.path` and `ansible.ansiblePlaybook.path`?

ssbarnea:
For the moment lets avoid "verbalising" tool names to avoid confusing user. If we do decide otherwise we can think at better name. now "run ansible-lint" instead of "lint",...

webknjaz:
> For the moment lets avoid "verbalising" tool names to avoid confusing user. If we do decide otherwise we can think at better name. now "run ansible-lint" instead of "lint",...

Not sure what you mean. I haven't changed anything related to the lint.

package.json
webknjaz: I've changed these already. But it seems like you're confusing the commands (verbs, they are essentially actions that users invoke) and the settings (just values, like an executable path).

This specific place defines the commands.

ssbarnea (approved):

ssbarnea:
> > For the moment lets avoid "verbalising" tool names to avoid confusing user. If we do decide otherwise we can think at better name. now "run ansible-lint" instead of "lint",...
> 
> Not sure what you mean. I haven't changed anything related to the lint.

It was just an example. I looked at review and looks ok.

package.json
ganeshrn: I added the comment at wrong line, reading at correct line number

package.json
```diff
@@ -75,6 +89,24 @@
           "type": "string",
           "default": "",
           "description": "Path to the virtual environment activation script. Use only if you have a custom activation script. It will be sourced using bash before executing Ansible commands. If set, the Interpreter Path setting is ignored."
+        },
+        "ansible.navigate.executablePath": {
```
ganeshrn, line 93: `ansible.ansibleLint.path` is used for `ansible-lint` executable path and `ansible.ansible.path` for ansible excutable.
I think the naming convention should follow a uniform pattern. How about `ansible.ansibleNavigator.path` and `ansible.ansiblePlaybook.path`?

PR #142, opened 2021-07-31
feat: adopt the new language server

tomaciazek:
Replaced the code and assets with the content of tomaciazek/vscode-ansible.
Adjusted the build, run & debug to work with external language server.
Added documentation for running & debugging.


ganeshrn (approved):
@tomaciazek Nice work refactoring the client code to use [ansible-language-server](https://www.npmjs.com/package/ansible-language-server)

ganeshrn:
General Note:

Overall PR looks good from code perspective, some minor changes might be required which can be done in subsequent PRs.
Since this PR is a rewrite of client code using ansible-language-server at the backend it might break some of the existing functionality and result in some missing functionality.  
Please feel free to raise new issues on this repository so that we can gradually add the missing  features as applicable.

webknjaz:
@tomaciazek
In the future, please, construct smaller PRs so that they are atomic, less gigantic and are possible to review.
Here's a nice article on making good PRs: https://mtlynch.io/code-review-love/.

ganeshrn:
@webknjaz Since this a rewrite of the client plugin (much like an initial commit with a new direction) the PR size was big but I do agree that going forward we should be raising targeted and well scoped PR's

webknjaz:
@ganeshrn I understand why it's big. But there were a lot of changes that have nothing to do with the rewrite.

package.json
```diff
@@ -20,249 +19,228 @@
...
+    "webpack-cli": "^4.7.2"
   },
-  "displayName": "Ansible Language",
+  "displayName": "Ansible",
   "engineStrict": true,
   "engines": {
     "node": ">=16.0",
     "npm": "^7.12.1",
-    "vscode": "^1.52.0",
+    "vscode": "^1.56.0",
```
webknjaz, line 224: Was it really necessary to upgrade the minimum version of VS Code? Why?

webknjaz: cc @ssbarnea do you know anything about this?

ssbarnea: That was needed for two reasons: some required by yaml language server (so no longer needed), the second was related to file extension patterns, which were fixed my microsoft only recently, older versions not being able to use the kind of patterns we needed for identifying playbooks or other ansible files. It may now be possible to lower that number.

ChristianCiach:
Why did you remove the support for inline vault-strings? This was *the* main use-case of this extension for me.

ssbarnea:
@ChristianCiach removal of vault strings was not a desired outcome, only one of the several regressions we are aware about. We will work to bring it back.  Please create a bug for it if there is none already open.

Until we bring it back keep using the old extension, it may take few weeks.

PR #169, opened 2021-08-24
fix: restore schema verification for some files

ssbarnea:
With the exception of playbook and task files, the other ones do not have support from the language-server yet. So we restore the use of YAML schemas for these, like we had on old version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,57 @@
       "name": "Sorin Sbarnea"
     }
   ],
+  "yamlValidation": [
+    {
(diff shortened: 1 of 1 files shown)


webknjaz (approved):

JPinkney (approved):

ganeshrn (approved):
Looks good. As and when the completion support for missing schema's and file association logic is added to the language server this dependency on YAML extension can be removed from client.

PR #170, opened 2021-08-24
feat: prompt to uninstall incompatible extensions

ssbarnea:
This will detect presence of known incompatible extensions and prompt user to remove them. Implementation is based on the same feature that redhat.yaml extension has.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1,6 +1,6 @@
 /* "stdlib" */
 import * as path from 'path';
-import { ExtensionContext } from 'vscode';
+import { ExtensionContext, extensions } from 'vscode';
 
 /* third-party */
 import {
@@ -12,6 +12,7 @@ import {
 
 /* local */
 import { AnsiblePlaybookRunProvider } from './features/runner';
+import { getConflictingExtensions, showUninstallConflictsNotification } from './extensionConflicts';
 
 
 let client: LanguageClient;
@@ -49,6 +50,14 @@ export function activate(context: ExtensionContext): void {
 
   // start the client and the server
   client.start();
+
+  notifyAboutConflicts();
+  client.onReady().then(() => {
+    // If the extensions change, fire this notification again to pick up on any association changes
+    extensions.onDidChange(() => {
+      notifyAboutConflicts();
+    });
+  });
 }
 
 export function deactivate(): Thenable<void> | undefined {
@@ -56,4 +65,16 @@ export function deactivate(): Thenable<void> | undefined {
     return undefined;
   }
   return client.stop();
+}
+
+/**
+ * Finds extensions that conflict with our extension.
+ * If one or more conflicts are found then show an uninstall notification
+ * If no conflicts are found then do nothing
+ */
+function notifyAboutConflicts(): void {
+  const conflictingExtensions = getConflictingExtensions();
+  if (conflictingExtensions.length > 0) {
+    showUninstallConflictsNotification(conflictingExtensions);
+  }
 }
--- a/src/extensionConflicts.ts
+++ b/src/extensionConflicts.ts
@@ -1,0 +1,83 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Red Hat, Inc. All rights reserved..
+ *  Licensed under the MIT License. See LICENSE in the project root for details.
+ *--------------------------------------------------------------------------------------------*/
+import { commands, Extension, extensions, window } from 'vscode';
+
+// A set of VSCode extension ID's that conflict with our extension
+const conflictingIDs = [
+  'haaaad.ansible',
+  'sysninja.vscode-ansible-mod',
+  'tomaciazek.ansible',
+  'vscoss.vscode-ansible',
+  'zbr.vscode-ansible',
+];
+
+// A set of VSCode extension ID's that are currently uninstalling
+const uninstallingIDs = new Set();
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function isExtensionPresent(obj: any): obj is Extension<any> {
+  return typeof obj !== 'undefined' && !uninstallingIDs.has(obj.id);
+}
+
+/**
+ * Get all of the installed extensions that currently conflict with us
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function getConflictingExtensions(): Extension<any>[] {
+  return (
+    conflictingIDs
+      .map((x) => extensions.getExtension(x))
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      .filter<Extension<any>>((ext): ext is Extension<any> =>
+        isExtensionPresent(ext)
+      )
+  );
+}
+
+/**
+ * Display the uninstall conflicting extension notification if there are any conflicting extensions currently installed
+ */
+export async function showUninstallConflictsNotification(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  conflictingExts: Extension<any>[]
+): Promise<void> {
+  // Add all available conflicting extensions to the uninstalling IDs map
+  for (const ext of conflictingExts) {
+    uninstallingIDs.add(ext.id);
+  }
+
+  const uninstallMsg = 'Uninstall';
+
+  if (!conflictingExts.length) {
+    return;
+  }
+  // Gather all the conflicting display names
+  let conflictMsg = '';
+  if (conflictingExts.length === 1) {
+    conflictMsg = `${conflictingExts[0].packageJSON.displayName} (${conflictingExts[0].id}) extension is incompatible with redhat.ansible. Please uninstall it.`;
+  } else {
+    const extNames: string[] = conflictingExts.map(
+      (ext) => `${ext.packageJSON.displayName} (${ext.id})`
+    );
+    conflictMsg = `The ${extNames.join(
+      ', '
+    )} extensions are incompatible with redhat.ansible. Please uninstall them.`;
+  }
+
+  await window
+    .showInformationMessage(conflictMsg, uninstallMsg)
+    .then((clickedMsg) => {
+      if (clickedMsg !== uninstallMsg) {
+        return;
+      }
+      for (const ext of conflictingExts) {
+        commands.executeCommand(
+          'workbench.extensions.uninstallExtension',
+          ext.id
+        );
+        uninstallingIDs.delete(ext.id);
+      }
+    });
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.5.0 (UNRELEASED)
 
 * Update default setting value to always use FQCN (#196)
+* Detect incompatible extensions and prompt to remove them.
 
 ## 0.4.5
 


src/extensionConflicts.ts
```diff
@@ -0,0 +1,69 @@
...
+// A set of VSCode extension ID's that are currently uninstalling
+const uninstallingIDs = new Set();
+
+/**
+ * Get all of the installed extensions that currently conflict with VSCode-YAML
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function getConflictingExtensions(): Extension<any>[] {
+  const conflictingExtensions: Extension<any>[] = [];
+  conflictingIDs.forEach((extension) => {
```
webknjaz, line 25: This is not a typical js idiom. Instead, just use an `Array.prototype.map()` to convert a list of objects into a different type.

webknjaz: You can then chain `Array.prototype.filter()` to drop unnecessary items.

src/extensionConflicts.ts
```diff
@@ -0,0 +1,69 @@
...
+  return conflictingExtensions;
+}
+
+/**
+ * Display the uninstall conflicting extension notification if there are any conflicting extensions currently installed
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function showUninstallConflictsNotification(conflictingExts: Extension<any>[]): void {
+  // Add all available conflicting extensions to the uninstalling IDs map
+  for (const extIndex in conflictingExts) {
```
webknjaz, line 40: It's more straightforward to just use a "for ... of ..." loop
```suggestion
  for (let ext of conflictingExts) {
```

src/extensionConflicts.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  const uninstallMsg = 'Uninstall';
+
+  // Gather all the conflicting display names
+  let conflictMsg = '';
+  if (conflictingExts.length === 1) {
+    conflictMsg = `${conflictingExts[0].packageJSON.displayName} extension is incompatible with VSCode-YAML. Please uninstall it.`;
+  } else {
+    const extNames: string[] = [];
+    conflictingExts.forEach((ext) => {
```
webknjaz, line 53: Plz use `map()` here.

webknjaz (changes requested):

src/extensionConflicts.ts
ssbarnea: The entire code was copied from yaml extension and I only made the minimum number of changes in order to pass our linting. I do not want to diverge more as it would make harder to backport changes from there.

webknjaz: Then it should go in a special `_vendor` location and never be linted.

ssbarnea: I implemented the changes you requested so vendoring is no longer needed.

src/extension.ts
```diff
@@ -12,11 +12,12 @@ import {
 
 /* local */
 import { AnsiblePlaybookRunProvider } from './features/runner';
-
+import { getConflictingExtensions, showUninstallConflictsNotification } from './extensionConflicts';
 
 let client: LanguageClient;
 
 export function activate(context: ExtensionContext): void {
+
```
webknjaz, line 20: Is this some autoformatting change? In Python, it's usually required not to have an empty line right after the function signature.

src/extension.ts
```diff
@@ -12,11 +12,12 @@ import {
 
 /* local */
 import { AnsiblePlaybookRunProvider } from './features/runner';
-
```
webknjaz, line 15: Same: plz keep a two-line separator between imports and the module body.

src/extension.ts
```diff
@@ -49,6 +50,15 @@ export function activate(context: ExtensionContext): void {
...
   client.start();
+
+  notifyAboutConflicts();
+  client.onReady().then(() => {
+    // If the extensions change, fire this notification again to pick up on any association changes
+    extensions.onDidChange(() => {
+      notifyAboutConflicts();
+    });
+  }
+  );
```
webknjaz, line 61: the indentation here is broken

src/extension.ts
```diff
@@ -57,3 +67,15 @@ export function deactivate(): Thenable<void> | undefined {
   }
   return client.stop();
 }
+
+/**
+ * Finds extensions that conflict with our extension.
+ * If one or more conflicts are found then show an uninstall notification
+ * If no conflicts are found then do nothing
+ */
+ function notifyAboutConflicts(): void {
```
webknjaz, line 76:
```suggestion
function notifyAboutConflicts(): void {
```

src/extension.ts
```diff
@@ -57,3 +67,15 @@ export function deactivate(): Thenable<void> | undefined {
...
+  const conflictingExtensions = getConflictingExtensions();
+  if (conflictingExtensions.length > 0) {
+    showUninstallConflictsNotification(conflictingExtensions);
+  }
+ }
```
webknjaz, line 81:
```suggestion
}
```

src/extensionConflicts.ts
```diff
@@ -0,0 +1,66 @@
...
+  if (conflictingExts.length > 0) {
+    window.showInformationMessage(conflictMsg, uninstallMsg).then((clickedMsg) => {
+      if (clickedMsg === uninstallMsg) {
+        conflictingExts.forEach((ext) => {
+          commands.executeCommand('workbench.extensions.uninstallExtension', ext.id);
+          uninstallingIDs.delete(ext.id);
+        });
+      }
+    });
+  }
```
webknjaz, line 65:
```suggestion
  if (!conflictingExts.length) {
    return;
  }
  if (clickedMsg === uninstallMsg) {
    conflictingExts = [];
  }
  window.showInformationMessage(conflictMsg, uninstallMsg).then(clickedMsg => {
    for (let ext of conflictingExts) {
      commands.executeCommand('workbench.extensions.uninstallExtension', ext.id);
      uninstallingIDs.delete(ext.id);
    }
  });
```

src/extension.ts
ssbarnea: Apparently this was not from eslint so I was able to remove it. Still, i think it would be preferable to keep formatting for the linter alone. If there is something it does not do,
we can look into reconfiguring it in a separate change.

src/extension.ts
ssbarnea: I fixed it but if you want this to be followed please make a change to eslint config to enforce it. I guess https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/newline-after-import.md should do the trick.

src/extension.ts
ssbarnea: It was fixed.

src/extensionConflicts.ts
```diff
@@ -0,0 +1,68 @@
...
+import { commands, Extension, extensions, window } from 'vscode';
+
+// A set of VSCode extension ID's that conflict with our extension
+const conflictingIDs = [
+  'haaaad.ansible',
+  'sysninja.vscode-ansible-mod',
+  'tomaciazek.ansible',
+  'vscoss.vscode-ansible',
+  'zbr.vscode-ansible',
+];
```
cidrblock, line 14: How is this list going to be maintained?

cidrblock (approved):
Did not pull and test, one question about maintaining the list moving forward, which shouldn't block the merge

src/extensionConflicts.ts
webknjaz: I wonder if this could be moved to `package.json`

src/extensionConflicts.ts
```diff
@@ -0,0 +1,68 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Red Hat, Inc. All rights reserved..
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
```
webknjaz, line 3: There's no `License.txt`. How to handle this correctly?

src/extensionConflicts.ts
```diff
@@ -0,0 +1,68 @@
...
+  'vscoss.vscode-ansible',
+  'zbr.vscode-ansible',
+];
+
+// A set of VSCode extension ID's that are currently uninstalling
+const uninstallingIDs = new Set();
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function isExtensionPresent(obj: any): obj is Extension<any> {
+  return obj !== undefined && !uninstallingIDs.has(obj.id);
```
webknjaz, line 21: IIRC, the correct check should be
```suggestion
  return typeof obj !== 'undefined' && !uninstallingIDs.has(obj.id);
```

src/extensionConflicts.ts
```diff
@@ -0,0 +1,68 @@
...
+  // Gather all the conflicting display names
+  let conflictMsg = '';
+  if (conflictingExts.length === 1) {
+    conflictMsg = `${conflictingExts[0].packageJSON.displayName} extension is incompatible with redhat.ansible. Please uninstall it.`;
+  } else {
+    const extNames: string[] = conflictingExts.map(ext => ext.packageJSON.displayName);
+    conflictMsg = `The ${extNames.join(', ')} extensions are incompatible with redhat.ansible. Please uninstall them.`;
+  }
+
+  if (!conflictingExts.length) {
```
webknjaz, line 56: This guard expression should probably be the first thing in the function.

src/extensionConflicts.ts
```diff
@@ -0,0 +1,68 @@
...
+    return;
+  }
+  window.showInformationMessage(conflictMsg, uninstallMsg).then(clickedMsg => {
```
webknjaz, line 59: Why not make the outer function async and avoid the "promise hell" with an `await`?

src/extensionConflicts.ts
ssbarnea: done

src/extensionConflicts.ts
ssbarnea: done

src/extensionConflicts.ts
ssbarnea: done

src/extensionConflicts.ts
ssbarnea: done

src/extensionConflicts.ts
ssbarnea: There is no need to move this anywhere as I do not expect us to do much maintenance on it. Probably we will have to modify the list less than once a year. All the other extensions that have a similar method did the same.

src/extensionConflicts.ts
webknjaz: Did you forget to push?

ssbarnea: yep, done.

src/extensionConflicts.ts
tomaciazek: That would be the case for JavaScript, and if you expect the variable to be completely undeclared. Here we obviously do have the variable declared and the TypeScript compiler will prevent using undeclared variables anyway. So, `obj !== undefined` is perfectly fine in TypeScript, and preferable since it's shorter/less complex.

src/extensionConflicts.ts
webknjaz: Structurally, it's more obvious when it's in some sort of a data file / metadata. Having to update a list rarely is a questionable excuse for suboptimal design decisions. If the metadata is in a wrong place, it's wrong regardless of when it's going to be updated because the problem in question is of a different category.
Anyway, I'm not going to block this PR on this.

src/extensionConflicts.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  await window.showInformationMessage(conflictMsg, uninstallMsg).then(clickedMsg => {
+    if (clickedMsg !== uninstallMsg) {
+      return;
+    }
+    for (const ext of conflictingExts) {
+      commands.executeCommand('workbench.extensions.uninstallExtension', ext.id);
+      uninstallingIDs.delete(ext.id);
+    }
+  });
```
webknjaz, line 67:
```suggestion
  const clickedMsg = await window.showInformationMessage(conflictMsg, uninstallMsg);
  if (clickedMsg !== uninstallMsg) {
    return;
  }
  for (const ext of conflictingExts) {
    commands.executeCommand('workbench.extensions.uninstallExtension', ext.id);
    uninstallingIDs.delete(ext.id);
  }
```

webknjaz:
Plz also fix the commit message / PR title to contain useful info (such as the high-level effect of the change for the user).

ssbarnea:
> Plz also fix the commit message / PR title to contain useful info (such as the high-level effect of the change for the user).

Done, make pr title more in sync with the changelog entry, stating clearly
what this change does.

src/extensionConflicts.ts
```diff
@@ -0,0 +1,67 @@
...
+// A set of VSCode extension ID's that are currently uninstalling
+const uninstallingIDs = new Set();
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function isExtensionPresent(obj: any): obj is Extension<any> {
+  return typeof obj !== 'undefined' && !uninstallingIDs.has(obj.id);
+}
+
+/**
+ * Get all of the installed extensions that currently conflict with VSCode-YAML
```
ganeshrn, line 25: Replace vscode-yaml with redhat.ansible extension?

ganeshrn (approved):
LGTM and good merge after minor doc comment is fixed

src/extensionConflicts.ts
ssbarnea: Thanks! I fixed that.

webknjaz (approved):

Issue #158, opened 2021-08-03
[master] Cannot read property 'slice' of undefined


ssbarnea:
### Summary

When opening https://github.com/ansible-community/ansible-lint/blob/main/examples/galaxy.yml file and setting the type as `Ansible`, our extension reports: Cannot read property 'slice' of undefined

### Extension version

master

### VS Code version

1.58.2

### Ansible Version

```shell
-
```

### OS / Environment

-

### Relevant log output

```shell
[2021-08-03 14:33:39.534] [exthost] [warning] redhat.vscode-yaml - Code actions of kind 'quickfix 'requested but returned code action does not have a 'kind'. Code action will be dropped. Please set 'CodeAction.kind'.
[2021-08-03 14:33:40.653] [exthost] [warning] redhat.vscode-yaml - Code actions of kind 'quickfix 'requested but returned code action does not have a 'kind'. Code action will be dropped. Please set 'CodeAction.kind'.
```

ssbarnea:
I am almost sure this comes from the language server because we have no call of slice inside the extension, but there are few ones on ALS.

ganeshrn:
@ssbarnea I am not able to reproduce the issue. Please refer attached snapshot. 

<img width="1392" alt="Screen Shot 2021-09-30 at 2 05 17 PM" src="https://user-images.githubusercontent.com/5292461/135417837-90c08a5e-4f92-4507-8ae1-8977dc440176.png">


Can you please share detailed reproducers steps which you dev environment setup details and relevant snapshots.

ssbarnea:
This popup appears like 5-10 seconds after vscode starts, the time needed for Ansible extension to fully initialize:

![](https://sbarnea.com/ss/Screen-Shot-2021-10-01-14-57-31.38.png)

As you can see the document already uses Ansible language.

Other observations:
* When vscode just opens a file (no workspace open) aka "violet status bar mode", extension does not give this error. Still, it does seem to provide any auto-complete activity and the ansible-lint errors are not recognized.
* When vscode has a workspace open (blue sidebar) the slice error message appears

I even tested is VSCodium with a fresh empty repo and the minimal set of extensions and the same error happens.

![](https://sbarnea.com/ss/Screen-Shot-2021-10-01-15-17-26.74.png)

PR #177, opened 2021-09-05
fix: restore original ansible vault code

### LINKED ISSUE #171: Restore support for ansible-vault
### Problem

The huge PR https://github.com/ansible/vscode-ansible/pull/142 removed the support of ansible-vault, which could previously be used to encrypt/decrypt whole files or just the current text selection as an inline-vault-string. 

Please bring this back.

### Solution

Restore the removed functionality. The implementation may differ from the previous implementation, but this extension should have (at least) feature parity with the state before https://github.com/ansible/vscode-ansible/pull/142 merged.

### Alternatives

_No response_

### Additional context

_No response_

jeinwag:
A quick & dirty "backport" of the original ansible vault code which got lost when PR #142 was merged. Fixes #171.


package.json
```diff
@@ -90,6 +95,15 @@
           "default": "",
           "description": "Path to the virtual environment activation script. Use only if you have a custom activation script. It will be sourced using bash before executing Ansible commands. If set, the Interpreter Path setting is ignored."
         },
+        "ansible.vault.executablePath": {
+          "default": "ansible-vault",
+          "description": "%configuration.vault.executablePath%",
```
webknjaz, line 100: Pls add this to https://github.com/ansible/vscode-ansible/blob/main/package.nls.json

jeinwag: Looks like it is already there? https://github.com/ansible/vscode-ansible/blob/5ef7df70ea5834b3b163c37149153027f543ffbf/package.nls.json#L13

ganeshrn (commented):
Thank you for re-adding the vault feature and apologies that it was removed during the refactoring. 

For achieving feature parity with the older plugin this PR looks good to me.
However, in long term, we can explore the possibility of supporting this functionality in the ansible-language-server https://github.com/ansible/ansible-language-server so that other clients can also use it.

ssbarnea:
@jeinwag one mistake that I observed is that you forgot to run `npm install` and include changes made to the package lock. Due to this the lockfile is missing the newly added deps.

ssbarnea (changes requested):
I tried to encrypt a vars file and I observed an unnamed input box which I expected to be the password but entering something there did not had any effect. Few seconds later I got an error popup reporting: "Command 'Ansible Vault: Encrypt/Decrypt' resulted in an error (Running the contributed command: 'extension.ansible.vault' failed.)"

By looking at the Log (Extension Host) output I seen some weird messaging:
```
[2021-09-09 12:33:07.598] [exthost] [error] Error: Command failed: ansible-vault encrypt "/var/folders/3q/pc1jcyjj3qqbj6_x7n4kbtqw0000gn/T/tmp-43559-JYoNwmt0OU03" --encrypt-vault-id="default"
/Users/ssbarnea/.pyenv/versions/3.9.6/lib/python3.9/getpass.py:91: GetPassWarning: Can not control echo on the terminal.
  passwd = fallback_getpass(prompt, stream)
Warning: Password input may be echoed.
New Vault password: [1;35m[WARNING]: Error in vault password prompt (default): EOFError (ctrl-d) on[0m
[1;35mprompt for (default)[0m
[0;31mERROR! EOFError (ctrl-d) on prompt for (default)[0m

	at checkExecSyncError (child_process.js:625:11)
	at Object.execSync (child_process.js:661:15)
	at Object.<anonymous> (electron/js2c/asar_bundle.js:5:12288)
	at exec (/Users/ssbarnea/c/a/vscode-ansible/out/client/features/vault.js:194:15)
	at encryptFile (/Users/ssbarnea/c/a/vscode-ansible/out/client/features/vault.js:176:9)
	at encryptInline (/Users/ssbarnea/c/a/vscode-ansible/out/client/features/vault.js:144:5)
	at toggleEncrypt (/Users/ssbarnea/c/a/vscode-ansible/out/client/features/vault.js:95:48)
	at runMicrotasks (<anonymous>)
	at processTicksAndRejections (internal/process/task_queues.js:93:5)
	at async _executeContributedCommand (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:94:110871) extension.ansible.vault
```

I think that this encountered some interactive ansible output which is something that can easily go wrong. I think we should do our best to avoid any interactive prompts from ansible (run it in non interactive mode).

jeinwag:
@ssbarnea it's not asking for the password, but for the vault identity to use to encrypt the file. I added a fix to display an error message if there are no identities defined.

src/features/vault.ts
```diff
@@ -0,0 +1,247 @@
...
+  if (identityList.length !== 0) {
+    await vscode.window.showQuickPick(identityList).then((chosen) => {
+      if (chosen) {
+        vaultId = chosen;
+      }
+    });
+    return vaultId;
+  } else {
+    return undefined;
+  }
```
webknjaz, line 39: Let's use a guard expression here, plus it'd be cleaner to just use await everywhere w/o combining it with then-chains:
```suggestion

  if (!identityList.length) {
    return undefined;
  }

  const chosenVault = await vscode.window.showQuickPick(identityList);
  return chosenVault || vaultId;
```

src/features/vault.ts
```diff
@@ -0,0 +1,247 @@
...
+  let identityList: string[] | undefined;
+  if (ansibleCfg === 'ANSIBLE_VAULT_IDENTITY_LIST') {
+    const envVar: string | undefined = process.env.ANSIBLE_VAULT_IDENTITY_LIST;
+    if (envVar !== undefined) {
+      identityList = envVar
+        .split(',')
+        .map((id: string) => id.split('@', 2)[0].trim());
+    } else {
+      identityList = [];
+    }
```
webknjaz, line 18:
```suggestion
    identityList = typeof envVar === 'undefined'
      ? []
      : envVar
        .split(',')
        .map((id: string) => id.split('@', 2)[0].trim());
```

src/features/vault.ts
```diff
@@ -0,0 +1,247 @@
...
+  } else {
+    const cfg: utilAnsibleCfg.AnsibleVaultConfig | undefined =
+      utilAnsibleCfg.getValueByCfg(ansibleCfg);
+    if (cfg !== undefined) {
+      identityList = cfg.defaults.vault_identity_list
+        .split(',')
+        .map((id: string) => id.split('@', 2)[0].trim());
+    } else {
+      identityList = [];
+    }
```
webknjaz, line 28:
```suggestion
    identityList = typeof cfg === 'undefined'
      ? []
      : cfg.defaults.vault_identity_list
        .split(',')
        .map((id: string) => id.split('@', 2)[0].trim());
```

src/features/vault.ts
```diff
@@ -0,0 +1,247 @@
...
+  } else {
+    const cfg: utilAnsibleCfg.AnsibleVaultConfig | undefined =
+      utilAnsibleCfg.getValueByCfg(ansibleCfg);
+    if (cfg !== undefined) {
+      identityList = cfg.defaults.vault_identity_list
+        .split(',')
+        .map((id: string) => id.split('@', 2)[0].trim());
```
webknjaz, line 25: I noticed that this transformation is copy-pasted in two places, so it's better to just apply it after the whole if-else block.

src/features/vault.ts
```diff
@@ -0,0 +1,247 @@
...
+  // Read `ansible.cfg` or environment variable
+  const rootPath: string | undefined = utilAnsibleCfg.getRootPath(
+    editor.document.uri
+  );
+  let keyInCfg: string;
+  if (!!process.env.ANSIBLE_VAULT_IDENTITY_LIST) {
+    keyInCfg = 'ANSIBLE_VAULT_IDENTITY_LIST';
+  } else {
+    keyInCfg = utilAnsibleCfg.scanAnsibleCfg(rootPath);
+  }
```
webknjaz, line 71: This should really be a `const`
```suggestion
  const keyInCfg: string = !!process.env.ANSIBLE_VAULT_IDENTITY_LIST
    ? 'ANSIBLE_VAULT_IDENTITY_LIST'
    : utilAnsibleCfg.scanAnsibleCfg(rootPath);
```

src/features/vault.ts
```diff
@@ -0,0 +1,247 @@
...
+  const encryptedText = fs.readFileSync(tmpFilename, 'utf8');
+  console.log(`encryptedText == '${encryptedText}'`);
+
+  if (!!tmpFilename) {
+    fs.unlinkSync(tmpFilename);
+    console.log(`Removed temporary file: '${tmpFilename}'`);
+  }
+
+  return encryptedText.trim();
+};
```
webknjaz, line 181: Why isn't this async?

src/features/vault.ts
```diff
@@ -0,0 +1,247 @@
...
+  if (!!rootPath) {
+    exec(cmd, { cwd: rootPath });
+  } else {
+    exec(cmd);
+  }
+};
+
+const exec = (cmd: string, opt = {}) => {
+  console.log(`> ${cmd}`);
+  return cp.execSync(cmd, opt);
```
webknjaz, line 246: I don't like that there's a lot of blocking invocations that can potentially be disruptive from the UX perspective and annoy the end-users by blocking the rendering. Any chance to change all of this to async?

tsconfig.json
```diff
@@ -8,7 +8,8 @@
     "outDir": "out/client",
     "rootDir": "src",
     "sourceMap": true,
-    "strict": true
+    "strict": true,
+    "esModuleInterop": true
```
webknjaz, line 12: Would you clarify the motivation behind adding this?

src/features/utils/ansibleCfg.ts
```diff
@@ -0,0 +1,90 @@
...
+      !!parsed.defaults.vault_identity_list
+    ) {
+      const cfg: AnsibleVaultConfig = {
+        defaults: { vault_identity_list: parsed.defaults.vault_identity_list },
+      };
+      return cfg;
+    }
+  }
+
+  return undefined;
```
webknjaz, line 89: I believe this is easier to digest:
```suggestion
  if (!fs.existsSync(path)) {
    return undefined;
  }

  const vault_identity_list =
    ini.parse(fs.readFileSync(path, 'utf-8')).?defaults.?vault_identity_list;
  if (!vault_identity_list) {
    return undefined;
  }

  return {
    defaults: { vault_identity_list },
  } as AnsibleVaultConfig;
```

src/features/utils/ansibleCfg.ts
```diff
@@ -0,0 +1,90 @@
...
+      console.log(`Found 'defaults.vault_identity_list' within '${cfgPath}'`);
+      return cfgPath;
+    }
+  }
+
+  console.log("Found no 'defaults.vault_identity_list' within config files");
+  return '';
+}
+
+export function getValueByCfg(path: string): AnsibleVaultConfig | undefined {
```
webknjaz, line 72: It'd be way more responsive to have this whole function async.

src/features/utils/ansibleCfg.ts
```diff
@@ -0,0 +1,90 @@
...
+export function getRootPath(editorDocumentUri: vscode.Uri): string | undefined {
+  let rootPath: string | undefined = undefined;
+
+  if (!!vscode.workspace.workspaceFolders) {
+    rootPath = vscode.workspace.workspaceFolders.length
+      ? vscode.workspace.workspaceFolders[0].name
+      : undefined;
+  }
+
+  if (!!vscode.workspace.getWorkspaceFolder) {
```
webknjaz, line 16: A semantically correct check would be
```suggestion
  if (typeof vscode.workspace.getWorkspaceFolder === 'function') {
```

src/features/utils/ansibleCfg.ts
```diff
@@ -0,0 +1,90 @@
...
+    const workspaceFolder =
+      vscode.workspace.getWorkspaceFolder(editorDocumentUri);
+
+    if (!!workspaceFolder) {
+      rootPath = workspaceFolder.uri.path;
+    } else {
+      // not under any workspace
+      rootPath = undefined;
+    }
```
webknjaz, line 25:
```suggestion
    rootPath = workspaceFolder.?uri.?path;
```

src/features/utils/ansibleCfg.ts
```diff
@@ -0,0 +1,90 @@
...
+  if (!!rootPath) {
+    cfgFiles.unshift(`${rootPath}/ansible.cfg`);
+  }
+
+  if (!!process.env.ANSIBLE_CONFIG) {
+    cfgFiles.unshift(process.env.ANSIBLE_CONFIG);
+  }
+
+  for (let i = 0; i < cfgFiles.length; i++) {
+    const cfgFile = cfgFiles[i];
```
webknjaz, line 58:
```suggestion
  for (let cfgFile of cfgFiles) {
```

src/features/utils/ansibleCfg.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    const cfg = getValueByCfg(cfgPath);
+    if (!!cfg && !!cfg.defaults && !!cfg.defaults.vault_identity_list) {
+      console.log(`Found 'defaults.vault_identity_list' within '${cfgPath}'`);
+      return cfgPath;
+    }
+  }
+
+  console.log("Found no 'defaults.vault_identity_list' within config files");
+  return '';
```
webknjaz, line 69: Why empty string? It's semantically confusing. The path cannot be an empty string. Using an `undefined` would signal "not found" better.

src/features/utils/ansibleCfg.ts
```diff
@@ -0,0 +1,90 @@
...
+  if (!!process.env.ANSIBLE_CONFIG) {
+    cfgFiles.unshift(process.env.ANSIBLE_CONFIG);
+  }
+
+  for (let i = 0; i < cfgFiles.length; i++) {
+    const cfgFile = cfgFiles[i];
+    const cfgPath = untildify(cfgFile);
+
+    const cfg = getValueByCfg(cfgPath);
+    if (!!cfg && !!cfg.defaults && !!cfg.defaults.vault_identity_list) {
```
webknjaz, line 62:
```suggestion
    if (!!cfg?.defaults?.vault_identity_list) {
```

src/extension.ts
```diff
@@ -13,7 +15,6 @@ import {
 /* local */
 import { AnsiblePlaybookRunProvider } from './features/runner';
 
-
```
webknjaz, line 16: Keep the two-line section separator here.

src/extension.ts
```diff
@@ -13,7 +15,6 @@ import {
 /* local */
 import { AnsiblePlaybookRunProvider } from './features/runner';
```
webknjaz, line 17:
```suggestion


```

src/extension.ts
```diff
@@ -49,6 +50,12 @@ export function activate(context: ExtensionContext): void {
 
   // start the client and the server
   client.start();
+
+  const disposable = vscode.commands.registerCommand(
+    'extension.ansible.vault',
+    toggleEncrypt
+  );
+  context.subscriptions.push(disposable);
```
webknjaz, line 58: plz move this to the beginning of the function because the rest of the client-side setup is done on line 21 so I'd keep this closeby

src/features/utils/ansibleCfg.ts
```diff
@@ -0,0 +1,90 @@
...
+import * as ini from 'ini';
+
+// Get rootPath based on multi-workspace API
+export function getRootPath(editorDocumentUri: vscode.Uri): string | undefined {
+  let rootPath: string | undefined = undefined;
+
+  if (!!vscode.workspace.workspaceFolders) {
+    rootPath = vscode.workspace.workspaceFolders.length
+      ? vscode.workspace.workspaceFolders[0].name
+      : undefined;
```
webknjaz, line 13: This could be just
```suggestion
    rootPath = vscode.workspace.workspaceFolders?.[0]?.name;
```

tsconfig.json
jeinwag: The untildify dependency requires this flag to be set.

jeinwag:
@webknjaz Thanks for the feedback, I believe I applied all your suggestions and removed all the synchronous calls except one call to fs.existsSync.

src/features/utils/ansibleCfg.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  if (typeof vscode.workspace.getWorkspaceFolder === 'function') {
+    const workspaceFolder =
+      vscode.workspace.getWorkspaceFolder(editorDocumentUri);
+
+    rootPath = workspaceFolder?.uri?.path;
+  }
+
+  return rootPath;
```
webknjaz, line 21: If I understand the intent behind this correctly, it could be simplified even more:
```suggestion
  if (typeof vscode.workspace.getWorkspaceFolder !== 'function') {
      return vscode.workspace.workspaceFolders?.[0]?.name;
  }

  return vscode.workspace.getWorkspaceFolder(editorDocumentUri)?.uri?.path;
```

src/features/utils/ansibleCfg.ts
```diff
@@ -0,0 +1,82 @@
...
+    const cfgPath = untildify(cfgFile);
+
+    const cfg = await getValueByCfg(cfgPath);
+    if (!!cfg?.defaults?.vault_identity_list) {
+      console.log(`Found 'defaults.vault_identity_list' within '${cfgPath}'`);
+      return cfgPath;
+    }
+  }
+
+  console.log('Found no \'defaults.vault_identity_list\' within config files');
```
webknjaz, line 60:
```suggestion
  console.log("Found no 'defaults.vault_identity_list' within config files");
```

src/features/utils/ansibleCfg.ts
```diff
@@ -0,0 +1,82 @@
...
+  return undefined;
```
webknjaz, line 61: How about
```suggestion
  const cfgPath = cfgFiles
  	.map(cf => untildify(cf))
  	.map(async cp => await getValueByCfg(cp))
  	.find(cfg => !!cfg?.defaults?.vault_identity_list)

  console.log(
    typeof cfgPath != 'undefined'
    : `Found 'defaults.vault_identity_list' within '${cfgPath}'`
    ? "Found no 'defaults.vault_identity_list' within config files"
  );

  return cfgPath;
```

src/features/utils/ansibleCfg.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+export async function getValueByCfg(
+  path: string
+): Promise<AnsibleVaultConfig | undefined> {
+  console.log(`Reading '${path}'...`);
+
+  if (!fs.existsSync(path)) {
```
webknjaz, line 69: Why was it impossible to turn this async?

src/features/vault.ts
```diff
@@ -0,0 +1,241 @@
...
+  const vaultId = 'default';
+  let identitySource: string | undefined;
+
+  if (ansibleCfg === 'ANSIBLE_VAULT_IDENTITY_LIST') {
+    identitySource = process.env.ANSIBLE_VAULT_IDENTITY_LIST;
+  } else {
+    const cfg: utilAnsibleCfg.AnsibleVaultConfig | undefined =
+      await utilAnsibleCfg.getValueByCfg(ansibleCfg);
+    identitySource =
+      typeof cfg === 'undefined' ? undefined : cfg.defaults.vault_identity_list;
```
webknjaz, line 20: Is this just
```suggestion
    identitySource = cfg?.defaults?.vault_identity_list;
```
?

src/features/vault.ts
```diff
@@ -0,0 +1,241 @@
...
+) => {
+  const tmpFilename = await tmp.tmpName();
+  await fs.promises.writeFile(tmpFilename, Buffer.from(text, 'utf8'));
+  console.log(`Wrote encrypted string to temporary file '${tmpFilename}'`);
+
+  encryptFile(tmpFilename, rootPath, vaultId, config);
+  const encryptedText = await fs.promises.readFile(tmpFilename, 'utf8');
+  console.log(`encryptedText == '${encryptedText}'`);
+
+  if (!!tmpFilename) {
```
webknjaz, line 169: I don't understand how it works. Is it modified by reference somewhere within the `readFile()` invocation?

src/features/vault.ts
```diff
@@ -0,0 +1,241 @@
...
+};
+
+const encryptInline = async (
+  text: string,
+  rootPath: string | undefined,
+  vaultId: string,
+  config: vscode.WorkspaceConfiguration
+) => {
+  const tmpFilename = await tmp.tmpName();
+  await fs.promises.writeFile(tmpFilename, Buffer.from(text, 'utf8'));
```
webknjaz, line 162: Is it possible to avoid writing an encrypted file? If not, we need to take great care to actually clean it up from disk. This is why the removal should be put in a try/finally block. And we need to think about the possible corner cases thoroughly since this presents a potential security problem if we miss something. What can we do to make this more robust?

src/features/vault.ts
```diff
@@ -0,0 +1,241 @@
...
+  console.log(`Wrote encrypted string to temporary file '${tmpFilename}'`);
+
+  encryptFile(tmpFilename, rootPath, vaultId, config);
+  const encryptedText = await fs.promises.readFile(tmpFilename, 'utf8');
+  console.log(`encryptedText == '${encryptedText}'`);
```
webknjaz, line 167: I'm worried that we may be leaking sensitive information into places we don't control. While this may be okay for debugging (you could check if the extension is in dev mode programmatically), I wouldn't recommend keeping this as is. Also, maybe change the log level too?
```suggestion
  console.debug(`encryptedText == '${encryptedText}'`);
```

src/features/vault.ts
```diff
@@ -0,0 +1,241 @@
...
+
+  if (!!tmpFilename) {
+    await fs.promises.unlink(tmpFilename);
```
webknjaz, line 170: If unlinking fails, we should probably alert the user loudly to give them a chance to clean up the sensitive data from disk.

src/features/vault.ts
```diff
@@ -0,0 +1,241 @@
...
+  return text.indexOf('$ANSIBLE_VAULT;') === 0 ? 'encrypted' : 'plaintext';
+};
+
+const encryptInline = async (
+  text: string,
+  rootPath: string | undefined,
+  vaultId: string,
+  config: vscode.WorkspaceConfiguration
+) => {
+  const tmpFilename = await tmp.tmpName();
```
webknjaz, line 161: We may want to also make sure to have as strict ACL setup as possible. Ansible itself even uses SElinux (where available), not only regular UNIX permissions to secure such files better.

src/features/vault.ts
```diff
@@ -0,0 +1,241 @@
...
+
+  if (!!tmpFilename) {
+    await fs.promises.unlink(tmpFilename);
+    console.log(`Removed temporary file: '${tmpFilename}'`);
+  }
+
+  return encryptedText.trim();
+};
+
+const decryptInline = async (
```
webknjaz, line 177: Oh, I've missed with the previous comments about ACLs, most of them are probably more important wrt this function.

webknjaz:
@jeinwag thanks for updating the PR with the requested changes! I've performed the second round of review. Please, take a look.

@ansible/devtools folks, may I ask you all to participate in getting this PR to the finish line? It's rather good but needs more eyes on it. TIA!

src/features/utils/ansibleCfg.ts
jeinwag: Sorry, I just noticed that there is no fs.promises.exists and didn't look any further (I'm not a JS or node programmer). I looked at the docs again and replaced existsSync with a call to fs.promises.access.

src/features/vault.ts
jeinwag: I'll see if I can make this work without writing temporary files to disk.

src/features/vault.ts
jeinwag: I've modified the original code so that it doesn't create any temporary files during inline de- or encryption, it just pipes the text directly through ansible-vault.

ssbarnea (changes requested):
When I try to run encrypt/decrypt, I get an error `Command 'Ansible Vault: Encrypt/Decrypt' resulted in an error (Running the contributed command: 'extension.ansible.vault' failed.)`.

I am not saying that this is necessarily caused by current code as I am affected by https://github.com/ansible/vscode-ansible/issues/158 but we need to wait to get the CI pipelines fixed for both extension and language server before merging this change.

jeinwag:
@ssbarnea I noticed an issued with error handling when there are no ansible-vault identities configured. Now the extension should display an error message, stating that it can't de- or encrypt files without an identity.

jeinwag:
@FlorianLaunay ah dang, I think I messed up the executable path setting during the rebase of my changes, I had it in there before :-(
Added it again, but please, feel free to help. This feature really is a must-have for us.

jeinwag:
@ssbarnea  I believe c02411e5ad7d39cc7d57ba19ec8bed633953d046 broke the pre-commit hooks. `npm install --production` removes `@commitlint/config-conventional` from node_modules, which then causes `commitlint` to fail.

jeinwag:
@ssbarnea I sent you a message on matrix. I fixed some issues with package.json.

PR #187, opened 2021-09-13
chore: ensure synced package-lock.json

ssbarnea:


webknjaz (approved):

.github/workflows/ci.yaml
```diff
@@ -51,3 +51,7 @@ jobs:
 
       - name: Run package
         run: npm run package
+
+      # extra safety measure that ensures code was not modified during build
```
webknjaz, line 55: I think this should be after install.

ssbarnea: The reason why I put it here is because also build, test and packaging can produce leftovers and I do not want to call it after each command.

webknjaz: Why is it important?

ssbarnea: Any untracked file is a bug (missing from `.gitignore`) and any modified file after run indicate possibility that we do not ship what we have in git. I do not want to allow changes that create such files that are missing to add them to .gitignore. I seen that as quite a common mistake as it is very unlikely to catch these during normal review process.

relrod (approved):

PR #191, opened 2021-09-15
chore: prevent incorrect versions

ssbarnea:
Basically we enforce `npm version` to run again with current version. This would bring repository into dirty state if version was out of think to be a noop otherwise.

The next command we run is a git dirty check, which would
stop the pipeline.


relrod (approved):
Approving.

For posterity: The `npm version` command here will cause file modifications in the working directory, if the versions don't match up. These file modifications are what will cause CI to fail, indicating a version mismatch.

PR #193, opened 2021-09-15
chore: cleaning codeowners files

ssbarnea:
Keeping devtools team as default reviewer and removing our
redhat-devtools friends as we now have enough people to
perform reviews. We will add the manually to the reviews
where we need their expertise.


webknjaz (approved):
I don't have a strong preference. I think it's fine as is but if you insist, here's your approval.

relrod (approved):

PR #195, opened 2021-09-16
chore: standardize indentation

ssbarnea:
Ensure that our js/ts code is indented using 2 spaces while
fixing it automatically. If we enable pre-commit.ci checks it should
fix incoming PRs automatically.

### DIFF
--- a/src/features/runner.ts
+++ b/src/features/runner.ts
@@ -9,8 +9,8 @@ import * as vscode from 'vscode';
 export class AnsiblePlaybookRunProvider {
     private disposableTerminal: vscode.Terminal | undefined;
     constructor(private vsCodeExtCtx: vscode.ExtensionContext) {
-        // this.disposableTerminal = this.makeTerminal();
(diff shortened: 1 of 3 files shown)


src/features/runner.ts
```diff
@@ -9,29 +9,29 @@ import * as vscode from 'vscode';
 export class AnsiblePlaybookRunProvider {
     private disposableTerminal: vscode.Terminal | undefined;
     constructor(private vsCodeExtCtx: vscode.ExtensionContext) {
-        // this.disposableTerminal = this.makeTerminal();
```
webknjaz, line 12: Urgh... Did I write this? Need to remember to remove in a separate PR.

webknjaz (approved):
Go for it!

src/features/runner.ts
ssbarnea: I doubt and I agree we need to take it out.

relrod (approved):

PR #196, opened 2021-09-22
feat: update default setting value for using FQCN

priyamsahoo:
As discussed in the team meetings, to ensure best practices, the default value for `ansible.ansible.useFullyQualifiedCollectionNames` is set to `true`.

That means, by default, the auto-completion in the editor would produce FQCNs for module names.

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
         "ansible.ansible.useFullyQualifiedCollectionNames": {
           "scope": "resource",
           "type": "boolean",
-          "default": false,
+          "default": true,
           "description": "Always use fully qualified collection names (FQCN) when inserting a module name. Disabling it will only use FQCNs when necessary."
         },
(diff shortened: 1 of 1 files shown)


ganeshrn (approved):

tomaciazek (commented):
We should at least put a note on that in the release notes, as for everyone who hasn't explicitly configured this setting, the behavior of the plugin will suddenly change.

ssbarnea:
> We should at least put a note on that in the release notes, as for everyone who hasn't explicitly configured this setting, the behavior of the plugin will suddenly change.

@tomaciazek Release notes are build automatically on https://github.com/ansible/vscode-ansible/releases when a PR is merged based on the title and labels, so we will not miss the behavior change. In fact that is the reason why I market this as "feature" instead of "bug".

ssbarnea (approved):

PR #199, opened 2021-09-26
fix: update compile withserver script

ganeshrn:
Update the `compile:withserver` script to create a soft
link in the `out/sever` folder for `node_modules` directory
of `ansible-language-server` to handle new package
depedency added in the language-server code while setting
up development environment

### DIFF
--- a/package.json
+++ b/package.json
@@ -365,7 +365,7 @@
   },
   "scripts": {
     "compile": "tsc -b && copyfiles -u 4 'node_modules/ansible-language-server/out/server/**/*' out/server",
-    "compile:withserver": "tsc -b && tsc -p ../ansible-language-server --outDir out/server",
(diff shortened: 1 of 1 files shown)


priyamsahoo:
LGTM !

priyamsahoo (approved):
LGTM !

ssbarnea:
@ganeshrn Are we going to support development of this extension under Windows? Does `ln` work under WSL? Asking because we already have a Windows jobs for the language server, known to be broken.

If we do not support development we will need to setup jobs that only install the already built extension artifact and test it under Windows, a bit more complex pipeline but doable.

If we consider that as a supported use case, I would prefer not to deepen the breakage. I personally do not use Windows at all, so I am ok with each path.

matburt:
If you are staying in wsl, everything works the way you would expect it to:

image

It works best if you aren't crossing the filesystem boundary into Windows. The file will exist it probably just won't be usable:

image

This is a familiar pattern to WSL users and the WSL Remote capability in VSCode also follows this pattern. Users of that remote will also understand it well.

ssbarnea (approved):

Issue #139, opened 2021-07-30
Add e2e integration testing for extension


webknjaz:
**Is your feature request related to a problem? Please describe.**

No tests.

**Describe the solution you'd like**

Tests must exist.

**Describe alternatives you've considered**

Giving up on quality.

**Additional context**

There's not many good examples of writing integration tests for VS Code extensions so here's some pointers I've found:
- https://developers.redhat.com/blog/2019/11/18/new-tools-for-automating-end-to-end-tests-for-vs-code-extensions
- https://github.com/redhat-developer/vscode-extension-tester/wiki
- https://github.com/redhat-developer/vscode-extension-tester
- https://vscode.rocks/testing/
- https://github.com/mlewand-org/vscode-test-content
- https://viatsko.github.io/awesome-vscode/#resources-for-extension-developers

webknjaz:
I don't think this should be closed. There's no test examples, it's not solved.

ssbarnea:
True, it should be open until we finish setting the framework and a minimal battery of testing, so others can easily add more of them while adding other features of bugfixes.

`vscode-extension-tester` is already used now and it looks promising.

PR #206, opened 2021-09-29
fix: lower minimal vscode version to 1.26.1

ssbarnea:
Configure the project to use test runner and just install the extension
using it.

Partial-Fix: #139- chore: bootstrap test runner
- fix: lower minimal vscode version to 1.26.1


.github/workflows/ci.yaml
```diff
@@ -43,11 +43,15 @@ jobs:
...
 
-      # Run tests
-      - name: Run test
+      - name: Run test (current vscode)
         uses: GabrielBB/xvfb-action@v1.0
         with:
-          run: npm test
+          run: npm run test
+
+      - name: Run test (oldest vscode)
```
webknjaz, line 51: Separate tests should be in separate checks via a matrix. Otherwise, it makes it harder to navigate through the CI results.

webknjaz (changes requested):
What exactly are you fixing? If anything, I'd say it's a feature, not a bugfix.

.github/workflows/ci.yaml
ssbarnea: I really want to avoid an exploding matrix like we already have on the language server. Running the same tests with oldest vscode after and only if first fails makes sense and avoids creating an unmanageable list of individual jobs.

webknjaz: Well, to me it's unmanageable to debug weird failures when one thing runs in an unclean env. Plus, it's impossible to tell what failed by looking at a number of jobs.

PR #208, opened 2021-09-30
feat: introduce ui tests based on vscode-extension-tester

odockal:
Signed-off-by: Ondrej Dockal <odockal@redhat.com>

Partial-Fix: #139


ssbarnea (approved):
I also manually downloaded the pr and run the test locally and they worked as expected with `npm run ui-test`.

odockal:
@ssbarnea I had to update npm script "webpack" with `npm run compile` in order to compile the ts classes on out folder. also their placement in out/client/ui-test is no quite fortunate. It all could be probably done better.

ssbarnea:
> @ssbarnea I had to update npm script "webpack" with `npm run compile` in order to compile the ts classes in out folder. also their placement in out/client/ui-test is no quite fortunate. It all could be probably done better.

While trying to setup chai, I tried to use the `test/` root repository but I did not know all the [tricks](https://stackoverflow.com/questions/59042716/trying-es6-style-import-gives-cannot-use-import-statement-outside-a-module) required for making ts compile automatically. If we could add a follow-up to enable us to use `npx mocha` directly it would be even better.

.github/workflows/ci.yaml
```diff
@@ -49,6 +49,12 @@ jobs:
         with:
           run: npm test
 
+      # Run UI tests
+      - name: Run test
+        uses: GabrielBB/xvfb-action@v1.0
+        with:
+          run: npm run ui-test
```
webknjaz, line 56: This should go into a separate factor in the matrix. At least until the team agrees on how to handle it.

webknjaz (changes requested):
I like it, except for one bit below.

src/ui-test/extensionUITest.ts
```diff
@@ -0,0 +1,57 @@
...
+        const actionButton = await titlePart.getAction(
+          'Clear Extensions Search Results'
+        );
+        if (await actionButton.isEnabled()) {
+          await actionButton.click();
+        }
+      }
+      if (quickBox && (await quickBox.isDisplayed())) {
+        await quickBox.cancel();
+      }
```
webknjaz, line 54:
```suggestion
      if (!quickBox) {
        return;
      }

      await quickBox.isDisplayed();
      await quickBox.cancel();
```

src/ui-test/extensionUITest.ts
```diff
@@ -0,0 +1,57 @@
...
+} from 'vscode-extension-tester';
+
+/**
+ * @author Ondrej Dockal <odockal@redhat.com>
+ */
+export function extensionUIAssetsTest(): void {
+  describe('Verify base assets are available after installation', () => {
+    let view: ViewControl;
+    let sideBar: SideBarView;
+    let quickBox: InputBox;
```
webknjaz, line 17: It's not clear why this exists. It seems like in every code path it's `undefined`.

src/ui-test/extensionUITest.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    before(async function () {
```
webknjaz, line 19: Nit: why isn't it an arrow function? This is inconsistent with the outer lambda. Does `this` need to be bound to a different context?

odockal: It is mocha thing, arrow function wont let you get into mocha context to set timeout for the test: https://mochajs.org/#arrow-functions

.github/workflows/ci.yaml
odockal: I have removed it, please go ahead and set it on your own the way you prefer.

src/ui-test/extensionUITest.ts
odockal: very good point, unused it is indeed. Removing.

.github/workflows/ci.yaml
```diff
@@ -17,7 +17,11 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        npm-target:
+        - test
+        - ui-test
         os: [macos-latest, windows-latest, ubuntu-latest]
+
```
webknjaz, line 24:
```suggestion
```

.github/workflows/ci.yaml
webknjaz: done

ssbarnea:
@webknjaz Shortly `npm test` is just a [shortcut to `npm run test`](https://stackoverflow.com/questions/57031559/difference-between-npm-t-and-npm-run-test#:~:text=TL%3BDR%20there%20is%20no,same%20action%20in%20this%20case.) and we had no real test running.

#204 added the first part of extest preparation of vscode and installation of the extension, this channels is the follow-up for completing the test enablement. The are no reasons to have different commands, especially as you cannot really test the extension without installing it with extest.

package.json
```diff
@@ -407,10 +408,11 @@
     "prepare": "husky install",
     "pretest": "npm run compile && npm run lint",
     "test": "extest get-vscode && extest install-vsix",
+    "ui-test": "extest setup-and-run -i out/client/ui-test/allTestsSuite.js",
     "vscode:prepublish": "npm run webpack",
     "watch": "tsc -b -w",
     "watch:server": "tsc -p ../ansible-language-server --outDir out/server -w",
-    "webpack": "npm run clean && webpack --mode production --config ./webpack.config.js",
+    "webpack": "npm run clean && npm run compile && webpack --mode production --config ./webpack.config.js",
```
tomaciazek, line 415: Why do we need to compile first? Webpack does compilation too.

tomaciazek (approved):
LGTM. Just one comment, but not a showstopper.

package.json
odockal: Webpack does not compile ui test related files (*.ts) into *.js. It places only *.ts file under out folder.

odockal: So webpack-like compilation differs from what ```npm run compile``` does.

webknjaz:
> @webknjaz Shortly `npm test` is just a [shortcut to `npm run test`](https://stackoverflow.com/questions/57031559/difference-between-npm-t-and-npm-run-test#:~:text=TL%3BDR%20there%20is%20no,same%20action%20in%20this%20case.) and we had no real test running.

Oh, I know about this alias. I think we could have lightweight unit tests there and have heavy UI tests separately. This could be debated separately.

package.json
webknjaz: I'm going to mark this as resolved to unblock the release.

webknjaz (approved):

PR #212, opened 2021-10-04
fix: switch debug listening port to fix unbound breakpoints

tomaciazek:
This fixes the issue with unbound breakpoints in new VS Code versions, reported by @ganeshrn. Apparently something else overshadows language server on the 6009 debug port, but I haven't been able to pinpoint what or why (or how).

VS Code was now also reporting long times searching for source-maps, which was caused by recently added mechanism linking `node_modules`. I've narrowed the `outFiles` configuration down to `src` avoid this problem.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -24,18 +24,18 @@
       "type": "node",
       "request": "attach",
       "timeout": 30000,
-      "port": 6009,
+      "port": 6010,
       "restart": true,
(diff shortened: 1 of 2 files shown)


ssbarnea (approved):
I also seen the 6009 port being used problem, not sure about the source.

ganeshrn (approved):

ganeshrn:
@tomaciazek This is a good find. Thank you for the PR

PR #220, opened 2021-10-08
fix: update npm dependencies

ssbarnea:
Result of running `npm update`. Hopefully we will do these using dependabot in  the future, with weekly or bi-monthly updates.


webknjaz:
Why is it "fix"? Nothing's broken.

PR #288, opened 2021-10-29
Fix indentation when using inline vault encrypt

jeinwag:
This should fix the non-perfect indentation when using inline encrypt.

### DIFF
--- a/src/features/vault.ts
+++ b/src/features/vault.ts
@@ -291,12 +291,10 @@ const getIndentationLevel = (
 ): number => {
   if (!editor.options.tabSize) {
     // according to VS code docs, tabSize is always defined when getting options of an editor
-    throw 'tabSize undefined, this should never happen';
+    throw new Error('The `tabSize` option is not defined, this should never happen.');
   }
   const startLine = editor.document.lineAt(selection.start.line).text;
   const indentationMatches = startLine.match(/^\s*/);
-  const leadingWhitespaces = !!indentationMatches
-    ? indentationMatches[0].length
-    : 0;
+  const leadingWhitespaces = indentationMatches?[0].?length || 0;
   return leadingWhitespaces / Number(editor.options.tabSize);
 };
--- a/src/features/vault.ts
+++ b/src/features/vault.ts
@@ -92,8 +92,11 @@ export const toggleEncrypt = async (): Promise<void> => {
         editBuilder.replace(
           selection,
           encryptedText.replace(
-            /\n/g,
-            `\n${' '.repeat(selection.start.character)}`
+            /\n\s*/g,
+            `\n${' '.repeat(
+              (getIndentationLevel(editor, selection) + 1) *
+                Number(editor.options.tabSize)
+            )}`
           )
         );
       });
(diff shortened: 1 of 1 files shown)


src/features/vault.ts
```diff
@@ -92,8 +92,11 @@ export const toggleEncrypt = async (): Promise<void> => {
         editBuilder.replace(
           selection,
           encryptedText.replace(
-            /\n/g,
-            `\n${' '.repeat(selection.start.character)}`
+            /\n\s*/g,
+            `\n${' '.repeat(
+              (getIndentationLevel(editor, selection) + 1) *
+                Number(editor.options.tabSize)
+            )}`
```
webknjaz, line 99: This is quite complex. Maybe move it out and assign it to a var?

src/features/vault.ts
```diff
@@ -281,3 +284,19 @@ const execCwd = (cmd: string, cwd: string | undefined) => {
...
   return exec(cmd, { cwd: cwd });
 };
+
+const getIndentationLevel = (
+  editor: vscode.TextEditor,
+  selection: vscode.Selection
+): number => {
+  if (!editor.options.tabSize) {
+    // according to VS code docs, tabSize is always defined when getting options of an editor
+    throw 'tabSize undefined, this should never happen';
```
webknjaz, line 294:
```suggestion
    throw new Error('The `tabSize` option is not defined, this should never happen.');
```

src/features/vault.ts
```diff
@@ -281,3 +284,19 @@ const execCwd = (cmd: string, cwd: string | undefined) => {
...
+  }
+  const startLine = editor.document.lineAt(selection.start.line).text;
+  const indentationMatches = startLine.match(/^\s*/);
+  const leadingWhitespaces = !!indentationMatches
+    ? indentationMatches[0].length
+    : 0;
```
webknjaz, line 300: Could this be something like
```suggestion
  const leadingWhitespaces = indentationMatches?[0].?length || 0;
```
?

src/features/vault.ts
```diff
@@ -281,3 +284,19 @@ const execCwd = (cmd: string, cwd: string | undefined) => {
...
+  return leadingWhitespaces / Number(editor.options.tabSize);
```
webknjaz, line 301: Should this be
```suggestion
  return leadingWhitespaces / new Number(editor.options.tabSize);
```

webknjaz: Ah, never mind. The docs seem to imply that it's not a constuctor.

src/features/vault.ts
jeinwag: Done.

ssbarnea (changes requested):
Apparently it is not working as intended because the indentation level of the encrypted text needs to be one level down from the line with !vault marker.

Keep in mind that original line can be at any indentation level.

```
  vars:
    username: !vault |
    $ANSIBLE_VAULT;1.1;AES256
    ...
    password:
      xxx: !vault |
     $ANSIBLE_VAULT;1.1;AES256
     ....
```

If you looked at the example above, for the first case the indentation was 4 chars instead of 6 and for the second case it was 5 chars instead of of expected 8.

My expectation is that the encrypted text to be one level down (2 more chars) than the vault. Am I wrong about it?

This was result of testing code from the change (or at least I hope that is what I did).

jeinwag:
@ssbarnea sorry, my last commit broke it. I fixed it again.

ssbarnea (approved):
Good that I tested. Indeed the last version works as expected. I personally find the popup notification message distracting but changing it would be outside the scope of this change. I will see if other users find it the same or not.

PR #291, opened 2021-11-01
Fix the line wrapping consistency in the changelog

webknjaz:


ssbarnea (changes requested):
If you really want to enforce line length limits for markdown files, please activate a linter for it.

I do not think that these kind of manual fixes are the best use of our time, especially when we do not include a preventive measure (lint)

webknjaz:
I'll look into linting when I have time. As you said, let's not waste it by blocking the merge.

ssbarnea (approved):

ssbarnea:
I do remember that one of the marketplaces had some special requirements regarding markdown. We will see if that breaks it or not and if so, we will need to add a note.

PR #300, opened 2021-11-09
changed default execution environment image

ganeshrn:
*  Update default execution environment image
   from `quay.io/ansible/ansible-navigator-demo-ee:0.6.0`
   to `quay.io/ansible/ansible-devtools-demo-ee:v0.1.0`

### DIFF
--- a/package.json
+++ b/package.json
@@ -182,7 +182,7 @@
         "ansible.executionEnvironment.image": {
           "scope": "resource",
           "type": "string",
-          "default": "'quay.io/ansible/ansible-devtools-demo-ee:v0.1.0",
+          "default": "quay.io/ansible/ansible-devtools-demo-ee:v0.1.0",
           "description": "Specify the name of the execution environment image."
         },
         "ansible.executionEnvironment.pullPolicy": {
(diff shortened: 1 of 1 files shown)


ssbarnea:
I think qe really need to use a :latest label as default and avoid having to update that. Is almost like adding a time-bomb, as we will definitely add more things into the dev-ee.

ganeshrn:
> I think we really need to use a :latest label as default and avoid having to update that. Is almost like adding a time-bomb, as we will definitely add more things into the dev-ee.

The default value for EE image is set so that users can get started with exploring EE easily and since this is user-facing I think we should pin it to a particular version.

ssbarnea (approved):
I will not block that change but I really think hardcoded version will force us to do more maintenance, not less. Once we publish a new `dee`, we will obviously desire for all users to make use of it.

I have no desire having to do support for old-versions and I do not think we should need to release a new extension just to ensure our users will not use an outdated `dee`. 

That container is owned by out team, we are responsible for releasing new versions so we cannot really claim we would protect ourselves from arbitrary changes made by others. In fact I would say that part of publishing a newer version should be testing it with our language server.

Approving it, but we should get feedback from @ansible/devtools and raise ticket to use a mobile label (latest).

Issue #302, opened 2021-11-09
Add script that verifies if current version was not already published


ssbarnea:
In order to force ourselves to bump version on default branch as soon we make a new release, we need a small script `tools/check-version.js` that ensures that the current version of the package from `package.json` was not already released to our two marketplaces.

Once we have the script we can easily add it to pre-commit config as a local hook, so we will always know to bump the version.

PR #304, opened 2021-11-10
Check version sanity script

priyamsahoo:
As discussed with @ssbarnea, the PR adds a script that 
1. basically checks the current project version with the latest extension version released in VSCode Marketplace and Open VSX. If it matches, then exits throwing an error that states "The version is already published".
2. runs check_version.ts using npm script called 'sanity'.
3. Resolves https://github.com/ansible/vscode-ansible/issues/302

### DIFF
--- a/src/tools/check_version.ts
+++ b/src/tools/check_version.ts
@@ -1,0 +1,95 @@
+import { execSync } from 'child_process';
+import axios from 'axios';
+
+/**
+ * A function to make an API call to open VSX in order to get details about the latest release of redhat.ansible extension
+ * @param currentVersion string representing the current version against which information is required
+ * @returns version of the latest release
+ */
+function makeGetRequest() {
+  return new Promise(function (resolve, reject) {
+    axios
+      .get('https://open-vsx.org/api/redhat/ansible')
+      .catch(function (error) {
+        if (error.response) {
+          // Request made and server responded
+          console.log(error.response.status);
+          resolve(error.response.status);
+        } else if (error.request) {
+          // The request was made but no response was received
+          console.log(error.request);
+          reject(error.request);
+        } else {
+          // Something happened in setting up the request that triggered an Error
+          console.log('Error', error.message);
+          reject(error.message);
+        }
+      })
+      .then((response) => {
+        if (response) {
+          const result = response.data.version;
+          resolve(result);
+        }
+      });
+  });
+}
+
+/**
+ * A function that makes a call to the makeGetRequest and checks if the curren version of redhat.ansible is released in Open VSX or not
+ * @param currentVersion string representing the current version against which information is required
+ * @returns true if current version is released, false otherwise
+ */
+
+async function releasedInOpenVSX(currentVersion: string): Promise<boolean> {
+  const responseVersion = await makeGetRequest();
+
+  return true ? responseVersion === currentVersion : false;
+}
+
+/**
+ * A function that makes a call to the makeGetRequest and checks if the curren version of redhat.ansible is released in VSCode Marketplace or not
+ * @param currentVersion string representing the current version against which information is required
+ * @returns true if current version is released, false otherwise
+ */
+
+function releasedInVSCMarketplace(currentVersion: string): boolean {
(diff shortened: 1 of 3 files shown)


src/tools/check_version.ts
```diff
@@ -0,0 +1,95 @@
+import { execSync } from 'child_process';
+import axios from 'axios';
+
+/**
+ * A function to make an API call to open VSX in order to get details about the latest release of redhat.ansible extension
+ * @param currentVersion string representing the current version against which information is required
+ * @returns version of the latest release
+ */
+function makeGetRequest() {
```
webknjaz, line 9: This name is not descriptive. Change it.

src/tools/check_version.ts
```diff
@@ -0,0 +1,95 @@
...
+  return new Promise(function (resolve, reject) {
```
webknjaz, line 10: No need to create a new promise, just make the outer function async.

src/tools/check_version.ts
```diff
@@ -0,0 +1,95 @@
...
+          // The request was made but no response was received
+          console.log(error.request);
+          reject(error.request);
+        } else {
+          // Something happened in setting up the request that triggered an Error
+          console.log('Error', error.message);
+          reject(error.message);
+        }
+      })
+      .then((response) => {
```
webknjaz, line 28: Instead of having a complex promise chain, just use async/await with try/catch.

src/tools/check_version.ts
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * A function that makes a call to the makeGetRequest and checks if the curren version of redhat.ansible is released in Open VSX or not
+ * @param currentVersion string representing the current version against which information is required
+ * @returns true if current version is released, false otherwise
+ */
+
+async function releasedInOpenVSX(currentVersion: string): Promise<boolean> {
+  const responseVersion = await makeGetRequest();
+
+  return true ? responseVersion === currentVersion : false;
```
webknjaz, line 46: This looks rather pointless. What are you doing here exactly?

src/tools/check_version.ts
```diff
@@ -0,0 +1,95 @@
...
+
+function releasedInVSCMarketplace(currentVersion: string): boolean {
+  // Grab the version of extension from vscode marketplace
+  const result = execSync(
+    'node ./node_modules/vsce/out/vsce show --json redhat.ansible'
+  ).toString();
+
+  const VSCMarketplaceVersion = JSON.parse(result).versions[0].version;
+
+  return true ? VSCMarketplaceVersion === currentVersion : false;
```
webknjaz, line 63: Same here: what is this? Are you confusing it with Python's if-else?

webknjaz (changes requested):
Also, this doesn't seem to be integrated into the CI.

src/tools/check_version.ts
priyamsahoo: Same here. Corrected.

src/tools/check_version.ts
priyamsahoo: Thanks. I confused it with Python's if-else.

src/tools/check_version.ts
priyamsahoo: I did this, in order for the http request to finish before executing the rest of the code. I didn't find a better way to make it synchronous.

src/tools/check_version.ts
```diff
@@ -0,0 +1,95 @@
...
+
+function releasedInVSCMarketplace(currentVersion: string): boolean {
+  // Grab the version of extension from vscode marketplace
+  const result = execSync(
+    'node ./node_modules/vsce/out/vsce show --json redhat.ansible'
+  ).toString();
+
+  const VSCMarketplaceVersion = JSON.parse(result).versions[0].version;
+
+  return VSCMarketplaceVersion === currentVersion;
```
webknjaz, line 63: Why don't you move the comparison outside and just return the version?

ssbarnea:
Closing it for now but hopefully @priyamsahoo will not delete his branch as we may want to resurrect it in the future.

PR #313, opened 2021-11-12
Add a guard for misleading selected lang in issues

webknjaz:
This should straighten out the understanding issues of how VS Code selects languages and the extensions for them. At least, it should improve the situation with not-really-a-bug reports.

The rendering is previewable here: https://github.com/webknjaz/vscode-ansible/blob/maintenance/issue-forms-sanity/.github/ISSUE_TEMPLATE/bug_report.yml.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -35,7 +35,7 @@ body:
         filling out this form.
 
         💡 Long story short, it is important to verify that you actually
-        use `redhat.ansible` extention for a given file.
+        use the `redhat.ansible` extension for a given file.
 
         <details>
           <summary>
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -23,6 +23,78 @@ body:
 
         [discussions]: ../discussions
 
+  - type: checkboxes
+    id: it-is-really-this-extension
+    attributes:
+      label: Sanity check
+      description: >
+        _Hold on, wait a second, **🛑 STOP!**_
+
+
+        🙏 We need you to confirm something first before proceeding with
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -23,6 +23,78 @@ body:
...
+      label: Sanity check
+      description: >
+        _Hold on, wait a second, **🛑 STOP!**_
+
+
+        🙏 We need you to confirm something first before proceeding with
+        filling out this form.
+
+        💡 Long story short, it is important to verify that you actually
+        use `redhat.ansible` extention for a given file.
```
relrod, line 38:
```suggestion
        use the `redhat.ansible` extension for a given file.
```

relrod (commented):
This is probably fine for the typical bug report. I could imagine cases where the user runs into a bug *installing* and tries to report it, and then we ask them to verify that `Ansible` is selected when they can't even install, in which case the user has to lie in order to submit a bug report, which is weird.

But for the general post-install issue, this is probably fine.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -23,6 +23,78 @@ body:
...
+
+          ![❗ VS Code Ansible language selected
+          ](https://user-images.githubusercontent.com/578543/141507613-ef7ec8eb-f75b-4f6e-8580-750678768b90.png)
+        </details>
+      options:
+      - label: >-
+          I certify that the `redhat.ansible` extension is in use and
+          the language of the document in this bug report shows
+          up as `Ansible`
+        required: true
```
webknjaz, line 96: @relrod I suppose we could make it optional or add a drop-down with an entry saying it's an install-time problem. But that wouldn't be our bug either.

ssbarnea (approved):

PR #319, opened 2021-11-16
Document the need to have an opened workspace

ssbarnea:
As we will need to release another version before we get the
linked bug fixed, we should warn users about the need to have
a workspace open.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,11 @@ This extension adds language support for Ansible to
 and [OpenVSX](https://open-vsx.org/extension/redhat/ansible) compatible editors
 by leveraging [ansible-language-server](https://github.com/ansible/ansible-language-server).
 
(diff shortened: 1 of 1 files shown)


relrod (approved):

README.md
```diff
@@ -5,6 +5,11 @@ This extension adds language support for Ansible to
 and [OpenVSX](https://open-vsx.org/extension/redhat/ansible) compatible editors
 by leveraging [ansible-language-server](https://github.com/ansible/ansible-language-server).
 
+Keep it mind that due to [#177](https://github.com/ansible/ansible-language-server/issues/117)
```
cidrblock, line 8: I think this was intended to be #117 not 177?

cidrblock (dismissed):
I also cross linked this PR to the LS repo issue #117 so we don't forget to remove this

PR #330, opened 2021-11-18
e2e Tests

priyamsahoo:
The PR contains
- the test setup for end-to-end (e2e) tesing of the extension.
- basic testing for diagnostics using ansible-lint and syntax-check.

TODO:
- [x] Structure the test fixtures in a more appropriate way.
- [x] Generalize diagnostic test script.
- [x] Modify assert statements to use non-deprecated functions.


ssbarnea:
I do not understand why you did not just add the next tests to the already existing `extest` suite. That is what runs when you execute `npm run ui-test` and already takes care about installing an isolated version of vscode and out extension. Basically you only need to add extra operations  to https://github.com/ansible/vscode-ansible/tree/main/src/ui-test

If I am correct, you should not need to modify anything inside `package.json`, no extra scripts.

src/test/testDiagnotics.test.ts
```diff
@@ -0,0 +1,96 @@
...
+describe('TEST FOR DIAGNOSTICS IN LOCAL ENVIRONMENT (without ee)', () => {
+  const docUri1 = getDocUri('diagnostics/1.yml');
+  const docUri2 = getDocUri('diagnostics/2.yml');
+
+  beforeEach(async () => {
+    await vscode.commands.executeCommand('workbench.action.closeAllEditors');
+  });
+
+  describe('Diagnostic test with ansible-lint', () => {
+    it('Test 1', async () => {
```
webknjaz, line 19: The test names should contain a meaningful continuation of `it` and probably use the word `should`. So that it'd read `it should do/result in something`.
```suggestion
    it('should result in a result', async () => {
```

src/test/helper.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export async function testDiagnostics(
+  docUri: vscode.Uri,
+  expectedDiagnostics: vscode.Diagnostic[]
+): Promise<void> {
+  const actualDiagnostics = vscode.languages.getDiagnostics(docUri);
+
+  assert.strictEqual(actualDiagnostics.length, expectedDiagnostics.length);
+
+  if (actualDiagnostics.length !== 0 && expectedDiagnostics.length !== 0) {
```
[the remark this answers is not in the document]
webknjaz: Even better. Since the check above already verifies that the lengths are equal, just check one.
```suggestion
  if (!actualDiagnostics.length) {
    return;
  }
```

src/test/helper.ts
```diff
@@ -0,0 +1,74 @@
...
+    expectedDiagnostics.forEach((expectedDiagnostic, i) => {
```
webknjaz, line 64: Don't use `Array.prototype.forEach()` here. Just use the normal `for (let i in expectedDiagnostic) {}` loop.

src/test/testDiagnotics.test.ts
ssbarnea: I do agree that "test 1" does not say anything meaningful but neither `should result in a result` ;)

priyamsahoo: I agree. I did not intend to have this message. It was only for testing purpose and then I forgot to remove from before pushing the PR.

src/test/helper.ts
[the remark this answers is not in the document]
priyamsahoo: Thanks for the suggestion !

src/test/helper.ts
priyamsahoo: Is there any specific reason for that ? 
Generally, I prefer writing this way when I do not want to break out of the loop.

src/test/testDiagnotics.test.ts
priyamsahoo: Done !

webknjaz: > I do agree that "test 1" does not say anything meaningful but neither `should result in a result` ;)

Totally, it was mostly to demonstrate how hard it is to guess what the test expectation is if the author does not provide it.

src/test/helper.ts
webknjaz: It just seems cleaner.

src/test/test.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
```
webknjaz, line 1: Do we actually need this JS counterpart of the TS file?

src/test/test.ts
```diff
@@ -0,0 +1,34 @@
+export const defaultAnsibleConfigurations = {
```
webknjaz, line 1: Why is this duplicate declaration needed?

src/test/test.ts
```diff
@@ -0,0 +1,34 @@
...
+  'executionEnvironment.image':
+    'quay.io/ansible/ansible-devtools-demo-ee:v0.1.0',
+
+  'executionEnvironment.pullPolicy': 'missing',
+  'python.activationScript': '',
+  'python.interpreterPath': 'python3',
+  'ansible.path': 'ansible',
+};
+
+// defaultAnsibleConfigurations.forEach((config) => {
```
webknjaz, line 19: Maybe don't commit this?

src/test/testRunner.ts
```diff
@@ -0,0 +1,67 @@
...
+} from 'vscode-test';
+
+async function main(): Promise<void> {
+  try {
+    const executable = await downloadAndUnzipVSCode();
+    const cliPath = resolveCliPathFromVSCodeExecutablePath(executable);
+
+    // Install the latest released redhat.ansible extension
+    const installLog = cp.execSync(
+      `"${cliPath}" --install-extension redhat.ansible --force`
```
webknjaz, line 16: Why is this necessary?

src/test/testRunner.ts
```diff
@@ -0,0 +1,67 @@
...
+    );
+    console.log(installLog.toString());
+
+    // Install the dependant extensions
```
webknjaz, line 20: Why can't this just come from the deps?

src/test/testScripts/diagnostics/testDiagnoticsAnsibleLocal.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    describe('Diagnostic test with ansible-lint', () => {
+      beforeEach(async () => {
+        await updateSettings('ansibleLint.enabled', true);
+      });
+
+      it('should complain about no task names', async () => {
+        await activate(docUri1);
+        await vscode.commands.executeCommand('workbench.action.files.save');
+
+        await sleep(2000); // Wait for the diagnostics to compute on this file
```
webknjaz, line 34: Is there a more reliable way to wait?

src/test/testScripts/diagnostics/testDiagnoticsAnsibleLocal.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+        await testDiagnostics(docUri1, []);
+      });
+
+      // eslint-disable-next-line quotes
+      it("should complain about missing 'hosts' key", async function () {
```
webknjaz, line 84: how about
```suggestion
      it('should complain about missing `hosts` key', async function () {
```

src/test/testScripts/hover/testHoverLocal.test.ts
```diff
@@ -0,0 +1,143 @@
...
+  resetDefaultSettings,
+} from '../../helper';
+
+export function testHoverLocal(): void {
+  describe('TEST FOR HOVER', () => {
+    const docUri1 = getDocUri('hover/1.yml');
+
+    beforeEach(async () => {
+      await resetDefaultSettings();
+      // await vscode.commands.executeCommand('workbench.action.closeAllEditors');
```
webknjaz, line 17: Any comment on this?

src/test/testScripts/hover/testHoverLocal.test.ts
```diff
@@ -0,0 +1,143 @@
+/* eslint-disable quotes */
```
webknjaz, line 1: Any linting exception must have reasoning behind it.

webknjaz (changes requested):
I've added some notes + add this to the CI plz.

.gitignore
```diff
@@ -5,3 +5,5 @@ out
 .python-version
 test-resources
 test-resources-oldest
+.vscode-test
+e2eTestReport/
```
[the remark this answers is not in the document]
priyamsahoo: .vscode-test has a different purpose all together. It contains the downloaded vscode package for test. Are you suggesting to put e2eTestReport inside .vscode ? If that's the case, I don't think that's a great idea.

package.json
```diff
@@ -448,7 +450,9 @@
     "webpack": "npm run clean && npm run compile && webpack --mode production --config ./webpack.config.js",
     "webpack:dev": "npm run clean && webpack --mode development --config ./webpack.config.js",
     "webpack:watch": "webpack --mode development --config ./webpack.config.js --watch",
-    "clean": "rimraf out/client out/server out/tsconfig.tsbuildinfo"
+    "clean": "rimraf out/client out/server out/tsconfig.tsbuildinfo",
+    "test-compile": "npm run clean && tsc -p ./",
+    "test:e2e": "npm run test-compile && node ./out/client/test/testRunner"
```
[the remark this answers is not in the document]
priyamsahoo: Thanks for the suggestion.

src/test/helper.ts
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Activates the redhat.ansible extension
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export async function activate(docUri: vscode.Uri): Promise<any> {
+  const extension = vscode.extensions.getExtension('redhat.ansible');
+  const activation = await extension?.activate();
+
+  try {
```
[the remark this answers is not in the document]
priyamsahoo: Basically, there are 3 statements inside that and all of them are kind of linked to each other and serve the same purpose (open the doc and set language to 'ansible'). That's the reason I have kept it that way. I don't think I can further refactor it.

src/test/helper.ts
```diff
@@ -0,0 +1,120 @@
...
+    doc = await vscode.workspace.openTextDocument(docUri);
+    editor = await vscode.window.showTextDocument(doc, {
+      preview: true,
+      preserveFocus: false,
+    });
+    await vscode.languages.setTextDocumentLanguage(doc, 'ansible');
+
+    await sleep(5000); // Wait for server activation
```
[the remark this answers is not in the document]
priyamsahoo: As of now, I don't have a better solution for this, since there's no such actionable indication when the server is initiated. Maybe, this can be discussed and a better implementation (if any), can be added in subsequent PRs.

src/test/helper.ts
```diff
@@ -0,0 +1,120 @@
+import * as vscode from 'vscode';
+import * as path from 'path';
+import { assert } from 'chai';
+
+export let doc: vscode.TextDocument;
+export let editor: vscode.TextEditor;
+
+// Default ansible configurations
+export const defaultAnsibleConfigurations = {
```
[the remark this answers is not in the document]
priyamsahoo: Working on it. I want that to be the part of a separate PR.

src/test/helper.ts
```diff
@@ -0,0 +1,120 @@
...
+  const ansibleConfiguration = vscode.workspace.getConfiguration('ansible');
+  return ansibleConfiguration.update(setting, value, false);
+}
+
+export async function resetDefaultSettings(): Promise<void> {
+  const ansibleConfiguration = vscode.workspace.getConfiguration('ansible');
+  Object.entries(defaultAnsibleConfigurations).forEach((config) => {
+    ansibleConfiguration.update(config[0], config[1], false);
+  });
+  await sleep(1000);
```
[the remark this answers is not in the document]
priyamsahoo: When the settings all reset, if we do not wait, they are not reflected in the config. This is one is tested. Resetting of configs, take some time. That's why, there's a wait statement over there.

src/test/test.js
priyamsahoo: That's a mistake, this file should not be a part of the PR altogether, will remove it.

src/test/test.ts
priyamsahoo: Same here, `test.ts` and `test.js` are not actually the part of the PR. I will remove it. Thanks.

src/test/test.ts
priyamsahoo: Same as above. The files are unnecessary. I will remove them.

src/test/testRunner.ts
priyamsahoo: When test starts for the first time and vscode is downloaded in the space, installations of the extensions need to be performed so that they can be tested. Plus `--force` flag is necessary so that the latest version of the extension is downloaded and all testings can be performed on it.

src/test/testRunner.ts
priyamsahoo: Sorry, unable to understand it. Could you please elaborate?

src/test/testScripts/diagnostics/testDiagnoticsAnsibleLocal.test.ts
priyamsahoo: As of now, I could not find a better way to do it. Maybe if we find it in future, the changes can be made in the subsequent PRs.

src/test/testScripts/hover/testHoverLocal.test.ts
priyamsahoo: I will remove the commented code. Thanks.

src/test/testScripts/diagnostics/testDiagnoticsAnsibleLocal.test.ts
priyamsahoo: Thanks for the suggestion, this seems a better representation. Will do it this way.

src/test/testScripts/hover/testHoverLocal.test.ts
priyamsahoo: Yes. There's reason behind this as well.
But seems like if implemented the way you suggested in the previous comment, that linting exception can be removed from at least this file. Thanks for the suggestion.

.gitignore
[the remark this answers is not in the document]
ssbarnea: I think we should start sticking temp folders inside `out/`. If I remember well we already have this folder and we can other things to it and avoid cluttering repository root with other folders.

I do remember even moving the test-resources folder there, maybe i forgot to remove them from gitignore or failed to push that change. The reason the same: avoiding temp folders in root level.

I am not sure if we can do the some about the `.vscode-test` folder but at least for the report we can clearly do it.

ssbarnea:
@priyamsahoo @webknjaz Please mark threads as resolved if they were resolved, it would make it easier to realise which ares still need work or not.

ganeshrn (approved):

.gitignore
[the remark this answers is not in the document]
priyamsahoo: Thanks for the suggestion. The `e2eTestReport` folder has been moved inside the `out` folder.

ssbarnea:
I tried to run `npm run test:e2e` on macos and it failed to run with multiple tests. IMHO, review process should not even start before we have a running and passing job on GHA. I pushed a change for enabling it, so we should see results soon.

.github/workflows/ci.yaml
```diff
@@ -29,31 +29,32 @@ jobs:
       fail-fast: false
       matrix:
         os:
-        - ubuntu-latest
-        - macos-11
-        - windows-latest
+          - ubuntu-latest
```
webknjaz, line 32: Plz don't add unnecessary indents to YAML sequences.

.github/workflows/ci.yaml
```diff
@@ -29,31 +29,32 @@ jobs:
...
+          - windows-latest
         npm-target:
-        - test:ui
-        - test:ui-oldest
+          # - test:ui
+          # - test:ui-oldest
+          - test:e2e
         upload-artifact:
-        - false
+          - false
```
webknjaz, line 40: Looks like most of the changes to this file must be reverted.

ssbarnea (changes requested):
Once we have the workaround for https://github.com/ansible/vscode-ansible/issues/373 added to it it should be ok.

priyamsahoo:
All the smaller parts of this PR are merged successfully and the tests are passing in the CI as well.
- PR for hover e2e-test: https://github.com/ansible/vscode-ansible/pull/370
- PR for ansible diagnostic e2e-test: https://github.com/ansible/vscode-ansible/pull/382
- PR for YAML diagnostic e2e-test: https://github.com/ansible/vscode-ansible/pull/396

PR #331, opened 2021-11-18
Make ansible/creator-ee:latest default execution environment

ssbarnea:
Especially as we are under heavy development, we should have a default
value that uses the latest creator-ee, so we avoid extra maintenance
when we update it.

This corrects https://github.com/ansible/vscode-ansible/pull/300 which used a temporary name.

Partial: https://github.com/ansible-community/devtools/issues/16

### DIFF
--- a/package.json
+++ b/package.json
@@ -170,7 +170,7 @@
         "ansible.executionEnvironment.image": {
           "scope": "resource",
           "type": "string",
-          "default": "quay.io/ansible/ansible-devtools-demo-ee:v0.1.0",
+          "default": "quay.io/ansible/creator-ee:latest",
           "description": "Specify the name of the execution environment image."
         },
         "ansible.executionEnvironment.pullPolicy": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -170,7 +170,7 @@
         "ansible.executionEnvironment.image": {
           "scope": "resource",
           "type": "string",
-          "default": "quay.io/ansible/ansible-devtools-demo-ee:v0.1.0",
+          "default": "quay.io/ansible/creator-ee:latest",
```
webknjaz, line 173: Not sure about the "latest".

ssbarnea: I do not want to have to update it in at least 3 repositories every time we release a new one. We will need to add a feature to display its versioned tag at runtime, but when it comes to defaults, using latest is the only flexible version we have. At least of a long time we will only support use of latest container. Once we release a new one, 

PS. I hope that latest meaning is about `latest` tag meaning translates to latest tagged version, not every random merge made to main branch.

webknjaz (approved):
Not sure if everyone is in agreement here but because of https://github.com/ansible/ansible-language-server/pull/132 in ALS, I'm approving this. I'd be more confident with this if somebody else approved it too.

cidrblock (approved):
+1, hopefully we should not have to rename this again, but there is a small chance

PR #337, opened 2021-11-19
Fix inline encryption of multiline strings

jeinwag:
I noticed that the inline encryption of multiline strings does not give the expected results.

This PR fixes the issue and implements proper handling of the different types of multiline blocks in YAML.


ssbarnea (commented):
If you would not mind, next time you find such a bug file a ticket before starting to work on fixing it and mention that you will try to fix it yourself. That is better for several reasons: letting others be aware of existing bug and also avoiding having two working on fixing the same bug.

I mentioned that because your PR went in while I was running the release pipeline. If I knew about that one hour earlier I could have delayed it and include the fix in 0.7.0 release.

PS. I am currently testing the code.

ssbarnea:
@jeinwag Can you please mention on clear operation where this failed before your patch and that started working correctly after your patch.

Based on my testing I seen that round-trip operation worked all time, even if the original block was a multiline one, with `|` or with `>` yaml forms. Maybe I did not run the test correctly but I would really want to ensure it before merging it.

jeinwag:
@ssbarnea sorry, I should have clarified.
En- and decryption of multiline strings does work fine, but the encrypted value is the literal string - including the multiline indicator (e.g. `|` or `>`) and indentation.

Running the following example with the value of `multiline` in encrypted vs decrypted state should make the issue obvious:
```
- name: test multiline inline encryption
  hosts: localhost
  vars:
    multiline: |
      this is a
      multiline block
  tasks:
  - name: output multiline
    debug: msg="{{ multiline }}"
```

ssbarnea (approved):
Now that I understand it better I will approve it but not merge it yet, mainly because I want to merge first the change that adds extra testing.

With extra testing it should become quite easy to add a test for this bugfix, so we would prevent regressions.

src/features/vault.ts
```diff
@@ -306,3 +341,115 @@ const getIndentationLevel = (
...
+  currentIndex: number,
+  array: string[]
+): string => {
+  if (
+    currentValue === '' ||
+    currentValue.match(/^\s/) ||
+    array[currentIndex - 1].match(/^\s/)
+  ) {
+    return `${accumulator}\n${currentValue}`;
+  } else if (accumulator.charAt(accumulator.length - 1) !== '\n') {
```
webknjaz, line 357:
```suggestion
  if (accumulator.charAt(accumulator.length - 1) !== '\n') {
```

PR #343, opened 2021-11-25
Updated readme with install and activate instructions

ganeshrn:
*  Added a gif to demonstrate how to install Red Hat
   ansible extension and activate it.
*  Remove message for ansible-language-server issue with
   file without workspace as the issue is fixed by
   PR https://github.com/ansible/ansible-language-server/pull/140


ssbarnea (changes requested):
While removal of the warning paragraph is needed, I do not think that we really need an 8MB animation about just how a user is supposed to install our extensions, a process that is identical for any vscode extension. 

I do not remember any user not being able to do that part w/o asking. The real tricky part is only about selecting the language of of the document after opening a YAML file. That would likely deserve a video of better a wizard explaining it. @webknjaz likely knows what I am referring to as he found some documentation about that subject.

Even if we do an animated gif, it should be a small one that highlights only the area about selecting language and not a full screen one.

Please do the removal of the paragraph separated so we can merge it before the release.

Keep in mind that anything we add at the top of the document will move the rest down, making less visible.

webknjaz (changes requested):
Extending what Sorin wrote, I'd like to emphasize several distinct problems:
1) Adding binaries to the repo is bad — Git is designed to be handling text well but not blobs. Both adding and removing a file from the repo adds its size to the size of the bare Git repo, we must be mindful about that.
2) If this file gets into the package, it'll enlarge the artifact the users have to download.
3) The URLs used points to a non-static location. The README for the current version will refer to the current file but when it changes or gets deleted, the old docs will be broken.

To address this, we could just upload the GIF to GitHub's CDN (by using a comment/issue/PR form) and copy-paste the URL to that immutable image.

ssbarnea:
The file size was the least problematic bit, i am more concerned about the readme becoming bloated to a point where nobody will bother reading it in full. In general i prefer avoiding videos but there are cases where you cannot easily explain things with static images (like auto-complete feature).

ganeshrn:
@ssbarnea I have reduced the file size. As discussed offline this is a temporary solution to highlight how to activate the Ansible extension. Since many people are facing this issue after installing an extension it does not work out of the box and are not aware of how to check if extension is active or not

ganeshrn:
@webknjaz 

>  To address this, we could just upload the GIF to GitHub's CDN (by using a comment/issue/PR form) and copy-paste the URL to that immutable image.

The comment is out of scope for this PR and can be addressed at all places in a separate PR.

webknjaz:
> @webknjaz 
> 
> >  To address this, we could just upload the GIF to GitHub's CDN (by using a comment/issue/PR form) and copy-paste the URL to that immutable image.
> 
> The comment is out of scope for this PR and can be addressed at all places in a separate PR.

It's not out of the scope for as long as you are going to commit a huge blob to Git. Both adding **and removing** blobs from Git make the repo bigger. So if you commit this GIF, and then remove it in a separate commit, you'll add 8x2=16Mb to the repository that would be impossible to reduce without rewriting the history with a force push.

webknjaz:
If it's 1.2Mb, then it'll be around 2.5Mb if waste.

ssbarnea:
We could enable LFS on this repo. I wonder how others are doing in that area. We could also use the wiki side repo for storing them if size is the only issue.

My main concern with fullscreen video will bring everything below it less visible. Ideally I would look for the approach were we add the popup/wizard/helper when extension is installed which would teach the user directly while keeping the readme smaller.

webknjaz:
@ssbarnea LFS is paid and requires an extra tool installed on the contributor machines which is less nice. The reason I suggested the CDN is that it's one drag-and-drop away. Once you have the URL, no need to save the comment/issue where you perform the upload.

tomaciazek:
On top of the other things mentioned by @ssbarnea and @webknjaz, the quality of the GIF could be improved. I used [ScreenToGif](https://www.screentogif.com/) to capture the feature highlights for the readme.

webknjaz:
> I used [ScreenToGif](https://www.screentogif.com/) to capture the feature highlights for the readme.

That won't work: nobody on the team uses Windows.

README.md
```diff
@@ -5,10 +5,18 @@ This extension adds language support for Ansible to
...
 
-Keep it mind that due to [ALS#117](https://github.com/ansible/ansible-language-server/issues/117)
-this extension works only when a workspace is present. If you open a standalone
-file outside a vscode workspace, you will not be able to use most of the
-features of this extension.
+## Activating Red Hat Ansible extension
+
+It is recommended to open folder containing Ansible files with VSCode workspace.
+
+![Linter support](https://github.com/ansible/ansible-language-server/raw/main/images/ansible-extension-activation.gif)
```
webknjaz, line 12: As a half-measure, I've uploaded it to GH CDN for you:
```suggestion
![Linter support](https://user-images.githubusercontent.com/578543/143657900-66789b17-ccda-4588-8945-1c936a849532.gif)
```
If you go for this and drop adding the blob from the PR, I think we could settle on the lower quality GIF for some time. The important thing is that it does not affect the Git repo.

README.md
```diff
@@ -5,10 +5,18 @@ This extension adds language support for Ansible to
...
+
+Note:
+
+* For Ansible files open in IDE window ensure the language mode is
```
webknjaz, line 16: VS Code is not an IDE
```suggestion
* For Ansible files open in an editor window ensure the language mode is
```

README.md
```diff
@@ -5,10 +5,18 @@ This extension adds language support for Ansible to
...
+  set to ``Ansible`` (bottom right of VSCode window).
```
webknjaz, line 17: Unlike RST, Markdown uses single backticks for inline code:
```suggestion
  set to `Ansible` (bottom right of the VS Code window).
```

README.md
```diff
@@ -5,10 +5,18 @@ This extension adds language support for Ansible to
...
+* The runtime status of extension should be in activate state. It can
+  be verified in the ``Extension`` window ``Runtime Status`` tab for ``Ansible`` extension.
```
webknjaz, line 19: I don't have the `Runtime Status` tab in my VS Code instance. It may be old, though.

priyamsahoo (approved):

ssbarnea:
I observed the that gif is using dithering and that is causing some undesirable artifacts on the video. Can you please regenerate it without dithering or replace it with webp maybe? Based on https://caniuse.com/webp it should be ok but I do not know if github is able to render these.

webknjaz:
> but I do not know if github is able to render these.

I'm 99% sure it can't. While I agree that GIF is suboptimal, I think it's okay to accept it as is _right now_ and file an issue to fix this later. webp would indeed be nicer.

README.md
```diff
@@ -5,6 +5,19 @@ This extension adds language support for Ansible to
 and [OpenVSX](https://open-vsx.org/extension/redhat/ansible) compatible editors
 by leveraging [ansible-language-server](https://github.com/ansible/ansible-language-server).
 
+## Activating Red Hat Ansible extension
+
+It is recommended to open folder containing Ansible files with VSCode workspace.
```
webknjaz, line 10: Grammar+consistency:
```suggestion
It is recommended to open a folder containing Ansible files with a VS Code workspace.
```

webknjaz (approved):
I'm happy with the current state for as long as we can replace the image with WebP later. It seems to be supported by GitHub and the browsers we care about per https://github.com/TomasHubelbauer/markdown-webp.

ssbarnea:
> I'm happy with the current state for as long as we can replace the image with WebP later. It seems to be supported by GitHub and the browsers we care about per [TomasHubelbauer/markdown-webp](https://github.com/TomasHubelbauer/markdown-webp).

We need to dig vscode marketplace for bugs related to this as I know for sure that they only support a subset of what markdown can. Keep in mind that the reame needs to be displayed in both vscode marketplace and eclipse one. Likely that if it works on microsoft one it will work on the other one too.

ssbarnea (approved):

webknjaz:
>  they only support a subset of what markdown can.

This shouldn't matter. From the Markdown PoV it's just an image and any converter would put whatever string you save as "src", from there, it's the browser's job to get it rendered.

ssbarnea:
That is not true, and is like that for security reasons. While md to html does work the way you say, they also have protections, so people do not inject junk as "img". There were security issues even with gif long time ago and I suspect the less common the format, the more likely to break or to be filtered.

There is another aspect related to image size. Some markdown to html formatters do read image info in order to produce better html output by including output size. Anyway, we will have to play with it to see what it works.

ssbarnea:
@ganeshrn @webknjaz I am almost sure that this change will introduce a regression on rendering the bullet list. if I remember correctly github marketplace does not render dashes and only asterisks. 

We will soon see it but that change should not have being included in this PR and merged, is something that should have being subject of a different review.

ssbarnea:
In fact I looked at few others and none is using asterisks:
* https://raw.githubusercontent.com/redhat-developer/vscode-yaml/main/CHANGELOG.md 
* https://raw.githubusercontent.com/microsoft/vscode-python/main/CHANGELOG.md
* https://raw.githubusercontent.com/microsoft/vscode-eslint/main/CHANGELOG.md

Please revert that bit!

ganeshrn:
@ssbarnea The change was added since the `npm run lint` CI job was failing. Revert the bit won't help. It needs to be fixed in the linter config.

ssbarnea:
> @ssbarnea The change was added since the `npm run lint` CI job was failing. Revert the bit won't help. It needs to be fixed in the linter config.

Let's leave it as is and we will see when we make the release. Even if it breaks is only a cosmetic issue.

PR #345, opened 2021-11-26
Increase testing timeouts

ssbarnea:


webknjaz:
Is this happening in the Windows/Mac workers or Ubuntu too? Maybe we could make timeouts different for them.

ssbarnea:
> Is this happening in the Windows/Mac workers or Ubuntu too? Maybe we could make timeouts different for them.

I did not see a pattern and I do not want to over-engineer it. If I would do something in the future would be to add a multiplier and just double it.

webknjaz:
Yeah, I've been doing the same in other projects: a multiplier for mac/win.

webknjaz (approved):

PR #352, opened 2021-12-15
[DNM]:Test diagnostics experiment

priyamsahoo:


.github/workflows/ci.yaml
```diff
@@ -29,27 +29,32 @@ jobs:
       fail-fast: false
       matrix:
         os:
-        - ubuntu-latest
-        - macos-11
-        - windows-latest
+          - ubuntu-latest
```
webknjaz, line 32: Please don't add unnecessary extra indentation to the YAML sequences.

.github/workflows/ci.yaml
```diff
@@ -128,9 +133,9 @@ jobs:
         if: ${{ !matrix.devel }}
         run: git diff --exit-code
 
-  check:  # This job does nothing and is only used for the branch protection
+  check: # This job does nothing and is only used for the branch protection
```
webknjaz, line 136: Keep two space before a comment requirement, just like in Python.
```suggestion
  check:  # This job does nothing and is only used for the branch protection
```
Plus note that this chunk of the workflow is synchronized across multiple repositories and changing it unnecessarily increases the maintenance burden because this causes merge conflicts when the updates to this chunk are distributed into the repositories that have a copy of it.

PR #354, opened 2021-12-15
Add e2e testing framework

ssbarnea:
This is not adding the effective tests, only the framework for running them, allowing us to do a better review of each test.


priyamsahoo (approved):
LGTM !

PR #356, opened 2021-12-17
Add e2e hover testing

ssbarnea:
This introduces few stable tests that were extracted from #330 which
still has problems with some of the tests, allowing us to gradually
add tests and ensure that we can run on both GHA and locally.


.github/workflows/ci.yaml
```diff
@@ -29,31 +29,32 @@ jobs:
       fail-fast: false
       matrix:
         os:
-        - ubuntu-latest
-        - macos-11
-        - windows-latest
+          - ubuntu-latest
```
webknjaz, line 32: Keep the sequences unindented.

.github/workflows/ci.yaml
```diff
@@ -111,9 +112,10 @@ jobs:
         run: npm run compile
 
       - name: Run ${{ matrix.npm-target }}
-        uses: GabrielBB/xvfb-action@v1.0
-        with:
-          run: npm run ${{ matrix.npm-target }}
+        # uses: GabrielBB/xvfb-action@v1.0
+        # with:
+        run: |
+          xvfb-run --server-args="-screen 0, 1920x1080x24" npm run ${{ matrix.npm-target }}
```
webknjaz, line 118: The action already solves this, no need for a manual command.

src/test/index.ts
```diff
@@ -21,7 +21,7 @@ export function run(): Promise<void> {
   const testsRoot = path.resolve(__dirname, '..');
 
   //   create default settings
-  resetDefaultSettings();
+  // resetDefaultSettings();
```
webknjaz, line 24: Why?

.github/workflows/ci.yaml
ssbarnea: `xvfb-action` is unmaintained and buggy and I remember finding at least two people online that stopped using it and starting to use `xvfb-run` directly, especially because they reported that the default screen was too small and caused errors running tests. 

BTW, that PR was a draft and i am not working on it. I rely on @priyamsahoo to do it, that is why I marked it as draft.  i will likely close it soon.

webknjaz: In that case, I'd rather fork the action. Its UX is a bit more pleasant + reusable.

PR #357, opened 2021-12-21
e2e: use custom user data dir

ssbarnea:
Ensures that we run in isolation from real user settings. This
also enables us to run tests without having to close our own
copy of vscode.


priyamsahoo (approved):

src/test/testRunner.ts
```diff
@@ -54,6 +56,10 @@ async function main(): Promise<void> {
       extensionDevelopmentPath,
       extensionTestsPath,
       launchArgs: [
+        // We want to avoid using developer data dir as this is likely to break
+        // testing and make its outcome very hard to reproduce acroos machines.
```
webknjaz, line 60:
```suggestion
        // testing and make its outcome very hard to reproduce across machines.
```

webknjaz (approved):

PR #366, opened 2022-01-06
Declare lextudio.restructuredtext ext as conflicting

ssbarnea:
As this extension has a serious known bug that was not addressed
for a very long time, it is better to suggest its removal.

Related: https://github.com/vscode-restructuredtext/vscode-restructuredtext/issues/286

### DIFF
--- a/src/extensionConflicts.ts
+++ b/src/extensionConflicts.ts
@@ -7,6 +7,7 @@ import { commands, Extension, extensions, window } from 'vscode';
 // A set of VSCode extension ID's that conflict with our extension
 const conflictingIDs = [
   'haaaad.ansible',
+  'lextudio.restructuredtext', // https://github.com/vscode-restructuredtext/vscode-restructuredtext/issues/286
   'sysninja.vscode-ansible-mod',
   'tomaciazek.ansible',
   'vscoss.vscode-ansible',


priyamsahoo (approved):

src/extensionConflicts.ts
```diff
@@ -7,6 +7,7 @@ import { commands, Extension, extensions, window } from 'vscode';
 // A set of VSCode extension ID's that conflict with our extension
 const conflictingIDs = [
   'haaaad.ansible',
+  'lextudio.restructuredtext', // https://github.com/vscode-restructuredtext/vscode-restructuredtext/issues/286
```
webknjaz, line 10: Side note: do we have a rule for the inline code comment spacing? I usually follow the one from Python — at least two spaces before the comment start.

webknjaz (approved):

src/extensionConflicts.ts
ssbarnea: I am not aware of one but you are welcome to dig into eslint options and if there is one check I would agree to enable it. Without it, i would find it not to worth our time.

kunaltyagi:
> As this extension has a serious known bug that was not addressed for a very long time, it is better to suggest its removal.

@ssbarnea Do you have more details on the bug?

PR #371, opened 2022-01-12
Fix the check job in GHA workflows

webknjaz:


webknjaz:
(not waiting for reviews because we've agreed to integrate this approach earlier and it's been fixed in other repos — this is probably the last one)

PR #386, opened 2022-01-17
chore: use prettier with markdown

ssbarnea:
That is enabling use of prettier with markdownfiles, so we do get surprise-reformatting when we edit .md files in our repo.

This should prevent recent problems that were seen with at least two incoming pull-requests, where prettier reformatted unrelated changes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -66,10 +66,16 @@ repos:
           (?x)^(
             package-lock.json
           )$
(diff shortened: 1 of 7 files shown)


priyamsahoo (approved):
I have always been in favor of having a prettier config file per project :)

cidrblock (approved):
This cool. TY. Consistency and automatic clean-up :)

PR #387, opened 2022-01-17
chore: use prettier with json

ssbarnea:


cidrblock (approved):
More automatic cleanup, ty

.pre-commit-config.yaml
```diff
@@ -71,7 +72,12 @@ repos:
     rev: "v2.5.1"
     hooks:
       - id: prettier
-        types: ["markdown"]
+        types_or: ["markdown", "json"]
```
webknjaz, line 75: It's nicer to have this diff-friendly:
```suggestion
        types_or:
        - markdown
        - json
```

.vscodeignore
```diff
@@ -17,3 +17,4 @@ out/**
 !out/client/extension.js
 server/**
 !out/server/src/server.js
+.pytest_cache
```
webknjaz, line 20: This doesn't seem related, submit separately.

.ncurc.json
```diff
@@ -1,6 +1,4 @@
 {
-    "upgrade": true,
-    "reject": [
-      "@types/vscode"
-    ]
+  "upgrade": true,
+  "reject": ["@types/vscode"]
```
webknjaz, line 3: Any way not to inline lists?

ssbarnea: No idea as I did not look into it. I see no problem having short lists inline and I am sure that once it will grow enough it will reformat it on multiple lines.

.pre-commit-config.yaml
ssbarnea: ditto, if formatter didn't do it, i don't care about it.

webknjaz: Well, I do — lists are naturally multi-line entries.

.ncurc.json
webknjaz: As I said above, lists are naturally multiline things, it's not like we need a compressed JSON. It's not resolved. Don't mark it as resolved arbitrarily. I think it's important to consistently have lists show up as lists.

priyamsahoo (approved):

PR #388, opened 2022-01-17
chore: enable yamllint

ssbarnea:
This enables yamllint and ensures that indentation is consistent
across our files, fixing problem where prettier would reformat the
file on save.

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -1,0 +1,7 @@
+rules:
+  document-start: disable
+  indentation:
+    level: error
+    indent-sequences: true
+ignore: |
+  .tox
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -2,8 +2,8 @@
 name: 🐛 Bug report
 description: Create a report to help us improve
 labels:
-- bug
-- new
+  - bug
+  - new
 
 body:
   - type: markdown
@@ -89,11 +89,11 @@ body:
           ](https://user-images.githubusercontent.com/578543/141507613-ef7ec8eb-f75b-4f6e-8580-750678768b90.png)
         </details>
       options:
-      - label: >-
-          I certify that the `redhat.ansible` extension is in use and
-          the language of the document in this bug report shows
-          up as `Ansible`
-        required: true
+        - label: >-
+            I certify that the `redhat.ansible` extension is in use and
+            the language of the document in this bug report shows
+            up as `Ansible`
+          required: true
 
   - type: textarea
     id: summary
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -2,8 +2,8 @@
 name: 💡 Feature request
 description: Suggest an idea for this project
 labels:
-- feature
-- new
+  - feature
+  - new
 
 body:
   - type: markdown
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -29,31 +29,31 @@ jobs:
       fail-fast: false
       matrix:
         os:
-        - ubuntu-latest
-        - macos-11
-        - windows-latest
+          - ubuntu-latest
+          - macos-11
+          - windows-latest
         npm-target:
-        - test:ui
-        - test:ui-oldest
+          - test:ui
+          - test:ui-oldest
         upload-artifact:
-        - false
+          - false
         name:
-        - false
+          - false
         include:
-        - name: lint
-          npm-target: lint
-          os: ubuntu-latest
-          upload-artifact: true
-        - name: devel
-          # this job will install latest version of ansible-language-server
-          # instead of the one mentioned in package[-lock].json
-          npm-target: test:ui
-          os: ubuntu-latest
-          devel: true
-        - name: test:e2e
-          npm-target: test:e2e
-          os: ubuntu-latest
-          experimental: true
+          - name: lint
+            npm-target: lint
+            os: ubuntu-latest
+            upload-artifact: true
+          - name: devel
+            # this job will install latest version of ansible-language-server
+            # instead of the one mentioned in package[-lock].json
+            npm-target: test:ui
+            os: ubuntu-latest
+            devel: true
+          - name: test:e2e
+            npm-target: test:e2e
+            os: ubuntu-latest
+            experimental: true
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
@@ -132,16 +132,16 @@ jobs:
         if: ${{ !matrix.devel }}
         run: git diff --exit-code
 
-  check:  # This job does nothing and is only used for the branch protection
+  check: # This job does nothing and is only used for the branch protection
     if: always()
 
     needs:
-    - build
+      - build
 
     runs-on: ubuntu-latest
 
     steps:
-    - name: Decide whether the needed jobs succeeded or failed
-      uses: re-actors/alls-green@release/v1
-      with:
-        jobs: ${{ toJSON(needs) }}
+      - name: Decide whether the needed jobs succeeded or failed
+        uses: re-actors/alls-green@release/v1
+        with:
+          jobs: ${{ toJSON(needs) }}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -58,6 +58,13 @@ repos:
         #   args: ['--autofix']
       - id: debug-statements
         language_version: python3
+  - repo: https://github.com/adrienverge/yamllint.git
+    rev: v1.26.3
+    hooks:
+      - id: yamllint
+        files: \.(yaml|yml)$
+        types: [file, yaml]
+        entry: yamllint --strict
   - repo: https://github.com/codespell-project/codespell
     rev: v2.1.0
     hooks:


webknjaz (changes requested):
Please, don't add unnecessary indents to the sequences.

.yamllint
```diff
@@ -0,0 +1,7 @@
+rules:
+  document-start: disable
+  indentation:
+    level: error
+    indent-sequences: true
```
webknjaz, line 5: This must be
```suggestion
    indent-sequences: false
```

ssbarnea:
@webknjaz That is only imposing Ansible own standards, which is to indent sequences. In fact these files **already contained inconsistent sequences inside the same file**.

Look at PR template where labels sequence was not indented but body sequence was. Similar inside ci.yml where there was fix of the two. Basically you cannot say that is switching an indentation type with another, when in fact is consolidating it.

Before this change, editing a file in vscode would have introduced undesired indentation. At least @priyamsahoo was affected by this. In fact there are two extensions that can reformat `yaml` files when in vscode and both of them are behaving the same way (vscode-yaml and prettier).

I had a longer talk with @cidrblock today regarding which path we should take on this and some of the ideas that went out were:
* having different indentation rules based on where the yaml files are located on disk would only confuse users and force us to add a lot of complexity to our checks, shortly does not worth our time
* having ability to auto-reformat files from incoming pull-request is a valuable as almost nobody enjoys reformatting code to match a linter expectation. Having prettier auto-fix that on file save (or pre-commit) saves a lot of human time.
* we are part of Ansible team, so diverging from Ansible standards would be... not wise

Obviously that we will wait for input from the entire team on this one.

webknjaz:
I asked some time ago, among the core team members, and most preferred no indent. So no, there's no such thing as an Ansible standard.

ssbarnea:
There are serious problems especially with blocks when not using indentation. See https://docs.ansible.com/ansible/latest/user_guide/playbooks_blocks.html

When using blocks lack of indentation can often cause keys like `when:` to be wrongly applied to the wrong level, instead of applying to a block to apply to the last task inside the block, or the opposite. That is why very long time ago the indentation went towards always indenting sequences.

The fact that documentation uses that is not a random mistake.

webknjaz:
The documentation represents something that tech writers chose to use, that's it. There is no official guidelines from the engineering.

cidrblock:
Although I tend to not indent when making a sequence, I find both approaches legible.  Why I do appreciate is tools that ensure consistency in a repo without any manual work or intervention.

I'm a 0 on indent vs. not

+1 for a precommit something that ensures consistency

priyamsahoo (approved):
I am always in favor of consistency, irrespective of my opinion on indentations.

tomaciazek:
> Please, don't add unnecessary indents to the sequences.

Just because something is optional, does not mean that it's unnecessary. Would you write code in Java/C++/C#/JavaScript without indentation (just because curly braces are enough)? I think that indentation of sequences provides much better legibility, especially with more complex cases, such as Ansible blocks.

tomaciazek (approved):

ganeshrn (approved):

cidrblock (approved):

webknjaz:
> Just because something is optional, does not mean that it's unnecessary. Would you write code in Java/C++/C#/JavaScript without indentation (just because curly braces are enough)?

When I see an indent, I expect it to be a mapping. Sequences don't need an extra bit of syntax (both indent and a dash) — there should be one distinction, not multiple.

tomaciazek:
> > Just because something is optional, does not mean that it's unnecessary. Would you write code in Java/C++/C#/JavaScript without indentation (just because curly braces are enough)?
> 
> When I see an indent, I expect it to be a mapping. Sequences don't need an extra bit of syntax (both indent and a dash) — there should be one distinction, not multiple. Besides, too much indentation contributes to overly long lines.

Is `a` a mapping:
```yaml
- a:
  - b
  - c
```
? No. Can you skip the indent? No. Hence, you can't possibly always expect an indent to be a mapping😉

Possibility to skip indentation on *some* sequences is inconsistent in itself, to say the least. In my opinion, the fact that creators of YAML made it possible did more harm than good, causing debates such as this. Sometimes less choice is better :)

Also, the indent in case of sequences is not an extra bit of syntax - it's a bit of formatting.

webknjaz:
> ? No. Can you skip the indent? No. Hence, you can't possibly always expect an indent to be a mappingwink

But the indent is counted since the position of `a:` so there's no indent :shrug:

webknjaz:
> Sometimes less choice is better :)

Sounds like it's one step away from dictatorship. If the "no-choice" option is wrong and there's no way to fix it — is it good or bad? Let the history lessons be the answer.

webknjaz:
> Possibility to skip indentation on _some_ sequences is inconsistent in itself, to say the least.

Syntax-wise, it's not skipping in some because of the place the indentation counter starts.

tomaciazek:
> > Sometimes less choice is better :)
> 
> Sounds like it's one step away from dictatorship. If the "no-choice" option is wrong and there's no way to fix it — is it good or bad? Let the history lessons be the answer.

That got political and escalated pretty quickly :P I'm by no means in favor of dictatorship and let's drop it at that.

tomaciazek:
> > Possibility to skip indentation on _some_ sequences is inconsistent in itself, to say the least.
> 
> Syntax-wise, it's not skipping in some because of the place the indentation counter starts.

It's right in the specification: https://yaml.org/spec/1.2.2/
Just under `Example 8.21. Block Scalar Nodes`:
```
Since people perceive the “-” indicator as indentation, nested block sequences may be indented by one less space to compensate, except, of course, if nested inside another block sequence ([BLOCK-OUT context] versus [BLOCK-IN context]).
```

webknjaz:
> That got political and escalated pretty quickly :P I'm by no means in favor of dictatorship and let's drop it at that.

Blame the Godwin's Law :) The internet is filled with rubbish references of disasters of the past (it gets me every time I see people jokingly "implementing" the stalin sort, for example). Besides, everything in the world is political.
FWIW I stand by the metaphor because this is how such cases _feel_, even when applied to software — and for me, it's a way to express the degree of how strongly I feel about this (I didn't mean to imply that you may have been a dictatorship supporter, I'm sorry if this upset you).
It's one thing if people choose a tool that doesn't have options and is completely different if it's forced upon them.

As for the yamllint configuration ­— it supports an option of per-file consistency. I could live with Ansible files indented but it's hard for me to accept this in other places. One of the reasons is that some of the configs get synchronized from other repositories and having the same indentation across projects certainly helps minify the merge conflicts in diffs.

tomaciazek:
> Blame the Godwin's Law :) The internet is filled with rubbish references of disasters of the past (it gets me every time I see people jokingly "implementing" the stalin sort, for example). Besides, everything in the world is political. FWIW I stand by the metaphor because this is how such cases _feel_, even when applied to software — and for me, it's a way to express the degree of how strongly I feel about this (I didn't mean to imply that you may have been a dictatorship supporter, I'm sorry if this upset you). It's one thing if people choose a tool that doesn't have options and is completely different if it's forced upon them.
> 
> As for the yamllint configuration ­— it supports an option of per-file consistency. I could live with Ansible files indented but it's hard for me to accept this in other places. One of the reasons is that some of the configs get synchronized from other repositories and having the same indentation across projects certainly helps minify the merge conflicts in diffs.

Don't worry, I didn't get upset, nor feel offended - I just didn't want this discussion to get derailed into a political debate.

As for those synchronized config files: which ones are these? How many are there?

webknjaz:
> As for those synchronized config files: which ones are these? How many are there?

Mostly top-level configs and things under `.github/` (not only YAML) + other config-like things (e.g. the Sphinx's `conf.py` etc.). Unfortunately, this synchronization is sporadic and manual right now but I'm hoping to adopt a technique of doing a `git merge` similar to what https://github.com/jaraco/skeleton provides (Jason actually uses it on hundreds of the repos, pretty fascinating stuff), except I want it to be a bit more flexible/composable/distributed. That is when I finally get to do this (automating it on the scale is non-trivial).
There are maintenance techniques that allow reducing merge conflicts that would only benefit from similar formatting.

tomaciazek:
> Mostly top-level configs and things under `.github/` (not only YAML) + other config-like things (e.g. the Sphinx's `conf.py` etc.). Unfortunately, this synchronization is sporadic and manual right now but I'm hoping to adopt a technique of doing a `git merge` similar to what https://github.com/jaraco/skeleton provides (Jason actually uses it on hundreds of the repos, pretty fascinating stuff), except I want it to be a bit more flexible/composable/distributed. That is when I finally get to do this (automating it on the scale is non-trivial). There are maintenance techniques that allow reducing merge conflicts that would only benefit from similar formatting.

Well, that doesn't sound like that many YAML files, and I suppose the synchronization process could take care of the indentation. As for the per-file consistency, I fear that it would not improve the consistency much. Even if some less or more strict guidelines are set, more often than not we'd have PRs with change requests just because of wrong indentation, and that's just counterproductive.

tomaciazek:
Another reason for which I stopped using un-indented sequences is issues with automatic code-folding in VS Code:
image

ganeshrn:
When it comes to indentation and code formatting there will always to conflicting opinions :-).
Considering the number of project maintainers we have in favour to have this PR in we do have a clear majority. I will go ahead and merge this PR. If in future we as a team collectively realise it is not the right way to go we can always revert the change.

PR #389, opened 2022-01-18
chore: improve vscode in-repo settings

ssbarnea:
- avoid reformatting yaml on save until we decide on #388
- avoid displaying temp folders or searching in them


priyamsahoo (approved):
LGTM !

PR #392, opened 2022-01-18
experiment: use prettier with typescript

ssbarnea:
Experiment, not ready for review. Findings so far:

* it seems that both `prettier` and `eslint` to prefer double quotes in javascate in their default settings, but our repository was configured to prefer single ones
* the above customization seems to cause a conflict between the two tools where eslint is trying to add single quotes on a string where it should do not that like: `"That's cool"`, where the logical expectation that use of double is normal as the contained string already has a quote inside.  Prettier seems to know to prefer avoiding escaping when possible.
* I was not able to find a way to change eslint behavior
* Out of curiosity I also tried to remove the custom rule about quotes on both tools and it seems that in that case they do not conflict.
* I also had to remove the custom indent rule on eslint because it was in conflict with prettier config.

I hope that @tomaciazek will be able to provide some insights regarding why the single quotes were added to the config as I suspect that came from the language server extension.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -15,8 +15,6 @@
   "plugins": ["@typescript-eslint"],
   "rules": {
     "eqeqeq": ["error", "smart"],
-    "prefer-template": "warn",
-    "quotes": ["error", "single"],
-    "indent": ["error", 2]
+    "prefer-template": "warn"
   }
 }
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -80,7 +80,7 @@ repos:
     rev: "v2.5.1"
     hooks:
       - id: prettier
-        types_or: ["markdown", "json"]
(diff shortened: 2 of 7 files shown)


tomaciazek:
> I hope that @tomaciazek will be able to provide some insights regarding why the single quotes were added to the config as I suspect that came from the language server extension.

AFAIR, I did that because single quotes were prevalent in the code examples (such as [VS Code Language Server Extension Guide](https://code.visualstudio.com/api/language-extensions/language-server-extension-guide)) I used as reference. It used to work for me quite well.

As for the indent rule, there are many more configuration options: https://eslint.org/docs/rules/indent; perhaps it's a matter of tweaking it. I wouldn't just disable it and rely just on Prettier, unless we incorporate Prettier (as a sort of a linter) in the CI checks, but I don't think that this is a good idea.

ssbarnea:
We will do it as part of #407

PR #396, opened 2022-01-21
e2e-test: add yaml diagnostic tests

priyamsahoo:
- The PR contains the necessary test scripts and fixtures for testing YAML diagnostic features of the `redhat.ansible` extension.
- The PR is a smaller part of https://github.com/ansible/vscode-ansible/pull/330 that contains only a single part in order to dig into the CI failures.


src/test/testFixtures/diagnostics/yaml/1.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name:
+  hosts: localhost
+  tasks:
+    - debug: msg="foo: bar"
```
ssbarnea, line 5: @priyamsahoo based on what I read from linter, I suspect you forgot to add some quotes on that string?

ssbarnea:
> Can we move the test folder out of the src folder? We can probably raise an issue to track it and can be fixed in another PR.

I wondered about the same thing myself but I am not yet sure which is the best practice for test folders for typescript projects. I see some benefits of having them in src, at least as long they do not endup in final package. Probably looking at default search location of various tools we should find the pattern to follow. No need to file bug, some things a better just to do instead of adding them to technical debt ;)

ganeshrn:
> ools we should find the pattern to follow. No need to file bug, some things a better just to do instead of adding them to technical debt ;)

The [vscode-yaml](https://github.com/redhat-developer/vscode-yaml) has `test` outside the `src` folder. If @priyamsahoo thinks my comment is not within the scope of PR we can raise an issue for it. I don't want to block the PR since test folder is already within the src folder in the main branch.

ssbarnea:
> > ools we should find the pattern to follow. No need to file bug, some things a better just to do instead of adding them to technical debt ;)
> 
> The [vscode-yaml](https://github.com/redhat-developer/vscode-yaml) has `test` outside the `src` folder. If @priyamsahoo thinks my comment is not within the scope of PR we can raise an issue for it. I don't want to block the PR since test folder is already within the src folder in the main branch.

I checked and mocha default settings are to look in `test/`, see https://mochajs.org/#the-test-directory -- so we will adopt this layout, but not in this change. ALS is already using this layout!

src/test/testFixtures/diagnostics/yaml/1.yml
priyamsahoo: Actually, I want to have this like this so that I validate the diagnostic messages that I get from the extension. That's the purpose of this particular test.

But the problem is, even if I disable yamllint for this file, the I do not get valid message from `ansible-playbook --syntax-check`

priyamsahoo:
@ganeshrn, @ssbarnea, I believe the restructuring may be a good idea but it is not in the scope of this PR. We can have it in a separate PR (if we decide to do so).

priyamsahoo:
> @ganeshrn, @ssbarnea, I believe the restructuring may be a good idea but it is not in the scope of this PR. We can have it in a separate PR (if we decide to do so).

The test dir has been moved to the standard location in the following PR: https://github.com/ansible/vscode-ansible/pull/397

ssbarnea:
Tested locally and `should return no diagnostics` failed but the other passed.

ssbarnea (approved):

ssbarnea (approved):

ssbarnea:
I run it 3 more times and did not see that error so GREEN!

PR #398, opened 2022-01-22
Recommend mocha-test-adapter vscode extension to contributors

ssbarnea:
Matches configuration from https://github.com/ansible/ansible-language-server/blob/main/.vscode/extensions.json#L6 as since #397 we started supporting mocha test runner even with this repository.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,7 +1,8 @@
 {
   "recommendations": [
-    "esbenp.prettier-vscode",
+    "DavidAnson.vscode-markdownlint",
     "dbaeumer.vscode-eslint",
-    "DavidAnson.vscode-markdownlint"
+    "esbenp.prettier-vscode",
+    "hbenl.vscode-mocha-test-adapter"
   ]
 }


.vscode/extensions.json
```diff
@@ -2,6 +2,7 @@
   "recommendations": [
     "esbenp.prettier-vscode",
     "dbaeumer.vscode-eslint",
-    "DavidAnson.vscode-markdownlint"
+    "DavidAnson.vscode-markdownlint",
+    "hbenl.vscode-mocha-test-adapter"
```
webknjaz, line 6: Does this file support trailing commas, by the way?

webknjaz (commented):
Also, it would be helpful to see the motivation, any PR description at all. Without this, I don't see a way to figure out that adding it is justified (or not).

.vscode/extensions.json
ssbarnea: Not sure but I will keep exiting. I had some chat with @tomaciazek few days ago regarding running and debugging tests and he was already using the same extension. I found that after I had some bad experience with two others. @priyamsahoo also knows as we has a pair-programming session mid week to narrow down testing issue with ALS. 

I would count that as very useful tool. Also recommandation does not force user to install it. Try, it works best on ALS but I will do some follow-ups to make it work here. The issues are not from extension but the fact that only one test suite is using mocha "by the book", the other one is more problematic.

ssbarnea: BTW, see https://github.com/ansible/ansible-language-server/blob/main/.vscode/extensions.json#L6

webknjaz:
> are already using this extension to run and debug tests on ALS project

Could you convert this into something that would be informative in a Git commit message?

.vscode/extensions.json
ssbarnea: Done. Feel free to tune it if you have extra time. I am sure everyone from the team will understand it.

webknjaz: Looks like you didn't do it after all.

PR #399, opened 2022-01-22
Add retries for extension finding

ssbarnea:
As we still see on CI the test that is searching
for the extension as failing, we add some retries,
hoping that is a timing issue.


ssbarnea:
Done 5 reruns and no failure seen. I guess that time we nailed it, it was not much the timeout but likely more of a timing issue where the search result missed to return the recently installed extension, doing retries apparently sorted it.  Before that it was almost a 50/50 chance for one of the tests to fail, more likely on macos or windows than on linux.

PR #407, opened 2022-01-28
Reformat typescript using prettier

ssbarnea:
That enables typescript on pre-commit prettier run as currently almost
any attempt to edit .ts files using vscode would introduce undesired
reformatting. That is because we recommend prettier extension which
auto-reformats.

The only manually made change is the one line in pre-commit config,
the rest is just pre-commit itself.

Sister change: https://github.com/ansible/ansible-language-server/pull/196
Related: https://github.com/ansible-community/devtools/issues/28

### DIFF
--- a/src/features/runner.ts
+++ b/src/features/runner.ts
@@ -1,6 +1,5 @@
 /* "stdlib" */
-import * as vscode from 'vscode';
-
+import * as vscode from "vscode";
 
 /**
  * A set of commands and context menu items for running Ansible playbooks using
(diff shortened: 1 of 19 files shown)


src/features/runner.ts
```diff
@@ -1,7 +1,6 @@
 /* "stdlib" */
 import * as vscode from 'vscode';
 
-
```
webknjaz, line 4: Keep a two-line separator after the imports section.

src/features/runner.ts
```diff
@@ -14,33 +13,33 @@ export class AnsiblePlaybookRunProvider {
...
+   * Register a set of command callbacks for executing Ansible playbooks
+   * within VS Code.
+   */
   private configureCommands() {
     this.vsCodeExtCtx.subscriptions.push(
       vscode.commands.registerCommand(
         'extension.ansible-playbook.run',
-        fileObj => this.makeCmdRunner()(fileObj),
-      ),
+        (fileObj) => this.makeCmdRunner()(fileObj)
```
webknjaz, line 23: The braces only add clutter, don't add them unnecessarily.

src/features/runner.ts
```diff
@@ -51,106 +50,101 @@ export class AnsiblePlaybookRunProvider {
...
       }
       console.debug(
         `Got a request to run ansible-${
-          useNavigator
-            ? 'navigator'
-            : 'playbook'
+          useNavigator ? 'navigator' : 'playbook'
         } against`,
-        playbookFsPath,
+        playbookFsPath
```
webknjaz, line 62: Always keep trailing commas.

webknjaz (commented):
I've identified a number of highly undesirable changes below. The config probably needs fixing before it's acceptable.

webknjaz (changes requested):
(missed a button in the previous review)

src/features/runner.ts
ssbarnea: Irrelevant because all changes are made by prettier, on save. There is nothing configurable about it.

priyamsahoo (approved):
LGTM

src/features/runner.ts
ssbarnea: Watch https://github.com/ansible/vscode-ansible/pull/407 for an answer. Outside our control.

ssbarnea:
There is on known bug https://github.com/prettier/prettier/issues/7976 that adds some extra parentheses but we can live with it, like we already did so far. Thus is why it was not fixes for a long time.

PR #421, opened 2022-02-03
ci: improve mocha defaults

ssbarnea:
- ensure mocha output is colored
- ensure we do not produce font assets on report (cdn) - saving 2.8mb/archive


priyamsahoo (approved):
Thanks for the enhancement.

PR #422, opened 2022-02-03
ci: ensure package.json commands are portable

ssbarnea:
Avoid using `&&`, `||`, `true` in npm scripts as they do not work on Windows and use `shelljs` for instead. This should also allow use to document better the purpose of each command we run.

While at this moment we cannot effectively run the tests on Windows GHA runners, other commands should work fine, compile, prepare, package,...


ganeshrn (commented):
Does this change work on windows with WSL2 enabled?

PR #425, opened 2022-02-08
Display error messages received from the language server

ssbarnea:
Depends-On: https://github.com/ansible/ansible-language-server/pull/195


src/extension.ts
```diff
@@ -60,6 +60,9 @@ export function activate(context: ExtensionContext): void {
 
   // start the client and the server
   client.start();
+  client.onNotification('ansible/showMessage', async (message) => {
```
ganeshrn, line 63: The notification message send from the server is `ansible/errorMessage`  whereas at this line the message checked in `ansible/showMessage`. Is this expected? Also if the server sends `ansible/showMessage` notification it will be displayed as an error message in the window which doesn't seem right. The logic can be updated to reflect below mapping?
 
`ansible/errorMessage` -> `window.showErrorMessage`
`ansible/showMessage` -> `window.showInformationMessage`

PR #444, opened 2022-03-03
Prepare fixtures for ansible-lint v6

ssbarnea:
As ansible-lint v6 makes fqcn-builtins required, this small change
ensures newer version will not break our tests.


priyamsahoo (approved):
Thanks the change :)

PR #526, opened 2022-06-14
Replace npm with yarn as package manager

ssbarnea:
To test this change run once `task clean` after checking it out as this will clean all non git tracked files.

Blocked-By: https://github.com/microsoft/vscode-vsce/pull/743


ssbarnea:
I will merge as is as the other failures are unrelated (already present) and will require different fixes (will do in follow-up).

PR #568, opened 2022-07-26
Improve e2e test timings

priyamsahoo:
This PR:
- decreases sleep timing between diagnostic tests.
- decreases sleep timing during re-initialization of the server.


ssbarnea (changes requested):
Sorry but that is not the solution for our testing issue as it will generate lots of random test failures due to "only sametime slower".

The real problem is that we should have never used sleeps and test the outcome in a loop that has a timeout, that is very different than having a sleep and checking the result after the sleep.

Based on my testing, once we implement that, we should see the test running ~8-10x faster in average.

priyamsahoo:
@ssbarnea, I would suggest keeping the issue open, but for the time being, we can merge this change as this PR improves the test timings significantly, both, in local jobs and the CI jobs.

We can look at the real issue when we are free with the works with the work that we are currently doing. It is not blocking anything currenlty.

ssbarnea (approved):

ssbarnea:
> @ssbarnea, I would suggest keeping the issue open, but for the time being, we can merge this change as this PR improves the test timings significantly, both, in local jobs and the CI jobs.
> 
> We can look at the real issue when we are free with the works with the work that we are currently doing. It is not blocking anything currently.

Yep. Approved and activated auto-merge. If it happens to pass, it will get merged, if not, probably is a sign that reducing might not be best idea. Our machines are always faster than GHA, so we should be fine.

Issue #582, opened 2022-08-04
Enable auto-switching between yaml and ansible language


ssbarnea:
### Problem

As one of the most common confusions with vscode extension is the fact that vscode often fails to decide which document language to use between generic `yaml` and `ansible` language. That is because at this moment only filename patterns are use to determine if a file uses one language or another. Once user manually change the language the IDE remembers the selection, but our goal is to avoid the need for this manual selection.

With @ganeshrn help, we identified several existing extensions that hook into document open event and decide to change the language of the document based on the file content, such the shebang line for example:
* https://marketplace.visualstudio.com/items?itemName=davidhewitt.shebang-language-associator (updated 2022)
* https://marketplace.visualstudio.com/items?itemName=ryzngard.vscode-header-source 
(updated 2019)

### Solution

We need to implement the same hook for document open even and when the current document is using either `yaml` or `ansible`, use a detection function that is guessing if the file is a playbook or a taskfile.

We currently have the `isPlaybook` function but we need a new one named `detectDocumentLanguage` which receives the open document and returns the name of the vscode language that vscode should use for it.

The function should return `null` if we were not able to make a decision. For example if current document is not YAML or the YAML loading fails, we are expected to return null, as we cannot make a decision.

Once the document is loaded, the logic should be:
- if current document language is not already `ansible` or `yaml`, return `null`.
- if document is `null`, empty list, string or empty mapping, return `null` (inconclusive)
- if document is a non empty mapping, return `yaml`
- if document is a non empty sequence, check if the first element is a dictionary having either one of the following keys: `name`, `hosts`, `import_playbook`, return `ansible`.
- return `null` as fallback, as we do not want to auto-switch document type if we are not confident enough.

### Additional context

Please look at current extensions that do this switching as all of them are open source and the implementation for document open is quite easy.

This feature does not require us to touch ansible-language-server project, as this hook is specific to vscode client.

bcoca:
#4, don't check for name, it is both not required and not deterministic of being a playbook

A problem does arise when trying to detect task files, like a role's tasks/main.yml, it is still a list, but the only requirement for a task is to have an action, which can be `action`, `local_action` or `module/action plugin name`.

ssbarnea:
In fact we do not have a problem with tasks because we always had a requirement for tasks file to be under tasks/ folder. That was the easy one.... unless user made the unfortunate decision not to put them there.

bcoca:
[...]

PR #600, opened 2022-08-23
Add feature to dynamically associate yaml files to `ansible` language

priyamsahoo:
This PR is the partial implementation of #582 and, in particular, this comment: https://github.com/ansible/vscode-ansible/issues/582#issuecomment-1206023082.

The following features are added in this PR:
1. Set language as ansible for all the yaml files under `/playbooks` dir.
2. Set language as ansible for all the files with the following double extension: `.ansible.yml` or `.ansible.yaml`.
3. Set language as ansible if modelines is present at the top of a yaml file: `# code: language=ansible.`
4. Set language as ansible for all the notable yaml names recognized by ansible. E.g.: `site.yml` or  `*site.yaml`
5. Set language as ansible for all the yaml files having `playbook` in their filename: `*playbook*.yml` or  `*playbook*.yaml`
6. Rest all the .yml, or .yaml files will remain yaml by default unless the user explicitly changes the language to ansible.


examples/playbook_with_double_ext.ansible.yml
```diff
@@ -0,0 +1,5 @@
+- hosts: all
```
ssbarnea, line 1: I you use `playbook` keyword in filename, it will still use the old glob pattern. Maybe it would better to use something that is not already included in the glob patterns like `play_` or `pb_`. It would make much easier to test the change in behavior, as we would know that these files are not expected to be picked as ansible by default.

package.json
```diff
@@ -289,8 +290,18 @@
...
+          ".ansible.yaml"
+        ],
+        "filenamePatterns": [
+          "**/playbooks/*.yml",
+          "**/playbooks/*.yaml"
+        ],
+        "filenames": [
+          "site.yml",
+          "site.yaml",
+          "playbook.yml",
```
ssbarnea, line 303: In linter we use `*playbook*/yml` pattern as being a little bit more flexible. See https://github.com/ansible/ansible-lint/blob/main/src/ansiblelint/config.py#L31-L32

priyamsahoo: Thanks for the suggestion. I can include this!

examples/playbook_with_double_ext.ansible.yml
priyamsahoo: Makes sense! Thanks.

priyamsahoo: updated.

package.json
priyamsahoo: updated.

PR #648, opened 2022-10-06
Client-Disable-Diagnostics-Settings

priyamsahoo:


ssbarnea (approved):

package.json
```diff
@@ -143,41 +143,85 @@
...
+          "default": "ansible-navigator",
+          "description": "%configuration.navigate.executablePath%",
+          "scope": "machine-overridable",
+          "type": "string",
+          "order": 5
+        },
+        "ansible.validation.enabled": {
+          "type": "boolean",
+          "default": true,
+          "description": "Toggles the validation provider.",
```
ganeshrn, line 195: Please add more description here. For eg:
`Toggles the validation provider. If the enabled it will run all the validation providers that are individually enabled, for example if  ansible.validation.lint.enabled is `true` it will run `ansible-lint`. It none of the provider is enabled it runs `ansible-playbook --syntax-check` to provide basic validations.

ssbarnea: That proposed description looks to long to me. I might clutter the configuration screen.

But I agree that original one, should be updated, let's find something in between.

priyamsahoo: Here is the PR addressing this: https://github.com/ansible/vscode-ansible/pull/649

PR #649, opened 2022-10-06
Improve description of configuration options

priyamsahoo:


package.json
```diff
@@ -192,7 +192,7 @@
         "ansible.validation.enabled": {
           "type": "boolean",
           "default": true,
-          "description": "Toggles the validation provider.",
+          "description": "Toggles the validation provider. If the enabled it will run all the validation providers that are individually enabled, for example if 'ansible.validation.lint.enabled' is 'true' it will run ansible-lint. If none of the provider is enabled it runs ansible-playbook --syntax-check to provide basic validations.",
```
ganeshrn, line 195: `s/Toggles the validation provider/Toggles all  the validation providers`

priyamsahoo: Looks better. Updated!

ganeshrn (approved):

package.json
ssbarnea: Screenshots! The crowd demands screenshots!

ssbarnea (changes requested):
I think we can do better, shorter and easier to understand.



I propose these descriptions for the two options:
* Uncheck in order to disable entire validation support.
* When checked, use `ansible-lint` for validation, otherwise only very basic validation with be done via `ansible-playbook --syntax-check` on playbook files.

priyamsahoo:
> I think we can do better, shorter and easier to understand.
> 
> 
> 
> I propose these descriptions for the two options:
> 
> * Uncheck in order to disable entire validation support.
> * If checked, use `ansible-lint`, otherwise use basic results from `ansible-playbook --syntax-check`.

This also looks good. WDYT @ganeshrn ?

ganeshrn:
> > I think we can do better, shorter and easier to understand.
> > 
> > I propose these descriptions for the two options:
> > 
> > * Uncheck in order to disable entire validation support.
> > * If checked, use `ansible-lint`, otherwise use basic results from `ansible-playbook --syntax-check`.
> 
> This also looks good. WDYT @ganeshrn ?

The setting is a boolean value and can be configured by setting `true` or `false` in `settings.json` so `check` and `uncheck` might be confusing. Also for second line ansible-lint will run only if it is enabled individually which is not reflect in the short description.

I don't get what's the issue with multi-line description. It's better to be clear of what a setting will do rather than trying to keep it short and ambiguous.

Below are the mulitine description for `typescript` extensions
Screen Shot 2022-10-10 at 1 18 34 PM



How about below 3 line description
```
If enabled run all the validation providers that are individually enabled, for example if 'ansible.validation.lint.enabled' is 'true' run ansible-lint. If none of the provider is enabled it runs ansible-playbook --syntax-check to provide basic validations. If disabled there will be no validations done at all.
```

priyamsahoo:
This is how it looks now:
image

priyamsahoo:
> > This is how it looks now: image
> 
> This is how it looks in my VSCode :-) Screen Shot 2022-10-10 at 2 02 03 PM

I think you haven't updated the branch> There's a line break, and F in 'for example' is capitalized.

ganeshrn:
> @ganeshrn, I think you haven't updated the branch. There's a line break, and F in 'for example' is capitalized.

Looks like there is a difference between the two strings. I am using a`,` any you are using `.`

priyamsahoo:
> > @ganeshrn, I think you haven't updated the branch. There's a line break, and F in 'for example' is capitalized.
> 
> Looks like there is a difference between the two strings. I am using a`,` any you are using `.`

Yes. I formatted it a little.

ssbarnea (approved):

PR #683, opened 2022-11-16
Enables reuse of existing ansible terminals

egnirra:
When using vscode commands to run ansible-playbook or ansible-navigator as seen in this image:
image

It will create a new terminal for each time you run the command.
image

Perhaps this is as expected, you want to run multiple playbooks via vscode at the same time, but I would argue it's not the best experience for most use cases.

Another way could be to have an option to enable/disable this behavior?

### DIFF
--- a/.config/requirements.txt
+++ b/.config/requirements.txt
@@ -1,6 +1,6 @@
-ansible-compat==2.2.1
+ansible-compat==2.2.4
 ansible-core==2.13.5
-ansible-lint==6.8.2
+ansible-lint==6.8.6
 attrs==21.4.0
 black==22.10.0
 bracex==2.3.post1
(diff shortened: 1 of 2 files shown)


priyamsahoo:
@egnirra, this looks good, but as you mentioned, having multiple terminals opened allows the user to run multiple playbooks simultaneously. Could you try to extend this feature to allow users to choose the way they wish to run it?

ganeshrn (approved):
Thanks for your contribution!

ssbarnea (approved):

priyamsahoo (approved):

egnirra:
If I have some time over I will look in to adding the configuration option, can do another PR if you want.

Thanks for the input!

ssbarnea:
@egnirra Take a look on how we implemented the other options, it should be easy for you to add it as configurable.

And, thanks for raising the PR!

PR #689, opened 2022-11-19
Reuse terminal setting

egnirra:
Added support to enable/disable the reuse terminal setting added in #683


priyamsahoo (approved):
Thanks for working on this. LGTM!

Issue #684, opened 2022-11-17
Ansible.cfg file location


mscreations:
### Summary

I am using a fresh Ubuntu 22.04 WSL2 install and the plugin is not finding my ansible.cfg file that is installed by default in `/etc/ansible/ansible.cfg`. 

Looking at the source [ansibleCfg.ts](https://github.com/ansible/vscode-ansible/blob/main/src/features/utils/ansibleCfg.ts#L38) file on line 38, it references `~/.ansible.cfg` and `/etc/ansible.cfg`, but not the default location of `/etc/ansible/ansible.cfg`.

Can this be added?

### Extension version

1.0.90

### VS Code version

1.73.1

### Ansible Version

```shell
ansible [core 2.13.6]
  config file = /etc/ansible.cfg
  configured module search path = ['/home/jon/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /home/jon/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.10.6 (main, Nov  2 2022, 18:53:38) [GCC 11.3.0]
  jinja version = 3.0.3
  libyaml = True
```

### OS / Environment

Windows 11 22H2

PR #691, opened 2022-11-21
Update search paths for ansible.cfg based on the documentation

priyamsahoo:
The extension will process the below list to order to locate the ansible configuration file and use the first file found:
-	ANSIBLE_CONFIG (an environment variable)
- ansible.cfg (in the current directory)
- .ansible.cfg (in the home directory)
- /etc/ansible/ansible.cfg

*source: https://docs.ansible.com/ansible/2.4/intro_configuration.html*

Resolves: #684

### DIFF
--- a/src/features/utils/ansibleCfg.ts
+++ b/src/features/utils/ansibleCfg.ts
@@ -29,7 +29,7 @@ export async function scanAnsibleCfg(
   rootPath: string | undefined = undefined
 ): Promise<AnsibleVaultConfig | undefined> {
   /*
-   * Reading order (based on the documentation: https://docs.ansible.com/ansible/2.4/intro_configuration.html):
+   * Reading order (based on the documentation: https://docs.ansible.com/ansible/latest/reference_appendices/config.html#ansible-configuration-settings):
    * 1) ANSIBLE_CONFIG
    * 2) ansible.cfg (in current workspace)
    * 3) ~/.ansible.cfg
--- a/src/features/utils/ansibleCfg.ts
+++ b/src/features/utils/ansibleCfg.ts
@@ -29,13 +29,13 @@ export async function scanAnsibleCfg(
   rootPath: string | undefined = undefined
 ): Promise<AnsibleVaultConfig | undefined> {
   /*
-   * Reading order:
+   * Reading order (based on the documentation: https://docs.ansible.com/ansible/2.4/intro_configuration.html):
    * 1) ANSIBLE_CONFIG
    * 2) ansible.cfg (in current workspace)
    * 3) ~/.ansible.cfg
-   * 4) /etc/ansible.cfg
+   * 4) /etc/ansible/ansible.cfg
    */
-  const cfgFiles = ["~/.ansible.cfg", "/etc/ansible.cfg"];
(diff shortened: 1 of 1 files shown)


src/features/utils/ansibleCfg.ts
```diff
@@ -29,13 +29,13 @@ export async function scanAnsibleCfg(
   rootPath: string | undefined = undefined
 ): Promise<AnsibleVaultConfig | undefined> {
   /*
-   * Reading order:
+   * Reading order (based on the documentation: https://docs.ansible.com/ansible/2.4/intro_configuration.html):
```
ganeshrn, line 32: This point to the latest stable release docs https://docs.ansible.com/ansible/latest/installation_guide/intro_configuration.html

ganeshrn: Also `ansible --version` has information about the ansible configuration file path read from.

priyamsahoo: Thanks for pointing that out.
I can link this page then, that points to the locations: https://docs.ansible.com/ansible/latest/reference_appendices/config.html#ansible-configuration-settings

priyamsahoo: Yes, I am aware of the `ansible --version` command, but this code is present on the extension side, and the dev who wrote this intends to scan the locations rather than using the command. Thus my PR is only a rectification for it.

src/features/utils/ansibleCfg.ts
```diff
@@ -29,13 +29,13 @@ export async function scanAnsibleCfg(
   rootPath: string | undefined = undefined
 ): Promise<AnsibleVaultConfig | undefined> {
   /*
-   * Reading order:
+   * Reading order (based on the documentation: https://docs.ansible.com/ansible/latest/reference_appendices/config.html#ansible-configuration-settings):
    * 1) ANSIBLE_CONFIG
    * 2) ansible.cfg (in current workspace)
    * 3) ~/.ansible.cfg
-   * 4) /etc/ansible.cfg
+   * 4) /etc/ansible/ansible.cfg
```
ssbarnea, line 36: We have people keeping ansible.cfg and ansible files in subfolders and this approach does not help them. Any idea what we can do for them? (not as part of tris PR)

ssbarnea (approved):

src/features/utils/ansibleCfg.ts
ssbarnea: Yeah, when user does not have ansible we cannot call ansible itself to report config location.

src/features/utils/ansibleCfg.ts
priyamsahoo: A similar thing (not the same) has been implemented in the ALS to locate the `.ansible-lint` config file [(ref)](https://github.com/ansible/ansible-language-server/blob/main/src/services/ansibleLint.ts#L236-L263). That is, even if the user has no .ansible-lint file in the CWD, we traverse to the root of the workspace and try to find out config files in any of the parent directories up till the root.

This does solve the issue to some extent.
But, again, if ansible doesn't recognize the config file associated with a playbook(if it does, `ansible --version` should show it), there is no point in writing code to locate it.

PR #819, opened 2023-03-21
Test with min/max versions supported by extest

ssbarnea:
As extest supports only a clear set of vscode versions, we aim to
match them, so we can move to newer version as soon they are
supported.

It should be noted that quite often extest needs changes in order
to properly test extensions with newer versions of vscode. This change
should make the testing outcome more predictable.


ganeshrn (changes requested):
We need to maintain the minimum version supported as 1.71.0 for inline suggestion API's to work properly.

PR #820, opened 2023-03-22
Inline suggestion improvements (#820)Co-authored-by: Ganesh Nalawade <ganesh634@gmail.com>

priyamsahoo:
The PR adds bugs fixes and minor UI improvements:
1. Fix the inline suggestion disappearing problem (https://issues.redhat.com/browse/AAP-9372).
2. Fix the problem of completion resolution not working (by changing the functionality of onEnter).
3. Fix the working of task completion in case of explicit request by the user (by properly invoking Ansible wisdom inline suggestion trigger).
4. Fix the problem of attempting to send requests to wisdom service even when related settings were disabled.
5. Remove "Processing..." and put a loading animation while sending a task suggestion request.


src/features/wisdom/inlineSuggestions.ts
```diff
@@ -1,26 +1,22 @@
...
-let suggestionId = "";
-let currentSuggestion = "";
-const taskRegexEp =
+const TASK_REGEX_EP =
   /(?<blank>\s*)(?<list>-\s*name\s*:\s*)(?<description>.*)(?<end>$)/;
 
+let suggestionId = "";
+let currentSuggestion = "";
 let inlineSuggestionData: InlineSuggestionEvent = {};
-let inlineSuggestionDisplayed = false;
```
ganeshrn, line 23: `inlineSuggestionDisplayed` and all the code that uses it are required, please revert back the changes for suggestion displayed logic

src/features/wisdom/inlineSuggestions.ts
```diff
@@ -214,43 +174,53 @@ async function getInlineSuggestions(
...
   }
-  console.log("inlineSuggestionCommitHandler triggered");
   const editor = vscode.window.activeTextEditor;
   if (!editor) {
     return;
   }
+
+  // Trigger the suggestion explicitly
+  console.log("inlineSuggestion Handler triggered Explicitly");
+  vscode.commands.executeCommand("editor.action.inlineSuggest.commit");
```
ganeshrn, line 191: Looks like this will directly commit the suggestion and not be displayed as inline suggestion if yes this doesn't seem right.

priyamsahoo: Thanks for pointing it out; it should be `editor.action.inlineSuggest.trigger.`

src/features/wisdom/inlineSuggestions.ts
priyamsahoo: Added.

ganeshrn: Why is that? Most likely you are not setting it properly at right place.

ganeshrn (approved):

PR #857, opened 2023-04-17
Add error message for invalid playbook

ganeshrn:
For prompts as shown below which is not a valid syntax the inline suggestion is not trigger is silently, instead add an error message for the user.
```
---
name: Deploy web servers
tasks:
- name: Install nginx
```


src/features/lightspeed/utils/data.ts
```diff
@@ -23,6 +23,9 @@ export function shouldRequestInlineSuggestions(
     typeof parsedAnsibleDocument === "object" &&
     !Array.isArray(parsedAnsibleDocument)
   ) {
+    vscode.window.showErrorMessage(
+      "Ansible Lightspeed expects valid Ansible syntax. For playbook files it should be a list of plays and for tasks files it should be list of tasks."
```
robinbobbitt, line 27: I'm concerned novice users will struggle with this message. Can we detect if it's a playbook at least and tailor the message better? "A playbook must be a list of plays." or "A tasks file must be a list of tasks."

ganeshrn: The problem is when user writes `name: Deploy web servers` and hits enter there is no certain way to identify if they are writing a playbook file or a task file (which they intend to include in another playbook).

priyamsahoo (approved):

Issue #775, opened 2023-02-06
Improve ansible* executable discovery


cidrblock:
Currently the user has to specify paths to ansible, ansible-navigator, and ansible-lint or the virtual environment or the virtual environment activation script.

This can require some detailed knowledge of python which all users may not have.

Since we require the python extension, can we leverage its configuration by default and leave our own settings in place as overrides?

A good reference might be here: https://github.com/microsoft/vscode-python-tools-extension-template/blob/main/src/extension.ts#L9

(that template was announced here: https://devblogs.microsoft.com/python/python-in-visual-studio-code-august-2022-release/#python-tools-extension-template

g0t4:
I was curious about the relationship between settings in the `python` extension (`ms-python.python`) and the `redhat.ansible` extension... so I did some testing...

In my case, pointing the python extension to my custom virtual env led to the `redhat.ansible` extension working too. Removing this setting breaks the ansible extension.

I do not have ansible installed globally. Instead, it's installed in a custom venv in my workspace. Along with ansible-lint. This way the ansible extension won't just work with its default config.

In `settings.json` I added/removed this one setting:
```json 
"python.defaultInterpreterPath": "test/.venv/bin/python3",
```

To test this setting, I used a file that I have setup to match as language mode = ansible. Each time I made a change to this setting, to test it, I closed the test ansible file in vscode, quit/reopened vscode, reopened the test file. Then I waited to see if I got error messages in toasts from the ansible extension, or if the ansible extension features loaded (ie ansible-lint output in the editor).

Without the setting I get the following two errors (as well as ansible extension features not working):

    Ansible-lint is not available. Kindly check the path or disable validation using ansible-lint

    Command failed: ansible-config dump
    /bin/sh: ansible-config: command not found

When I put the setting back, both errors did not show and instead I had ansible extension features operational once again.

If this isn't explicitly designed in the `redhat.ansible` extension then perhaps it's just by virtue of the python extension activating the virtual env (modifying the path) so that the default settings for ansible/ansible-lint/etc just work?

FYI, testing was unreliable if I didn't fully quit and reopen vscode. Commands like "Reload Window" aren't enough with this settings change.

g0t4:
It would be rock to fully document the expected behavior/interplay of the various path settings and python `venv`s settings within this `redhat.ansible` extension. As well as expected behavior with key settings in dependent plugins (ie python extension).

cidrbl0ck:
[...]

PR #871, opened 2023-05-03
Easy discovery of Python interpreter and Ansible executables

priyamsahoo:
The PR:
1. Adds the ability to communicate with the Python extension and sets the interpreter based on the recommended one by the Python extension.
2. Adds feature to select interpreter on the go by providing options for available interpreters (similar to Python's select interpreter feature).
3. fixes: https://github.com/ansible/vscode-ansible/issues/775


src/features/utils/python.ts
```diff
@@ -0,0 +1,263 @@
...
+    environment: undefined,
+    version: undefined,
+  };
+}
+
+export function checkVersion(
+  resolved: ResolvedEnvironment | undefined
+): boolean {
+  const version = resolved?.version;
+  if (version?.major === 3 && version?.minor >= 7) {
```
ganeshrn, line 254: This can be changed to `version?.major >= 3`

ganeshrn (approved):

PR #892, opened 2023-05-16
Update inline suggestion trigger condition (#890)

ganeshrn:
*  Currently inline suggestion is triggered after user hits new line key at end of the task name description. So on every space/tab a lightspeed request is triggered which is not expected.
*  With this fix the request will be triggered only when the number of spaces before cursor position on new line is equal to the number of spaces before `-` in the previous line (task name description).


priyamsahoo (approved):
looks good to me.

Issue #859, opened 2023-04-17
Disabling Lightspeed in settings does not disable features


timfallmk:
### Summary

With the introduction of [new Lightspeed integration in 2.0](https://github.com/ansible/vscode-ansible/discussions/855) I'm now getting bugged to "log in to ansible lightspeed to use ansible" on every open window. I have the "Lightspeed" enable option unchecked in the settings.
![image](https://user-images.githubusercontent.com/5316117/232562592-ba1ca03a-d5b1-44af-9d97-effba89e953c.png)
![image](https://user-images.githubusercontent.com/5316117/232562830-26907804-388c-4400-85bf-43794ef5286a.png)

There may be other parts of Lightspeed that are not being disabled when the user explicitly does not enable them.

### Extension version

v2.0.58

### VS Code version

1.77.3

### Ansible Version

```shell
ansible [core 2.14.4]
  config file = /Users/tfall/Code/stand-configuration/ansible.cfg
  configured module search path = ['/Users/tfall/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/homebrew/Cellar/ansible/7.4.0/libexec/lib/python3.11/site-packages/ansible
  ansible collection location = /Users/tfall/.ansible/collections:/usr/share/ansible/collections
  executable location = /opt/homebrew/bin/ansible
  python version = 3.11.3 (main, Apr  7 2023, 20:13:31) [Clang 14.0.0 (clang-1400.0.29.202)] (/opt/homebrew/Cellar/ansible/7.4.0/libexec/bin/python3.11)
  jinja version = 3.1.2
  libyaml = True
```

### OS / Environment

macOS Ventura 13.3.1
ansible-lint 6.14.6 using ansible 2.14.4

rothfralick:
Same here.  Very annoying.
Ansible extension v2.0.58
VSCode 1.77.3
Windows 11

Zombiefleischer:
Also autocompletion is not working with this new version
Ansible extension v2.0.58
VSCode 1.77.3
Linux Debian

ganeshrn:
@timfallmk Thanks for raising the issue, we will work on the fix.

@Zombiefleischer 

> Also autocompletion is not working with this new version


Autocompletion should work irrespective of Lightspeed is enabled or disabled.
Is it possible for you to share a screen recording?
[...]

PR #911, opened 2023-06-05
Fix notification for lightspeed login after the setting is disabled

ganeshrn:
Fixes https://github.com/ansible/vscode-ansible/issues/859

In addition to PR https://github.com/ansible/vscode-ansible/pull/877 update the logic to initialize the Lightspeed authentication provider only after the Lightspeed setting is enabled.

Note: 
*  As Lightspeed setting is disabled by default the login notification won't appear with this fix
*  However if the setting is enabled the login notification will appear and after that if it is disabled again the user will have to reload the window for the login notification to disappear.


priyamsahoo (approved):
Perfect!

PR #933, opened 2023-06-28
Follow-up changes for supporting semantic markup

felixfontein:
Ref: https://github.com/ansible/ansible-language-server/pull/563

This won't work until a new release of ansible-language-server is out. (My *guess* is that the next release will be 1.2.0, since there are new features.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -623,7 +623,7 @@
     ]
   },
   "dependencies": {
-    "@ansible/ansible-language-server": "^1.1.0",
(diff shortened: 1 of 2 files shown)


felixfontein:
Hmm, I think 33567bd43f098c3709163735f9d0d5075c3b1fec makes this PR no longer necessary. @priyamsahoo can you confirm?

ganeshrn:
@felixfontein Yup the PR is no longer required as it is fixed in @priyamsahoo commit. I will go ahead and close this one.

PR #940, opened 2023-07-10
Fix ansibleContent feedback event trigger

ganeshrn:
*  Don't send ansibleContent feedback if the file content is empty or previous file content is same as the current one.


ajinkyau (approved):
LGTM

priyamsahoo (approved):

PR #955, opened 2023-08-01
Fix var capturing regex

priyamsahoo:
The PR:
1. simplifies regex to capture `_var_` inside jinja brackets.
2. adds lightspeed e2e tests for asserting the correct syntax of variables within the jinja inline brackets returned in the suggestion.


ganeshrn (approved):
Thanks for working on it!

PR #971, opened 2023-09-13
Add support specifying model Id for Lightspeed subscribed users

ganeshrn:
*  Identify if the logged in user has a seat based on the response received from the `/me` endpoint
*  Add a model Id setting to provide the value of the model to use for inline suggestions.
*  If logged in user has a seat and if the value of modelId is specified in the setting pass the modelId value in the completion request.
*  Use `external_username` instead of `username` to display the value of logged in user.
* Change the name `watson` to `watsonx`


priyamsahoo (approved):
Looks good to me.

PR #973, opened 2023-09-13
Add additional context for lightspeed suggestions

ganeshrn:
For each of the completion request send additional context as below
*  The variables referred in the list of `vars_files` keyword if present and can be resolved to a path of the file system.
*  The variables referred in the `include_vars` task if present and can be resolved to a path of the file system.
*  The list of roles and it's data
   - The relative path of role with respect to the playbook path
   - The name of the role
   - List of tasks within the role
   - The default variables under the default folder
   - The vars variables under the vars folder

Note: Only the variable structure and key value is preserved and all the
      values are removed.

### DIFF
--- a/.config/dictionary.txt
+++ b/.config/dictionary.txt
@@ -119,6 +119,7 @@ untildify
 userdata
 usermod
 uuidv4
+varInfiles
 venvs
 vscoss
 vsix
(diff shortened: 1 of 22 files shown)


priyamsahoo (approved):
Looks good.

PR #992, opened 2023-10-19
Fix for multi-task and single-task lightspeed suggestions

ganeshrn:
*  Remove whitespaces from blank lines in between multi-task suggestion
*  Show user information message is model Id is invalid.
*  If file is under playbook folder don't trigger single-task suggestion in play context.
*  Add information message for user if both Github copilot for Ansible and Lightspeed setting in enabled.
*  Don't trigger single-task suggestion if `- name: <descrption>` pattern if found under `vars` context.
*  Remove `ansible_type` field from content matches display as it is no longer used.
*  If user is not logged in and lightspeed setting is enabled show login message for mulit-task suggestion trigger request.

### DIFF
--- a/src/features/lightspeed/api.ts
+++ b/src/features/lightspeed/api.ts
@@ -77,7 +77,9 @@ export class LightSpeedAPI {
       return {} as CompletionResponseParams;
     }
     console.log(
-      `Completion request sent to lightspeed: ${JSON.stringify(inputData)}`
+      `[ansible-lightspeed] Completion request sent to lightspeed: ${JSON.stringify(
+        inputData
+      )}`
     );
     try {
       const response = await axiosInstance.post(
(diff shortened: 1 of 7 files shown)


src/features/lightspeed/api.ts
```diff
@@ -76,6 +76,9 @@ export class LightSpeedAPI {
       console.error("Ansible Lightspeed instance is not initialized.");
       return {} as CompletionResponseParams;
     }
+    console.log(
```
priyamsahoo, line 79: Can we add a prefix for these logs, like we did for the previous features? Something like: `[ansible-lightspeed]`.

priyamsahoo (approved):

src/features/lightspeed/api.ts
ganeshrn: Fixed

PR #1026, opened 2023-12-06
CI fixes

goneri:


goneri:
@ganeshrn it looks like something is wrong with EE and MacOS.

ganeshrn:
@goneri The macos ci job is flaky and fails intermittently. We can ignore that failure.

tools/test-setup.sh
```diff
@@ -216,13 +216,14 @@ log notice "Installing $(python3 --version) venv and dependencies matching creat
 VIRTUAL_ENV=${VIRTUAL_ENV:-out/venvs/${HOSTNAME}}
 if [[ ! -d ${VIRTUAL_ENV} ]]; then
     log notice "Creating virtualenv ..."
-    python3 -m venv "${VIRTUAL_ENV}"
+    python3.11 -m venv "${VIRTUAL_ENV}"
```
ssbarnea, line 219: I do not fancy this option at all as it assumes you have a py311 interpreter. I might have py312 only and it will break the logic.

goneri: The CI runs 3.11 and I would prefer to avoid any behavioural change base on the local set-up.

ajinkyau (approved):
@ganeshrn Looks good to me. Let me know if we can merge this

ganeshrn:
@goneri Thanks for working on this.

PR #1028, opened 2023-12-08
lightspeed: a unit-test for the error handling

goneri:
Add test coverage for the error handling. The tests use `jest` and can be enabled with:
    
```
yarn unit-tests
```

or:

```
task unit-tests
```

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,5 @@
+/** @type {import('ts-jest').JestConfigWithTsJest} */
+module.exports = {
+  preset: "ts-jest",
+  testEnvironment: "node",
+};
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,3 +1,4 @@
+/* eslint-disable */
 /** @type {import('ts-jest').JestConfigWithTsJest} */
 module.exports = {
   preset: "ts-jest",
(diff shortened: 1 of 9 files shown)


jest.config.js
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable */
+/** @type {import('ts-jest').JestConfigWithTsJest} */
+module.exports = {
```
TamiTakamiya, line 3: I would like to generate reports when the unit test is executed locally.  How about adding jest options to generate reports:
```
  collectCoverage: true,
  coverageDirectory: "out/unitTestReport",
  coverageReporters: ["json", "html"],
```
here?  When e2e test is executed, html & json reports are crated under `out/e2eTestReport`. I think saving those files in `out/unitTestReport` would be consistent.  What do you think?

TamiTakamiya (approved):
LGTM!  Thank you for coding this. It works great! I left one minor comment on coverage report generation.

jest.config.js
goneri: e2e uses another framework called Mocha. Also, the GHA job expects the coverage to be in `./coverage/` (https://github.com/ansible/vscode-ansible/pull/1028/files#diff-a2115d277b5ca5a2f09a999e53440839cf332b94da177f3d1766334555b0f7c6R26.

ganeshrn (approved):
@goneri Thanks for working on this!

PR #1040, opened 2024-01-11
fix check for no seat assigned

robinbobbitt:
Fixes a small bug in the logic around when to show "No seat assigned".


goneri (approved):

goneri:
Thank you @robinbobbitt!

ganeshrn (approved):

PR #1041, opened 2024-01-11
unit-test coverage for contentMatchesWebview

goneri:
Extend the unit-test coverage for the contentMatchesWebview module.
Also refresh the `sinon` dependency.

### DIFF
--- a/.config/dictionary.txt
+++ b/.config/dictionary.txt
@@ -67,6 +67,7 @@ haaaad
 hbenl
 hostvars
 hotfixes
+icontent
 jeinwag
 johndoe
 keyfile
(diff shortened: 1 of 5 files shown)


ganeshrn (approved):
Thanks for working on this.

ganeshrn:
@goneri `codecov/project` is failing, I think the coverage file settings need to updated.

goneri:
> @goneri `codecov/project` is failing, I think the coverage file settings need to updated.

I suspect it's because https://docs.codecov.com/docs/commit-status#target the `target` is not set and it uses the current value from main (100% currently).
I cannot change the settings myself.

PR #1046, opened 2024-01-15
Show model ID to be used in statusbar tooltip

TamiTakamiya:
In the current (V2.8) implementation, the Lightspeed model Id shown on the Lightspeed statusbar tooltip is based on the response of a previous Lightspeed inference API call, i.e. if no inference API call was made, no model Id is shown there.

This PR will change the behavior.  With this PR, the model Id is displayed with the following logic:

1. If a model ID is specified for Model ID Override in Settings
    - Show the Model ID specified in Settings
        Screenshot from 2024-01-15 13-34-15

1. If a model ID is not specified for Model ID Override
    - Show `default`
        image


src/features/lightspeed/utils/uiUtils.ts
```diff
@@ -0,0 +1,11 @@
+export function getModelDetailsString(
+  modelIdOverride: string | undefined,
+  currentModelId: string | undefined
+): string {
+  const orgDefault = "(Red Hat org default)";
```
ganeshrn, line 5: Can move the constant here https://github.com/ansible/vscode-ansible/blob/main/src/definitions/lightspeed.ts
Do we need `Red Hat` in the name? This might be confusing as different org can have different base model.
How about just `default` to start with and `Model Id (default)` when extension receives the first LS response.
So the hover display would be
```
Model: default
```
```
Model: b8a86296-ef64-4ddb-bbf4-a2fd164577bb<|sepofid|>granite-3b (default)

Comment (TamiTakamiya) [reply src/features/lightspeed/utils/uiUtils.ts]: I have moved the string definition to lightspeed.ts.

On the model ID display for default one, I have modified the code following suggestions from Marty and Robin and dropped the code to show the actual ID.  So when the default model is used, always `Model: default` will be shown.

**Review by priyamsahoo — APPROVED** [review]

File: src/definitions/lightspeed.ts [hunk 2]
```diff
@@ -61,6 +61,8 @@ export const LIGHTSPEED_SERVICE_LOGIN_TIMEOUT = 120000;
 
 export type LIGHTSPEED_SUGGESTION_TYPE = "SINGLE-TASK" | "MULTI-TASK";
 
+export const LIGHTSPEED_DEFAULT = "default";
```
ganeshrn, line 64: Minor comment, can we change `LIGHTSPEED_DEFAULT` to `LIGHTSPEED_MODEL_DEFAULT` to me more precise.

PR #1050, opened 2024-01-17
Do not send duplicated inline suggestion feedback

TamiTakamiya:


src/features/lightspeed/api.ts
```diff
@@ -114,7 +128,6 @@ export class LightSpeedAPI {
     } catch (error) {
       const err = error as AxiosError;
       vscode.window.showErrorMessage(retrieveError(err));
-      this._completionRequestInProgress = false;
```
ganeshrn, line 117: Why is this line removed? The return in next line won't run the finally block.

src/features/lightspeed/api.ts
```diff
@@ -86,14 +100,14 @@ export class LightSpeedAPI {
...
       this._completionRequestInProgress = true;
+      this._inlineSuggestionFeedbackSent = false;
       const response = await axiosInstance.post(
         LIGHTSPEED_SUGGESTION_COMPLETION_URL,
         inputData,
         {
           timeout: ANSIBLE_LIGHTSPEED_API_TIMEOUT,
         }
       );
-      this._completionRequestInProgress = false;
```
ganeshrn, line 96: Why is this line removed? The return in next line won't run the finally block.

TamiTakamiya: Keeing the line will not hurt anything, but I think we'd better take it out because the `finally` block is executed even if `return` is inserted in the middle of the `try` block.  For example, the following code:
```
function foo(i) {
  try {
    if (i > 0) {
      console.log("i > 0");
      return "POSITIVE";
    } else if (i == 0) {
      throw new Error("i == 0");
    } else {
      console.log("i < 0");
    }
    return "NEGATIVE";
  } catch (e) {
     console.log(e.toString());
     return "ZERO";
  } finally {
     console.log("FINALLY");
  }
}

console.log(foo(1));
console.log(foo(0));
console.log(foo(-1));
```
executes like:
```
$ node foo.js
i > 0
FINALLY
POSITIVE
Error: i == 0
FINALLY
ZERO
i < 0
FINALLY
NEGATIVE
```
What do you think?

src/features/lightspeed/api.ts
TamiTakamiya: Pls look at my response above. Thanks!

src/features/lightspeed/api.ts
ganeshrn: Thanks for the explanation. The changes looks good to me.

ganeshrn (approved):

PR #1052, opened 2024-01-20
lightspeed/inlineSuggestions: refactoring to simplify the testing

goneri:
The goal is to break up the logic in two parts:
1. The code that we use parse the prompt is now isolated in:
  - `getCompletionState()` and
  - `InlineSuggestionState()`.
  It will only return a state that trigger a callback function.
2. Is the second part, where our `on*` callbacks are finally called
  to trigger potentially a change.

The benefits of this approach:

- Since we parsing of the content and prompt only creates a state without
  any side-effect, it becomes easier to properly test it.
- Also the callbacks are much smaller and simpler. They become more suitable
  for future unit-test integration.

Also, this PR reduces the number of global variables.

### DIFF
--- a/.config/dictionary.txt
+++ b/.config/dictionary.txt
@@ -68,6 +68,7 @@ hbenl
 hostvars
 hotfixes
 icontent
+inlinesuggestion
 jeinwag
 johndoe
 keyfile
@@ -125,6 +126,7 @@ tsdoc
 ttakamiy
 unstash
 untildify
+useractionhandler
 userdata
 usermod
 uuidv4
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -45,8 +45,7 @@ import {
   inlineSuggestionTriggerHandler,
   inlineSuggestionCommitHandler,
   inlineSuggestionHideHandler,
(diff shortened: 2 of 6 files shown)


.config/dictionary.txt
```diff
@@ -68,6 +68,7 @@ hbenl
 hostvars
 hotfixes
 icontent
+inlinesuggestion
```
ganeshrn, line 71: For camel case `inlineSuggestion` this entry is likely not required

.config/dictionary.txt
```diff
@@ -125,6 +126,7 @@ tsdoc
 ttakamiy
 unstash
 untildify
+useractionhandler
```
ganeshrn, line 129: For camel case `userActionHandler` this entry is likely not required

ganeshrn (commented):
@goneri Can you please raise this PR from a branch within the repo instead of fork. This will run the Lightspeed integration test suite.

goneri:
Superseded by https://github.com/ansible/vscode-ansible/pull/1055, cf @ganeshrn's comment.

PR #1055, opened 2024-01-22
lightspeed/inlineSuggestions: refactoring to simplify the testing

goneri:
The goal is to break up the logic in two parts:
1. The code that we use parse the prompt is now isolated in:
  - `getCompletionState()` and
  - `InlineSuggestionState()`. It will only return a state that trigger a callback function.
2. Is the second part, where our `on*` callbacks are finally called to trigger potentially a change.

The benefits of this approach:

- Since we parsing of the content and prompt only creates a state without any side-effect, it becomes easier to properly test it.
- Also the callbacks are much smaller and simpler. They become more suitable for future unit-test integration.

Also, this PR reduces the number of global variables.

### DIFF
--- a/src/features/lightspeed/inlineSuggestion/suggestionDisplayed.ts
+++ b/src/features/lightspeed/inlineSuggestion/suggestionDisplayed.ts
@@ -1,0 +1,25 @@
+import { InlineCompletionItem } from "vscode";
+
+export class SuggestionDisplayed {
+  inlineSuggestionDisplayed: boolean;
+  cachedCompletionItem: InlineCompletionItem[];
+
+  constructor() {
+    this.inlineSuggestionDisplayed = false;
+    this.cachedCompletionItem = [];
+  }
+
+  reset() {
+    this.inlineSuggestionDisplayed = false;
+    this.cachedCompletionItem = [];
+  }
+
+  set(inlineCompletionItem: InlineCompletionItem[]) {
+    this.inlineSuggestionDisplayed = true;
+    this.cachedCompletionItem = inlineCompletionItem;
+  }
+
+  get() {
+    return this.inlineSuggestionDisplayed;
+  }
+}
--- a/src/features/lightspeed/inlineSuggestions.ts
+++ b/src/features/lightspeed/inlineSuggestions.ts
@@ -12,40 +12,194 @@ import {
   CompletionResponseParams,
   InlineSuggestionEvent,
   CompletionRequestParams,
-  IRolesContext,
-  IRoleContext,
-  IStandaloneTaskContext,
 } from "../../interfaces/lightspeed";
 import {
+  LightSpeedCommands,
   LIGHTSPEED_SUGGESTION_TYPE,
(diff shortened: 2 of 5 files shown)


TamiTakamiya:
I was looking at the issue that pre-commit added unnecessary trailing comma in my PR and found [this issue in prettier](https://github.com/prettier/prettier/issues/15942), which was closed 5 days ago.

Although I guess this will be resolved eventually in prettier, I have coded a workaround [here].(https://github.com/ansible/vscode-ansible/blob/188d23affa819c471b938190eff0042a1530d54f/.prettierrc.yaml#L14-L17)

goneri:
> I was looking at the issue that pre-commit added unnecessary trailing comma in my PR and found [this issue in prettier](https://github.com/prettier/prettier/issues/15942), which was closed 5 days ago.
> 
> Although I guess this will be resolved eventually in prettier, I have coded a workaround [here].(
> 
> https://github.com/ansible/vscode-ansible/blob/188d23affa819c471b938190eff0042a1530d54f/.prettierrc.yaml#L14-L17
> )

thank you Tami. It works!

src/features/lightspeed/inlineSuggestions.ts
```diff
@@ -12,40 +12,194 @@ import {
...
+  return suggestionItems;
+};
+
+const CompletionState = {
+  TextEditorNotActive: onTextEditorNotActive,
+  NotForMe: onNotForMe,
+  CancellationRequested: onCancellationRequested,
+  LightspeedIsDisabled: onLightspeedIsDisabled,
+  LightspeedURLMisconfigured: onLightspeedURLMisconfigured,
+  RetrySuggestion: onRetrySuggestion,
```
ganeshrn, line 144: In this state we are returning the recommendation from cache, `RetrySuggestion` name can be misleading as it might suggest we are requesting Lightspeed service for a new suggestion. How about `CacheSuggestion: onCacheSuggestion`?

ganeshrn (approved):
@goneri Nice work! Thanks for doing this.
Apart from the minor naming comment looks good to me.

src/features/lightspeed/inlineSuggestion/suggestionDisplayed.ts
```diff
@@ -0,0 +1,25 @@
+import { InlineCompletionItem } from "vscode";
+
+export class SuggestionDisplayed {
+  inlineSuggestionDisplayed: boolean;
+  cachedCompletionItem: InlineCompletionItem[];
```
TamiTakamiya, line 5: Is this `cachedCompletionItem` needed? I cannot find where it is being read...

TamiTakamiya: Ok, I found it.

TamiTakamiya (approved):
LGTM. Honestly I cannot say I understand them 100%, but I would like to start writing tests with these codes. Thank you very much for your efforts on this.

PR #1059, opened 2024-01-22
lightspeed: show a proper message when the user has no seat

goneri:
Return the following message when the users cannot reach the service because
they have no seat:

`You don't have access to IBM watsonx Code Assistant. Contact your administrator.`


TamiTakamiya (approved):
LGTM

ganeshrn (approved):

PR #1056, opened 2024-01-22
Enable code coverage on e2e tests

TamiTakamiya:
Moved from #1053 (moved the code changes from a forked branch to a branch of the upstream repo)
----
Enable code coverage on End-to-End tests. HTML report is generated in the `coverage/index.html`.

(Reference)
- https://frenya.net/blog/vscode-extension-code-coverage-nyc

Screenshot from 2024-01-20 09-19-58

PR #1067, opened 2024-01-29
Integrate code coverage & mock Lightspeed server into CI

TamiTakamiya:
This PR is for integrating the code coverage support (#1056) and the Lightspeed mock server (#1062) into the CI pipeline. It also changes e2e test to include unit tests so that the code coverage by unit tests is also included in the coverage report.

Following are some details:

1. The new script `coverage-all` was added to `package.json`.  It will start the mock server, run tests (unit and e2e tests), generate coverage report, and terminates server.
    * For starting & terminating server, [start-server-and-test](https://www.npmjs.com/package/start-server-and-test) npm package is used.
1. 20 out of 21 unit tests could run successfully from e2e test runner, but one failed in spying `console.log()` funciton. `src/features/lightspeed/contentMatchesWebview.ts` was modified to bypass that issue.
1. Unit tests are executed separately from e2e tests, i.e. those tests are executed twice. I think it would be OK as they are executed using two different way (one is the standard way and this is from e2e runner) and it won't take long time.
1. The mock server seemed to be listening the ipv6 address in GitHub Actions Linux environment. And the axios library used in the extension does not support URLs that include ipv6 address like `http://[::1]:3000` ([referece](https://github.com/axios/axios/issues/5333)), a URL using the ipv6 loopback hostname `ip6-localhost` is given to the extension. For macos, `http://localhost:3000` is used as `ip6-localhost` is not available on macos.
1. The test coverage I got with this code was about 38%.


ganeshrn (approved):

tools/test-setup.sh
```diff
@@ -302,6 +302,10 @@ if [[ -n "${CI:-}" ]]; then
         log warning "Removing unexpected ~/.ansible folder found on CI to avoid test contamination."
         rm -rf "${HOME}/.ansible"
     fi
+    # Add an /etc/hosts entry for a local ipv6 address ("localhost6")
+    if ! grep -q localhost6 /etc/hosts; then
+        echo "::1         localhost6" | sudo tee -a /etc/hosts
```
ssbarnea, line 307: That will be incompatible with macos, where we will not want to touch it.

Interestingly on macos I see:

$ cat /etc/hosts
##
# Host Database
#
# localhost is used to configure the loopback interface
# when the system is booting.  Do not change this entry.
##
127.0.0.1   localhost
255.255.255.255 broadcasthost
::1             localhost

This makes me think that localhost works for both ipv4 and ipv6 on mac anyways.

TamiTakamiya: @ssbarnea  The reason I used `localhost6` was that `127.0.0.1`, which was used in the original code, did not work in GitHub Actions. When `localhost` was given as the mock server address, the server was listening the ipv6 address only. For enforcing the axios library used in our VS Code extension to look at the ipv6 address instead of `127.0.0.1`, I needed to have an ipv6-only entry in `/etc/hosts`.

I did some more research and found that another hostname `ip6-localhost` was available on the GitHub Action runner Ubuntu image and running server/extension with `127.0.0.1` working in macos tests.  That means:

- CI tests on (Ubuntu) Linux can use `ip6-localhost` as the hostname to resolve to the local ipv6 address
- CI tests on macos can use the ipv4 address `127.0.0.1`
- So there is no need to modify `/etc/hosts`

The new commits reflect those changes. Would you take a look? Thank you!

goneri: This seems to work fine: https://github.com/ansible/vscode-ansible/pull/1073/commits/a32d85e9bc633cce0c7f9e4a6275d9945dcf2fe9

TamiTakamiya: The test does not seem to fail with `localhost`, but actually [the log](https://github.com/ansible/vscode-ansible/actions/runs/7731339027/job/21078753280) contains:
```
Skipping lightspeed tests because of the following error message:  connect ECONNREFUSED 127.0.0.1:3000
```
This was from [these lines](https://github.com/ansible/vscode-ansible/blob/main/test/helper.ts#L172-L176).

goneri: Oh my bad.

goneri (approved):

ganeshrn (approved):

ganeshrn:
@TamiTakamiya Lightspeed test cases are failing on macos.

TamiTakamiya:
> @TamiTakamiya Lightspeed test cases are failing on macos.

@ganeshrn Thank you for informing me of this failure on macos.  In the macos test log, I see the following error:
```
[0205/042107.081127:ERROR:node_main.cc(135)] Node.js environment variables are disabled because this process is invoked by other apps.
Installing extensions...
```
which seems to be the same one, which was reported recently in https://github.com/microsoft/vscode/issues/204005
I do not see the error in the (linux) test log. I will investigate this further.

tools/test-setup.sh
TamiTakamiya: @ssbarnea Can I resolve this conversation? I would like to merge this to the main branch if it looks good.  As Ganesh pointed out in his comment below, Lightspeed tests are failing in macos test right now, but it is caused by [a known vscode bug](https://github.com/microsoft/vscode/issues/204005) and I believe it will be solved eventually (hopefully soon!).

ganeshrn:
@TamiTakamiya Thanks for working on this. The Lightspeed testcases is passing now. The macos CI is failing for execution environment test which care failing due to setup issues

PR #1070, opened 2024-01-29
Fix two test runner issues

TamiTakamiya:
This PR contains fixes for:
* Two test runner issues
    1. `test/index.ts` -- This caused the issue that e2e-test step did not fail even when a test case failed
    1. `test/testScripts/e2eTestSuiteLauncher.test.ts` -- This issue prevent non-Lightspeed e2e tests running regardless of the value of `RUN_LIGHTSPEED_TESTS_ONLY` envvar.
    
In addition to those fixes, this PR also contains changes for `testDiagnosticsAnsibleWithoutEE`. In testing with two fixes shown above, I found that the test `"should complain about command syntax-check failed"` did not work as expected, i.e. command syntax-check DID NOT fail against `diagnostics/ansible/without_ee/2.yml`. It looked like the file was recognized as a regular YAML file, not as a playbook. It was fixed by renaming `2.yml` to `playbook_2.yml`.  `1.yml` was also renamed to `playbook_1.yml` to be consistent.


ganeshrn (approved):

ssbarnea (approved):

ssbarnea:
@TamiTakamiya Because this change has no effect on shipped code, we use `skip-changelog`. This avoids cluttering the release notes with work done on testing or CI/CD.

PR #1075, opened 2024-02-01
Content Matches E2E test

TamiTakamiya:


test/testFixtures/settings.json
```diff
@@ -17,8 +17,8 @@
   "ansible.python.activationScript": "",
   "ansible.python.interpreterPath": "python3",
   "ansibleServer.trace.server": "verbose",
-  "ansible.lightspeed.enabled": false,
-  "ansible.lightspeed.suggestions.enabled": false,
+  "ansible.lightspeed.enabled": true,
```
TamiTakamiya, line 20: Unless I set these two settings to `true`, the Ansible panel won't open at the content matches text execution and it fails. It shouldn't fail because we explicitly set these two settings to `true` in the test code. There might be some timing or test-specific issues. I will take at this issue further before sending this PR for review.

test/testScripts/lightspeed/testLightspeed.test.ts
```diff
@@ -90,21 +91,96 @@ export function testLightspeed(): void {
...
         it(`Should give inline suggestion for task prompt '${taskName}'`, async function () {
+          isAuthenticatedStub.returns(Promise.resolve(true));
           await testInlineSuggestion(taskName, expectedModule);
         });
       });
+
+      after(async function () {
+        contentMatchesRequestSpy.restore();
+        sinon.restore();
+        const contentMatchesApiCalls = contentMatchesRequestSpy.getCalls();
```
goneri, line 181: I'm surprised you do the `getCalls()` after the `restore()`. I feel like this is something that may break in the future if the lib decide to reset the inner state when the `restore()` is done. Could you do the opposite, do the `getCalls()` first.

goneri (approved):

test/testScripts/lightspeed/testLightspeed.test.ts
TamiTakamiya: @goneri Thank you, I totally agree with you. I will update the code.

TamiTakamiya: Also a temporary change (function testLightspeed2()) was checked in by mistake. It will also be taken out.

ganeshrn (approved):

PR #1076, opened 2024-02-02
enable lightspeed connect when ready

robinbobbitt:
https://issues.redhat.com/browse/AAP-19148

Fixes issue of `Cannot set properties of undefined (setting 'currentModelValue')` when Connect button is pressed before LightSpeedManager is initialized. 

I considered instead just checking if `lightSpeedManager` is defined before trying to set `currentModelValue`, but this fix feels safer to me given the seemingly delicate nature of the interdependencies and potential timing issues in the code. We could consider moving the initialization of the LightSpeedManager up higher in the `actvate` function too, to decrease the amount of time the `Connect` button is disabled. I think it's ok it's disabled though because the user has the visual indicator of the Sentiment panel trying to load while they wait.

Fix tested on RHEL 9.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,8 @@
       },
       {
         "view": "lightspeed-explorer-treeview",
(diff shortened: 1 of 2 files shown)


robinbobbitt:
Test failures are also present in `main`. This should be good to go.

TamiTakamiya (approved):
Looks good to me. Thanks!

ganeshrn (approved):

PR #1077, opened 2024-02-02
handle invalid python interpreter path

robinbobbitt:
https://issues.redhat.com/browse/AAP-19148

When you have a path in your settings for python interpreter that does not exist, makeLabelFromPath throws an error that causes the `activate` function in extension.ts to fail part way through. As a result, the Lightspeed sentiment panel never loads properly and you can't successfully log in to Lightspeed. This change adds better handling for the scenario of an interpreter path that doesn't exist, and also catches any errors coming from `updatePythonInfoInStatusbar` to avoid disruption of Lightspeed capabilities initializing.

### DIFF
--- a/src/features/pythonMetadata.ts
+++ b/src/features/pythonMetadata.ts
@@ -86,8 +86,10 @@ export class PythonInterpreterManager {
     } else {
       if (!interpreterPath)
         this.pythonInterpreterStatusBarItem.text = "Select python environment";
-      else
+      else {
         this.pythonInterpreterStatusBarItem.text = "Invalid python environment";
+        console.error(`The specified python interpreter path in settings does not exist: ${interpreterPath} `)
+      }
       this.pythonInterpreterStatusBarItem.backgroundColor = new ThemeColor(
         "statusBarItem.warningBackground"
       );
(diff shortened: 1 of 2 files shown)


src/features/pythonMetadata.ts
```diff
@@ -71,9 +72,9 @@ export class PythonInterpreterManager {
     );
     this.pythonInterpreterStatusBarItem.show();
 
-    if (this.extensionSettings.settings.interpreterPath) {
-      const interpreterPath = this.extensionSettings.settings.interpreterPath;
-      const label = this.makeLabelFromPath(interpreterPath);
+    const interpreterPath = this.extensionSettings.settings.interpreterPath;
+    if (interpreterPath && fs.existsSync(interpreterPath)) {
```
TamiTakamiya, line 76: When fs.existsSync(interpreterPath))  returns false, I think we'd better log which interpreter path was not found to console.error.

TamiTakamiya (approved):
LGTM. I left one minor comment.

robinbobbitt:
> LGTM. I left one minor comment.

Good call @TamiTakamiya . Done.

ganeshrn (approved):

PR #1082, opened 2024-02-06
handle command in python interpreter path

robinbobbitt:
Follow-up to https://github.com/ansible/vscode-ansible/pull/1077
Thanks @TamiTakamiya for discovering my previous fix doesn't work if the python interpreter path just contains the command e.g. "python" or "python3". This new approach handles that and saves us an extra sync call.

Tested with valid absolute path, valid command, invalid values, and no value.


TamiTakamiya (approved):
LGTM. Thanks!

ganeshrn (approved):

Issue #1084, opened 2024-02-07
Getting "Cannot set properties of undefined (setting 'currentModelValue')" error message while trying to connect in Lightspeed


AlexonOliveiraRH:
### Summary

I'm not being able to login with Ansible Lightspeed using the Red Hat Ansible extension due a `Cannot set properties of undefined (setting 'currentModelValue')` error message.

If uninstall VSCode and install it again, it works, but I can't open any folder or workspace before trying to connect to Lightspeed the first time. If I open any folder/workspace afterwards, it stops working and I cannot connect to it again.

Environment info:

```
$ sudo rpm -qa | grep ^code
code-1.86.0-1706698221.el8.x86_64

$ cat /etc/system-release
Red Hat Enterprise Linux release 8.9 (Ootpa)

$ uname -r
4.18.0-513.11.1.el8_9.x86_64
```

Steps to reproduce it:

1. If you have VSCode installed, uninstall it and remove every remain.
2. Reinstall VSCode and install/enable Red Hat Ansible extension.
3. Before open any folder or workspace, try to connect to Ansible Lightspeed. It should work.
4. Now open any folder or workspace.
5. Get back to the Ansible Lightspeed connector. It will be disconnected.
6. Try to connect again and it won't be possible. You'll get the described error message.

![Screenshot from 2024-02-07 10-45-49](https://github.com/ansible/vscode-ansible/assets/45946599/0477dbe1-d567-4402-92c5-4a9651ead297)
![Screenshot from 2024-02-07 10-46-13](https://github.com/ansible/vscode-ansible/assets/45946599/da455e8c-d964-47dc-bbef-a5fd55ed7f12)
![Screenshot from 2024-02-07 10-46-57](https://github.com/ansible/vscode-ansible/assets/45946599/24fb1567-0b7b-44cb-b5ca-faa0aa806a73)
![Screenshot from 2024-02-07 10-55-14](https://github.com/ansible/vscode-ansible/assets/45946599/b785e8cd-c480-4ce4-a97d-82aaa0d3306d)

### Extension version

v2.10.130

### VS Code version

1.86.0

### Ansible Version

```shell
$ ansible --version
ansible [core 2.13.7]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/alolivei/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/alolivei/.local/lib/python3.8/site-packages/ansible
[...]

PR #1085, opened 2024-02-07
avoid an exception during init if 'roles' is not a directory

goneri:
When the `roles` directory of the local workspace is not a regular directory,
`readdirSync()` raises an exception errcode (`ENOTDIR`).
The error was not catched and this was preventing the initialization of the
extension.

The commit also move a couple of functions to be able to properly unit-test
the change. We cannot import from the unit-test any module that use the
`vscode` object.

Finally, the commit changes the `LightSpeedManager.resetContext()` method to
make it a synchronous, this because the method changes an inner state of the
class that is consumed by other method.

Closes: #1084

### DIFF
--- a/package.json
+++ b/package.json
@@ -729,6 +729,7 @@
     "@types/mocha": "^10.0.6",
     "@types/node": "^20.11.7",
     "@types/sinon": "^17.0.3",
+    "@types/tmp": "^0.2.6",
     "@types/uuid": "^9.0.8",
     "@types/vscode": "^1.85.0",
     "@types/vscode-webview": "^1.57.4",
--- a/src/features/lightspeed/utils/updateRolesContext.ts
+++ b/src/features/lightspeed/utils/updateRolesContext.ts
@@ -1,0 +1,109 @@
+import * as path from "path";
+import * as fs from "fs";
+
+import {
+  IVarsContext,
+  IWorkSpaceRolesContext,
+} from "../../../interfaces/lightspeed";
+import { readVarFiles } from "./readVarFiles";
+import { VarType } from "../../../interfaces/lightspeed";
+
+function getVarsFromRoles(
+  rolePath: string,
+  varType: VarType
+): IVarsContext | undefined {
+  const varsRootPath = path.join(rolePath, varType);
+  if (!fs.existsSync(varsRootPath)) {
+    return;
+  }
+  let dirContent;
+  try {
+    dirContent = fs.readdirSync(varsRootPath);
+  } catch (err) {
+    console.error(`Failed to read a var directory with error ${err}`);
+    return;
+  }
+
+  const varsFiles =
+    dirContent
+      .filter((name) => [".yml", ".yaml"].includes(path.extname(name)))
+      .map((name) => name)
+      .map((name) => path.join(varsRootPath, name)) || [];
+  for (const varsFile of varsFiles) {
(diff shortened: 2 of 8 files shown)


src/features/lightspeed/utils/updateRolesContext.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+  let roleNames: string[] = [];
+  try {
+    roleNames = fs
+      .readdirSync(rolesRootPath)
+      .filter((name) =>
+        fs.statSync(path.join(rolesRootPath, name)).isDirectory()
+      );
+  } catch (error) {
+    console.error(`Cannot read the directory ${rolesRootPath}: ${error}`);
```
goneri, line 62: The fix is the try/catch here.

package.json
```diff
@@ -819,7 +820,7 @@
     "mock-lightspeed-server": "DEBUG='express:*' node ./out/client/test/mockLightspeedServer/server.js",
     "kill-mock-lightspeed-server": "pkill -f mockLightspeedServer/server.js"
   },
-  "version": "2.10.0",
+  "version": "2.123.123",
```
TamiTakamiya, line 823: Please keep version to "2.10.0".  It is updated at release time.

src/features/lightspeed/utils/updateRolesContext.ts
```diff
@@ -0,0 +1,97 @@
...
+  rolePath: string,
+  varType: VarType
+): IVarsContext | undefined {
+  const varsRootPath = path.join(rolePath, varType);
+  if (!fs.existsSync(varsRootPath)) {
+    return;
+  }
+  const varsFiles =
+    fs
+      .readdirSync(varsRootPath)
```
TamiTakamiya, line 21: The bug found was caused by a regular file named `roles`.  Can't a regular file `vars` cause a similar problem here? I think we'd better add try/catch block here as well.

src/features/lightspeed/utils/updateRolesContext.ts
```diff
@@ -0,0 +1,97 @@
...
+    ansibleRolesCache[workSpaceRoot] = {};
+  }
+  const rolesContext = ansibleRolesCache[workSpaceRoot];
+  rolesContext[rolePath] = {
+    name: rolePath.split(path.sep).pop(),
+  };
+
+  // Get all the task files in the tasks directory
+  const tasksPath = path.join(rolePath, "tasks");
+  if (fs.existsSync(tasksPath)) {
```
TamiTakamiya, line 85: The same comment as above: Can't a regular file `tasks` cause a similar problem here? I think we'd better add try/catch block here as well.

TamiTakamiya (changes requested):
@goneri Thank for looking at this bug. I left three comments.

src/features/lightspeed/utils/updateRolesContext.ts
goneri: Yes! You're right. Let's me cover this case.

src/features/lightspeed/utils/updateRolesContext.ts
goneri: +1

TamiTakamiya (approved):
LGTM. Thanks!

ganeshrn (approved):
Thank you!

PR #1088, opened 2024-02-09
Re-enable Lightspeed UI tests

TamiTakamiya:
Re-enable Lightspeed UI tests, which have been commented out and not executed for a while. Changes include:

- Applied changes corresponding to the latest Lightspeed code base (e.g. side panel section name: "Ansible Lightspeed" --> "Ansbile Lightspeed Login")
- Replaced a few vscode-extension-tester fucntions, which seemed not working as expected, with findElement/findElements calls with XPath locators. Although this might be somehow controversial one, I think this is a realistic approach for writing reliable UI tests for VS Code now.
- Intentionally kept original codes for the reference purpose. We may want to clean them up in the future.

Codes were tested with `yarn test-ui-current` command.  `yarn test-ui-oldest` fails due to current extension's vscode version requirement.

### DIFF
--- a/src/features/lightspeed/statusBar.ts
+++ b/src/features/lightspeed/statusBar.ts
@@ -48,6 +48,7 @@ export class LightspeedStatusBar {
     );
     lightSpeedStatusBarItem.command =
       LightSpeedCommands.LIGHTSPEED_STATUS_BAR_CLICK;
+    lightSpeedStatusBarItem.text = this.getLightSpeedStatusBarTextSync();
     this.getLightSpeedStatusBarText().then((text) => {
       lightSpeedStatusBarItem.text = text;
     });
@@ -66,12 +67,23 @@ export class LightspeedStatusBar {
       rhOrgHasSubscription =
         await this.lightSpeedAuthProvider.rhOrgHasSubscription();
     }
+    return this.getLightSpeedStatusBarTextSync(
+      rhOrgHasSubscription,
+      rhUserHasSeat
+    );
+  }
+
+  private getLightSpeedStatusBarTextSync(
+    rhUserHasSeat?: boolean,
+    rhOrgHasSubscription?: boolean
+  ): string {
     const userTypeLabel = getUserTypeLabel(
       rhOrgHasSubscription,
       rhUserHasSeat
     ).toLowerCase();
     return `Lightspeed (${userTypeLabel})`;
   }
+
   private handleStatusBar() {
     if (!this.client.isRunning()) {
       return;
--- a/src/features/lightspeed/statusBar.ts
+++ b/src/features/lightspeed/statusBar.ts
@@ -5,6 +5,7 @@ import { SettingsManager } from "../../settings";
 import {
   LightSpeedCommands,
   LIGHTSPEED_MODEL_DEFAULT,
+  LIGHTSPEED_STATUS_BAR_TEXT_DEFAULT,
 } from "../../definitions/lightspeed";
 import { LightSpeedAuthenticationProvider } from "./lightSpeedOAuthProvider";
 import { LightspeedAuthSession } from "../../interfaces/lightspeed";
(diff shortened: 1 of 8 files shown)


test/ui-test/lightspeedUiTest.ts
```diff
@@ -49,15 +49,25 @@ export function lightspeedUIAssetsTest(): void {
...
-        });
+      const welcomeContent =
+        await lightspeedServiceSection.findWelcomeContent();
+      expect(welcomeContent).not.undefined;
+      const loginButton = await welcomeContent?.findElement(
+        By.xpath("//a[.//span/text()='Connect']")
+      );
+      expect(loginButton).not.undefined;
+
+      // **** ORIGINAL CODE ****
```
ganeshrn, line 60: Do we need the old code commented out? If not this can be removed and instead add some comments to explain the new approach.

ssbarnea: I fully agree here, we should almost never have code commented. We have git history for that.

src/features/lightspeed/statusBar.ts
```diff
@@ -48,6 +48,7 @@ export class LightspeedStatusBar {
     );
     lightSpeedStatusBarItem.command =
       LightSpeedCommands.LIGHTSPEED_STATUS_BAR_CLICK;
+    lightSpeedStatusBarItem.text = this.getLightSpeedStatusBarTextSync();
```
ganeshrn, line 51: Why is this assignment added, when the text is assigned below in an async call using `then`?

TamiTakamiya: When I sent the first review of this PR, it was not needed.  But after removing the comment lines, the status bar UI tests ([these ones](https://github.com/ansible/vscode-ansible/blob/main/test/ui-test/lightspeedUiTest.ts#L89-L116)) started failing and I found initializing the status bar text with a synchronous method fixed the issue. Probably some timining or call sequnce has been changed. Even though this line might not be needed in real world use cases as the text will be set by the async call eventually, it shouldn't hurt anything and I think it is good to initialize the value without expecting the value from the async call.

ganeshrn: can we set the text to a default value instead of an sync call? 
```
 lightSpeedStatusBarItem.text = "Lightspeed"
```

TamiTakamiya: Yes, it is definitely possible, but an additional change may be required here.  With #1064, I changed the code to process only two user types: "Licensed" or "Unlicensed" based on inputs from Marty. The logic is very simple and it determines that the user is "Unlicensed" if she/he is not "Licensed". So with the latest code, "Lightspeed (unlicensed)" is displayed by default even before login. If the default is "Lightspeed", a small change needs to be added to check whether a session is already established or not. I will double-check Marty on that point before changing the code. Thanks.

ganeshrn: Thanks for following up on this. Apart from this comment the PR looks good to be merged.

TamiTakamiya: I have discussed with Marty on this and he thinks "Lightspeed (not logged in)" should be displayed when the user is not logged in yet. I agree with him that it looks better than "Lightspeed (unlicensed)" or "Ligntspeed". I have defined it in [definitions/lightspeed.ts](https://github.com/ansible/vscode-ansible/blob/e26c599e5528ecba64eede4ffd15fd7f132a8dad/src/definitions/lightspeed.ts#L65) in the latest commit. Would you take a look?

@goneri The latest commit also contained[ the change](https://github.com/ansible/vscode-ansible/blob/e26c599e5528ecba64eede4ffd15fd7f132a8dad/src/features/lightspeed/utils/webUtils.ts#L85-L96) that[ you had commented before](https://github.com/ansible/vscode-ansible/pull/1062/commits/70182c9a0e4a682488be7131f3573bc419b7b856#r1465599842). Although I opened an issue for that, now I think that we'd better apply the original change I made because the  `getLoggedInSessionDetails` function is called before the user logs into Lightspeed and it needs to be able to handle the case when `sessionData` is undefined. Let me know if you have different thoughts. Thanks.

ganeshrn (approved):

src/features/lightspeed/statusBar.ts
TamiTakamiya: @goneri Would you take a minute to look at [my comment](https://github.com/ansible/vscode-ansible/pull/1088#discussion_r1491264803 )on #1088? Although I think it would be ok, I would like you to take a look before merging the PR.

goneri (approved):

PR #1099, opened 2024-02-16
avoid duplicated suggestionFeedback event

goneri:
The `inlineSuggestionData` is available only when the user requests a
suggestion. The rest of the time, the structure is empty.
We now read the `suggestionId` used in the suggestionFeedback from there.
This way, once the suggestion is accepted, there is no way we can sent new
associated events.

### DIFF
--- a/src/features/lightspeed/inlineSuggestions.ts
+++ b/src/features/lightspeed/inlineSuggestions.ts
@@ -767,6 +767,9 @@ export async function inlineSuggestionCommitHandler() {
   if (!editor) {
     return;
   }
+  if (!inlineSuggestionData["suggestionId"]) {
+    return;
+  }
 
   // Commit the suggestion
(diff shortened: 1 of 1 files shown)


TamiTakamiya (approved):
LGTM. Thanks!

Though this is not directly related to the original issue, if inlineSuggestionData["suggestionId"] contains a value on line 350, do you think we can assume inlineSuggestionFeedback event was not sent for that suggestionId? If we can assume that, I want to send a inlineSuggestionFeedback with IGNORED action and it would reduce the mismatch between inlineSuggestionFeedback and completions events we are observing. Thanks.

ganeshrn (commented):
Can we add a testcase here to check the feedback back api is called only once on multiple tab presses. There are some example test case in repo to execute a command and check is a function is invoked or not.

ganeshrn (approved):

PR #1105, opened 2024-02-26
No inlineSuggestionFeedback if user accepted it on widget

TamiTakamiya:
For [AAP-21026](https://issues.redhat.com/browse/AAP-21026)


test/mockLightspeedServer/me.ts
```diff
@@ -5,6 +5,6 @@ export function me() {
     rh_user_is_org_admin: true,
     external_username: "EXTERNAL_USERNAME",
     username: "USERNAME",
-    org_telemetry_opt_out: true,
+    org_telemetry_opt_out: false,
```
TamiTakamiya, line 8: This is for enabling feedback events when the mock lightspeed server is used.

test/mockLightspeedServer/server.ts
```diff
@@ -48,6 +48,7 @@ export default class Server {
       res.send({
         access_token: "ACCESS_TOKEN",
         refresh_token: "REFRESH_TOKEN",
+        expires_in: 3600,
```
TamiTakamiya, line 51: In order to prevent calling token refresh when the mock lightspeed server is used.

src/features/lightspeed/inlineSuggestions.ts
```diff
@@ -33,6 +33,7 @@ import { getAdditionalContext } from "./inlineSuggestion/additionalContext";
 let inlineSuggestionData: InlineSuggestionEvent = {};
 let inlineSuggestionDisplayTime: Date;
 let previousTriggerPosition: vscode.Position;
+let currentSuggestion: string | undefined;
```
goneri, line 36: It would feel more natural to me to store the whole answer (`CompletionResponseParams` type), instead of just one single field. Also we can above the `undefined` and just initalize the value with an empty `CompletionResponseParams`.
Even if we use one single attribute, this would reduce the number of objects in use at the same time.

The fact we rely on these global variables is a bit annoying, but I don't see an easy solution to avoid that.

TamiTakamiya: @goneri  Thanks for your comments. I have looked at the code more closely and now think that we'd better make `insertTexts` global instead of `currentSuggestion`.  In the initial code, I found I needed to run `trim()` before comparing strings, but that was because we call `adjustInlineSuggestionIndent()` before setting `insertTexts`.  By using `insertTexts` we can get rid of `trim()`, which I think would generate incorrect results depending on indents in the code. I will update the code in that way.

> The fact we rely on these global variables is a bit annoying, but I don't see an easy solution to avoid that.

I do not like dependence on global variables either. I am feeling these coding is more like application development on MS-DOS :-)

src/features/lightspeed/inlineSuggestions.ts
```diff
@@ -835,5 +823,54 @@ export async function inlineSuggestionUserActionHandler(
...
+  if (
+    inlineSuggestionPending() &&
+    insertTexts &&
+    e.document.languageId === "ansible" &&
+    e.contentChanges.length > 0
+  ) {
+    const suggestionId = inlineSuggestionData["suggestionId"];
+
+    e.contentChanges.forEach(async (c) => {
+      if (c.text === insertTexts[0]) {
```
ganeshrn, line 864: Does this work for both single task and multi-task correctly? iirc there is some extra handling done for single task suggestion.

TamiTakamiya: As far as I tested it, it seems working both for single & multi-task... Do you remember what was the extra handling for single task suggestion?

src/features/lightspeed/inlineSuggestions.ts
```diff
@@ -835,5 +823,61 @@ export async function inlineSuggestionUserActionHandler(
...
+  }
+  return true;
+}
+
+function resetSuggestionData(): void {
   inlineSuggestionData = {};
+  insertTexts = [];
+}
+
+export async function textDocumentChangeHandler(
```
goneri, line 848: I think the function name is a bit too generic and we can be more explicit about its relationship with the feedback mechanism.

goneri (approved):

src/features/lightspeed/inlineSuggestions.ts
```diff
@@ -835,5 +823,61 @@ export async function inlineSuggestionUserActionHandler(
...
+  ) {
+    const suggestionId = inlineSuggestionData["suggestionId"];
+
+    e.contentChanges.forEach(async (c) => {
+      if (c.text === insertTexts[0]) {
+        // If a matching change was found, send a feedback with the ACCEPTED user action.
+        console.log(
+          "[inline-suggestions] Detected a text change that matches to the current suggestion."
+        );
+        await inlineSuggestionUserActionHandler(
```
ganeshrn, line 876: If user accepts suggestion by TAB key will this method be called again? In the `inlineSuggestionCommitHandler` we already send a accept feedback. If the `textDocumentChangeHandler` is invoked after TAB key press as well in that case we are duplicate the accept feedback?

src/features/lightspeed/inlineSuggestions.ts
ganeshrn: For single task we add the task name to the accepted suggestion here https://github.com/ansible/vscode-ansible/blob/24af91bde2cbc1293b38dcff11018c8180f32f33/src/features/lightspeed/inlineSuggestions.ts#L423

src/features/lightspeed/inlineSuggestions.ts
ganeshrn: Got it

src/features/lightspeed/inlineSuggestions.ts
TamiTakamiya: If I read the code correctly, I think the additional task name added to `contentMatchesForSuggestion` is used only for content matches, not stored in `inlineSuggestionUserActionItems`. I think we can depend on the value in `insertTexts` for string comparison in text change handler. Or the `contentMatchesForSuggestion` would somehow influent text changes made by inline suggestion?

src/features/lightspeed/inlineSuggestions.ts
```diff
@@ -841,5 +840,61 @@ export async function inlineSuggestionUserActionHandler(
...
+  if (
+    inlineSuggestionPending() &&
+    insertTexts &&
+    e.document.languageId === "ansible" &&
+    e.contentChanges.length > 0
+  ) {
+    const suggestionId = inlineSuggestionData["suggestionId"];
+
+    e.contentChanges.forEach(async (c) => {
+      if (c.text === insertTexts[0]) {
```
ganeshrn, line 888: Also can we reuse the match logic to check is suggestion is coming from lighspeed and not from any other provider by adding the check here https://github.com/ansible/vscode-ansible/blob/main/src/features/lightspeed/inlineSuggestions.ts#L772. This check will also solve the issue of contentmatches API triggering for suggestion not coming from lightspeed.

If the document change handler is not triggered after hitting TAB within the `inlineSuggestionCommitHandler` method another approach is to compare the document content before the inline suggestion trigger request with the document content after inline suggestion is accepted and the diff between the two should be same as the one received in lightspeed completion response.

TamiTakamiya: At https://github.com/ansible/vscode-ansible/blob/main/src/features/lightspeed/inlineSuggestions.ts#L772 , is there a handy way to see which text is being inserted with the commit command? If it is possible to get the currently displayed text, I think it is a good idea to add the match logic there...

ganeshrn: I was not able to get the inserted text in the commit command handler through a callback approach but was to get the entire file content with inserted text using below snippet in commit command handler
```
const activeEditor = vscode.window.activeTextEditor;

if (activeEditor) {
    const document = activeEditor.document;
    const text = document.getText();
    console.log(text);
} 
```

That's the reason I mentioned another approach is to compare the document content before the inline suggestion trigger request with the document content after inline suggestion is accepted and the diff between the two should be same as the one received in lightspeed completion response in the earlier comment.

I think this comment can be handled in the follow-up PR

ganeshrn (approved):

src/features/lightspeed/inlineSuggestions.ts
TamiTakamiya: Ganesh, thanks for the explanation. Ok, let's handle this in the follow-up PR.

PR #1111, opened 2024-03-01
send ansible extension version in lightspeed requests

robinbobbitt:
Adds the extension version to Lightspeed API requests for telemetry and troubleshooting purposes.


ganeshrn (approved):

TamiTakamiya:
@robinbobbitt @ganeshrn I think we may also want to change HTTP "User-Agent" header to include this kind of information.  Currently I think we are using the default one provided by the axios npm library and it is not very usuful.

PR #1115, opened 2024-03-04
Do not send ACCEPTED feedback from commit handler

TamiTakamiya:
When there are more than one inline suggestions are available and user selected a suggestion, which was not from Lightspeed, we should not send a feedback with the ACCEPTED action. As the current code assumes that user accepted the Lightspeed suggestion when user pressed Tab key,

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -257,6 +257,8 @@ export async function activate(context: ExtensionContext): Promise<void> {
             editor.document,
             AnsibleContentUploadTrigger.TAB_CHANGE
           );
+        } else {
+          await ignorePendingSuggestion();
         }
       }
     )
--- a/src/features/lightspeed/inlineSuggestions.ts
+++ b/src/features/lightspeed/inlineSuggestions.ts
@@ -759,14 +759,6 @@ export async function inlineSuggestionTriggerHandler() {
 }
 
 export async function inlineSuggestionCommitHandler() {
-  if (vscode.window.activeTextEditor?.document.languageId !== "ansible") {
-    return;
-  }
-  const editor = vscode.window.activeTextEditor;
-  if (!editor) {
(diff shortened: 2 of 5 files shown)


src/extension.ts
```diff
@@ -257,6 +257,8 @@ export async function activate(context: ExtensionContext): Promise<void> {
             editor.document,
             AnsibleContentUploadTrigger.TAB_CHANGE
           );
+        } else {
```
TamiTakamiya, line 260: These two lines were incorrectly dropped from #1113 druing a merge operation.

src/features/lightspeed/inlineSuggestions.ts
```diff
@@ -924,6 +908,10 @@ export async function inlineSuggestionTextDocumentChangeHandler(
           suggestionId!,
           UserAction.ACCEPTED
         );
+        // Show training matches for the accepted suggestion.
```
TamiTakamiya, line 911: Moved those lines to `inlineSuggestionTextDocumentChangeHandler()` as content matches function should work when user accepted on a widget as well.

test/mockLightspeedServer/server.ts
```diff
@@ -25,12 +25,12 @@ export default class Server {
     app.get("/", (req, res) => res.send("Lightspeed Mock"));
 
     app.post(`${API_ROOT}/ai/completions`, async (req, res) => {
-      await new Promise((r) => setTimeout(r, 100)); // fake 100ms latency
+      await new Promise((r) => setTimeout(r, 500)); // fake 500ms latency
```
TamiTakamiya, line 28: Increased the fake latency so that the another inline suggestion provider shows its suggestion first.

ganeshrn (approved):

PR #1116, opened 2024-03-06
Send IGNORED feedback if feedback was pending

TamiTakamiya:
When an inline suggestion is requested and there is the previous suggestion, whose feedback has not been sent, send a feedback with IGNORED action.  This code will not be used in normal use cases, but I believe we should have such a fall-back mechanism.

### DIFF
--- a/test/testScripts/lightspeed/testLightspeed.test.ts
+++ b/test/testScripts/lightspeed/testLightspeed.test.ts
@@ -273,13 +273,13 @@ export function testLightspeed(): void {
       });
     });
 
+    describe("Test ignore pending suggestions", () => {
+      testIgnorePendingSuggestion();
+    });
+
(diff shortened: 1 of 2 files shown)


test/testScripts/lightspeed/testLightspeed.test.ts
```diff
@@ -17,6 +17,7 @@ import { lightSpeedManager } from "../../../src/extension";
 import {
   testInlineSuggestionByAnotherProvider,
   testInlineSuggestionProviderCoExistence,
+  testIgnorePendingSuggestion,
```
TamiTakamiya, line 20: This test case (`testIgnorePendingSuggestion`) was written for #1113, but it was not executed by mistake. I added it here to be included in e2e tests.

priyamsahoo (approved):

PR #1117, opened 2024-03-06
Add comment crediting Ansible Lightspeed to suggestion ghost text

robinbobbitt:
When users have more than one inline suggestion provider configured in their VS Code, it can be hard to know which suggestion came from Lightspeed. With this PR, we add "# Content suggestion provided by Ansible Lightspeed" as a first line in the ghost text. We also add a setting that determines if this comment should be retained in the suggestion when the user accepts it. Thank you to @ganeshrn for sharing a first cut of this feature.

Related: https://github.com/microsoft/vscode/issues/205555


TamiTakamiya (approved):
LGTM. Thanks.

PR #1122, opened 2024-03-08
Address 'User not authorized to access Ansible Lightspeed' when request is blocked by CloudFront

manstis:
See https://issues.redhat.com/browse/AAP-21321

Fixes: AAP-21321


src/features/lightspeed/handleApiError.ts
```diff
@@ -20,38 +20,42 @@ export function retrieveError(err: AxiosError): string {
         return "Bad Request response. Please try again.";
       }
     } else if (err?.response?.status === 403) {
```
manstis, line 22: The change is not as bad as it looks.

I simply check for a CloudFront error otherwise let the existing 403 handling run:
```
...
    } else if (err?.response?.status === 403) {
      if((err?.response?.headers["server"] || "").toLowerCase() === "cloudfront") {
        return `CloudFront rejected the request. Please contact your administrator.`;
      } else {
        const responseErrorData = <AxiosError<{ message?: string }>>(
          err?.response?.data
        );
...
```

manstis:
> Label error. Requires exactly 1 of: bug, enhancement, major, minor, patch, skip-changelog

I don't have permission to add labels.

src/features/lightspeed/handleApiError.ts
```diff
@@ -20,38 +20,44 @@ export function retrieveError(err: AxiosError): string {
...
-        responseErrorData.hasOwnProperty("message") &&
-        responseErrorData.message?.includes("WCA Model ID is invalid")
+        (err?.response?.headers["server"] || "").toLowerCase() === "cloudfront"
       ) {
-        return `Model ID is invalid. Please contact your administrator.`;
-      } else if (responseErrorData) {
-        switch (
-          responseErrorData.hasOwnProperty("code") &&
-          responseErrorData.code
+        return `CloudFront rejected the request. Please contact your administrator.`;
```
robinbobbitt, line 26: Are you thinking contact your admin so they can contact support? Ultimately the admin needs to reach out to Red Hat support.

manstis: I used the same message as we use elsewhere for consistency (e.g. Cloudflare exceptions). Happy to change (them all?) if that's the preference.

robinbobbitt: The "Please contact your administrator." makes sense in the case of the invalid model id, since the admin presumably has access to the correct value and/or handed them the wrong value. Here though the admin can't do anything except point them to support. How about something like "Something in your editor content has caused your inline suggestion request to be blocked. Please open a ticket with Red Hat support and include the content of your editor up to the line and column where you requested a suggestion."

manstis: Changed to match your proposal... I'm not one to procrastinate on the wording; IMO it's very subjective.

ssbarnea (approved):

ssbarnea:
@manstis @robinbobbitt Now you have write rights. Use them with care. Don't forget to resolve conversations.

manstis:
@ssbarnea The same errors :sob:

src/features/lightspeed/handleApiError.ts
robinbobbitt: Agree. I'm mostly just glad to recognize this as a CloudFront issue now. Thanks Michael!

robinbobbitt (approved):
Thanks @manstis! LGTM.

manstis:
Sorry @ssbarnea could you run the failing workload again please?

PR #1130, opened 2024-03-13
WIP: Attempt to reduce activation wait time

TamiTakamiya:
This is an experiment to reduce the 20 sec wait time used in e2e test.


TamiTakamiya:
I think I could obtain good amount of information through these experiments. I will close this and open a new PR for sending it for review.

PR #1131, opened 2024-03-13
AAP-21329: Capture relevant model in feedback events

romartin:
See https://issues.redhat.com/browse/AAP-21329

This PR just send the model value from settings, if set, on the feedback requests. This way, given this another PR https://github.com/ansible/ansible-wisdom-service/pull/874, this information will flow into the wisdom's telemetry events.


romartin:
@TamiTakamiya @ganeshrn Working on the test case, but are failing for me locally, still looking...

romartin:
@ganeshrn @TamiTakamiya test case added as well. Ready for review, please. Thanks!!

src/features/lightspeed/api.ts
```diff
@@ -176,6 +177,12 @@ export class LightSpeedAPI {
     }
     const rhUserHasSeat = await this.lightSpeedAuthProvider.rhUserHasSeat();
 
+    const userProvidedModel =
+      lightSpeedManager.settingsManager.settings.lightSpeedService.model;
```
goneri, line 181: It's a bit of complexity just to avoid the situation where the model is undefined or empty. What the problem if you send an empty model value? Something like this:

```
inputData.model = lightSpeedManager.settingsManager.settings.lightSpeedService.model;
```

TamiTakamiya (approved):
LGTM. Thanks!

src/features/lightspeed/api.ts
romartin: that's true!! I was just copy pasting from another place, but that's true, no need for that complexity. Fixed. Thanks!

goneri (approved):

ganeshrn (approved):

ganeshrn (approved):

PR #1133, opened 2024-03-13
Playbook explanation and generation UI

goneri:
This PR introduces the early UI for two new features for Lightspeed. The code is not fully operational yet, but we would like to start to merge the elements that we feel are ready.
These new UI are not exposed to the users.

This PR also introduces a new LSP server that is dedicated to Lightspeed. Our goal is to use this server as an interface between the editor and the AI/model interaction.

In the future, we would like to:
- also move into this server the functions that are used for the task and multiline predictions.
- consolidate everything in the existing LSP server (ansible-language-server)

Co-Author: Tami Takamiya <ttakamiy@redhat.com>


goneri:
@ganeshrn @ssbarnea this PR add our new UI, there is already quiet a lot of changes, but nothing is enabled by default. We would like to get this PR land to avoid building a gigantic PR that we will have to rebase again and again. The PR slightly reduces the coverage, we are aware of that and already working on new vscode-extension-tester tests.

ganeshrn (approved):

PR #1163, opened 2024-03-20
remove extra whitespace after lightspeed suggestion

robinbobbitt:
Currently when adding tasks via Lightspeed, the cursor is automatically indented in the following row to the appropriate column for the content of the task. However, it's expected that most likely a user is moving from prompt to prompt/task to task, rather than line by line manual development. As a result, they end up with a number of tasks with lines between them containing a bunch of whitespace, whereas ansible-lint expects them to be empty. This PR removes the extra whitespace. It also removes a stray `vscode.commands.executeCommand("editor.action.inlineSuggest.trigger");` that should not have been in this function originally. This PR is most easily reviewed with whitespace ignored (`?w=1`).

### DIFF
--- a/src/features/lightspeed/inlineSuggestions.ts
+++ b/src/features/lightspeed/inlineSuggestions.ts
@@ -789,6 +789,45 @@ export async function inlineSuggestionReplaceMarker(position: vscode.Position) {
     return;
   }
 
+  console.log(
+    `[inline-suggestions] Inline Suggestion Marker Handler triggered using command at ${position.line}`
+  );
+
+  if (
+    lightSpeedManager.settingsManager.settings.lightSpeedService
+      .disableContentSuggestionHeader
+  ) {
+      // Get the current text
+      const line = position.line;
+
+      const text = editor.document.getText(
+        new vscode.Range(
+          new vscode.Position(line, 0),
+          new vscode.Position(line + 1, 0)
+        )
+      );
+      // Remove the prepended text
+      const commentPosition = text.indexOf(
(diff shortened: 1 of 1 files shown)


src/features/lightspeed/inlineSuggestions.ts
```diff
@@ -827,6 +824,26 @@ export async function inlineSuggestionReplaceMarker(position: vscode.Position) {
...
+    "[inline-suggestions] Inline Suggestion Marker Handler removing extra whitespace"
+  );
+
+  // Clear the line of extra whitespace after the suggestion
+  // that causes ansible-lint errors
+  const selection = editor.selection;
+  const lineText = editor.document.lineAt(selection.active.line).text;
+  if (/^\s*$/.test(lineText)) {
+    editor.edit((editBuilder) => {
+      editBuilder.delete(
```
TamiTakamiya, line 837: I think you want to add `await` in front of `editor.edit` as you modified line 818.

TamiTakamiya (commented):
I left one comment. Would you take a look?

src/features/lightspeed/inlineSuggestions.ts
robinbobbitt: Thanks Tami. This one I was thinking it didn't matter really when it completed since nothing depends on it, but for consistency and to avoid surprises later, yes, let me take a look at that. Thank you!

robinbobbitt: Done. Thanks!

TamiTakamiya (approved):
LGTM

ssbarnea (approved):

PR #1166, opened 2024-03-26
AAP-19001: No training matches found for the latest accepted suggestion.

manstis:
See https://issues.redhat.com/browse/AAP-19001

**Suggestion not set - this should not happen**

suggestion-not-set

**No codematches found**

codematches-not-found

**Codematch returns an error (string)**

In this example the URL for codematch was not found and a HTTP404 was returned.

The "Details" content is what was returned from a network device somewhere (e.g. CloudFront, nginx etc). 

I added "Content-Security-Policy" [metadata](https://github.com/ansible/vscode-ansible/pull/1166/files#diff-198a5926cd0cb3ec2d793da7b74c1d00b0b72abd8bc38a2f2b44022c7aedd4aaR160-R161) so we _should_ be safe rendering whatever we get). 

codematches-error-string

**Codematch returns an error (object)**

codematches-error-object

### DIFF
--- a/src/features/lightspeed/api.ts
+++ b/src/features/lightspeed/api.ts
@@ -220,7 +220,12 @@ export class LightSpeedAPI {
     } catch (error) {
       const err = error as AxiosError;
       const mappedError: IError = mapError(err);
-      vscode.window.showErrorMessage(mappedError.message ?? UNKNOWN_ERROR);
+      const errorMessage: string = mappedError.message ?? UNKNOWN_ERROR;
+      if (showInfoMessage) {
+        vscode.window.showErrorMessage(errorMessage);
+      } else {
+        console.error(errorMessage);
+      }
       return {} as FeedbackResponseParams;
     }
   }
@@ -262,7 +267,6 @@ export class LightSpeedAPI {
     } catch (error) {
       const err = error as AxiosError;
       const mappedError: IError = mapError(err);
-      vscode.window.showErrorMessage(mappedError.message ?? UNKNOWN_ERROR);
       return mappedError;
     }
   }
--- a/src/features/lightspeed/api.ts
+++ b/src/features/lightspeed/api.ts
@@ -9,6 +9,7 @@ import {
   FeedbackResponseParams,
   ContentMatchesRequestParams,
   ContentMatchesResponseParams,
+  IError,
 } from "../../interfaces/lightspeed";
 import {
   LIGHTSPEED_SUGGESTION_CONTENT_MATCHES_URL,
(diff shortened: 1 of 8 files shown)


src/features/lightspeed/contentMatchesWebview.ts
```diff
@@ -107,22 +108,82 @@ export class ContentMatchesWebview implements vscode.WebviewViewProvider {
...
+  ): contentMatchResponses is IError {
+    return (contentMatchResponses as IError).code !== undefined;
+  }
+
   private async getWebviewContent(): Promise<string> {
-    const noContentMatchesFoundHtml = `<html><body>No training matches found for the latest accepted suggestion.</body></html>`;
+    const noActiveSuggestionHtml = `
+      <html>
+        <body>
+          <p>Training matches cannot be retrieved. No active suggestion found.</p>
```
robinbobbitt, line 121: This shows up when you first open the extension. Shall we drop the `Training matches cannot be retrieved.` and say something like `Lightspeed training matches will be displayed here after you accept a Lightspeed inline suggestion.` or `Accept an inline suggestion from Lightspeed to see training matches here.`

src/features/lightspeed/contentMatchesWebview.ts
```diff
@@ -107,22 +108,82 @@ export class ContentMatchesWebview implements vscode.WebviewViewProvider {
...
+        <head>
+          <!-- https://code.visualstudio.com/api/extension-guides/webview#content-security-policy -->
+          <meta http-equiv="Content-Security-Policy" content="default-src 'none';">
+        </head>
+        <body>
+        <p>An error occurred trying to retrieve the training matches.</p>
+        <p><b>Message:</b> ${error.message}</p>
+        <details>
+          <summary><b>Detail:</b></summary>
+          <p>${detail}</p>
```
robinbobbitt, line 168: We should exclude the details section if we don't have any. A 503 from contentmatches currently gives me this:
image

robinbobbitt: Shall we skip the lower right pop-up when we're already displaying the error in the panel?

src/features/lightspeed/api.ts
```diff
@@ -215,14 +219,15 @@ export class LightSpeedAPI {
       return response.data;
     } catch (error) {
       const err = error as AxiosError;
-      vscode.window.showErrorMessage(retrieveError(err));
+      const mappedError: IError = mapError(err);
+      vscode.window.showErrorMessage(mappedError.message ?? UNKNOWN_ERROR);
```
robinbobbitt, line 223: I'm now seeing this error popup as a seated user. It won't make sense to them, and we also don't ever want to pop a visible error for implicit feedback problems (inline suggestion, ansible content).
image

robinbobbitt (changes requested):

src/features/lightspeed/api.ts
manstis: What did you do to see this?

src/features/lightspeed/contentMatchesWebview.ts
manstis: I've changed to `Training matches will be displayed here after you accept an inline suggestion.`

I avoided mentioning `Lightspeed` to avoid problems with upstream/downstream.

src/features/lightspeed/contentMatchesWebview.ts
manstis: Fair point. Removed.

robinbobbitt (approved):
I tested a bunch of scenarios and everything is working great. Thanks @manstis !

PR #1168, opened 2024-03-27
AAP-22175: Make CloudFront check more robust

manstis:
See https://issues.redhat.com/browse/AAP-22175

### DIFF
--- a/src/features/lightspeed/errors.ts
+++ b/src/features/lightspeed/errors.ts
@@ -1,50 +1,86 @@
+import { AxiosError } from "axios";
 import { IError } from "../../interfaces/lightspeed";
 
 class Error implements IError {
   readonly code: string;
   readonly message?: string;
   readonly detail?: unknown;
+  readonly check: (err: AxiosError) => boolean;
 
-  public constructor(code: string, message?: string, detail?: unknown) {
+  private getCode(err: AxiosError): string {
+    const responseErrorData = <AxiosError<{ code?: string; message?: string }>>(
+      err?.response?.data
+    );
+    const code: string = responseErrorData.hasOwnProperty("code")
+      ? (responseErrorData.code as string)
+      : "unknown";
+    return code;
+  }
+
+  public constructor(
+    code: string,
+    message?: string,
+    detail?: unknown,
+    check?: (err: AxiosError) => boolean
+  ) {
     this.code = code;
     this.message = message;
     this.detail = detail;
-  }
-
-  public withMessage(message?: string): Error {
-    return new Error(this.code, message, this.detail);
+    if (check) {
+      this.check = check;
+    } else {
+      this.check = (err: AxiosError) => {
+        return this.code === this.getCode(err);
+      };
+    }
   }
 
   public withDetail(detail?: unknown): Error {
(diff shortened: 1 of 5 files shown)


src/features/lightspeed/errors.ts
```diff
@@ -1,50 +1,86 @@
...
     if (!this.errors.has(statusCode)) {
       this.errors.set(statusCode, []);
     }
     this.errors.get(statusCode)?.push(error);
   }
 
-  public getError(statusCode: number, code: string): Error | undefined {
-    if (!this.errors.has(statusCode)) {
+  public getError(err: AxiosError): Error | undefined {
+    if (err && !("response" in err)) {
```
robinbobbitt, line 55: Testing content match with my server down (ECONNREFUSED), I end up with this:

image

You can recreate this by working against your local service, accepting a completion request, taking your service down, then running "Fetch Training Matches" in the command palette.

src/features/lightspeed/handleApiError.ts
```diff
@@ -6,62 +6,39 @@ import {
...
+  }
+  if (status === 429) {
+    return ERRORS_TOO_MANY_REQUESTS.withDetail(detail);
+  }
+  if (status === 500) {
+    return ERRORS_UNKNOWN.withDetail(detail);
+  }
+  if (status === AxiosError.ECONNABORTED) {
     return ERRORS_CONNECTION_TIMEOUT.withDetail(detail);
   }
```
robinbobbitt, line 44: When we fall through to ERRORS_UNKNOWN, we should log the actual error in the console so we have a shot at helping users troubleshoot their problem. I don't mean to expand scope of this PR, but I'm not sure if we've possibly lost some detail with all the refactoring.. Either way, it would be great to get this added.. just something like:
```
  console.log(`Lightspeed request failed with unknown error ${err}`);
``` 
should suffice.
```
Lightspeed request failed with unknown error Error: connect ECONNREFUSED 127.0.0.1:8000
```

robinbobbitt (changes requested):
Thanks Michael! I tested a CloudFront 403 and non-CloudFront 403 and they looked good. I noticed just a couple other things to tweak with all the error refactoring. Thank you!

src/features/lightspeed/errors.ts
manstis: Oee. Did you pull the latest changes first?

You reported this `undefined` on my previous PR and I thought it was fixed in such a way that seeing `undefined` would be impossible.

manstis: Doh. I just checked the code on `main` and it's eminently possible to still show `undefined`.

src/features/lightspeed/handleApiError.ts
manstis: Done.

src/features/lightspeed/errors.ts
manstis: Done.

TamiTakamiya (approved):
LGTM

TamiTakamiya:
@robinbobbitt Would you take a look at the latest changes that @manstis made? With your approval, I think we can merge this into main. Thanks!

PR #1187, opened 2024-04-02
Update package versioning in Jenkins jobs

TamiTakamiya:
After moving to CalVer, we need to have a way to release multiple version in one month and still want to keep a build number in .vsix packaging. Following are my proposal on GitHub versions and package versioning to satisfy those requirements (ref:  [Semantic Versioning 2.0.0](https://semver.org/#semantic-versioning-200)):

1. GitHub version (and version in package.json): Will use a CalVer. The patch segment increments per release.  For example, `24.4.0`, `24.4.1`, `24.4.2`
1. Marketplace version (set by Jenkins job).
     - For pre-releases, add -alpha
     - Then add build number with preceding +
     For example, `24.4.0+180`, `24.4.1-alpha+181`, `24.4.1-alpha+182`, `24.4.1+183`
     
This PR contains a change in `Jenkinsfile` to implement the design.


ganeshrn (approved):

ssbarnea:
I am (almost) sure that vsce will refuse uploading packages that do not use a very strict pattern `x.y.z` with only numbers. At least this was what I found more than an year ago when I added the prereleases and wanted to use some kind of SemVer.

I will try to test marketplace upload before merging this (another extension i managed personally).

In the (near) future we should,
- Get rid of Jenkins job
- Alternatively modify Jenkins job to just download the artifact from GHA (waiting for it) and publishing it, no need to even clone the repository.

ssbarnea (changes requested):
Please read https://github.com/microsoft/vscode-vsce/issues/148

The only acceptable format for versions is X.Y.Z.W

TamiTakamiya:
@ssbarnea Thank you for your valuable comment. I will take a look at your PR #1189.

TamiTakamiya:
As Sorin noticed me, we are going to stop using Jenkins soon for publishing exxtensions. I am going to close this PR without merging.

PR #1195, opened 2024-04-03
Prepare for 24.4.0 release

TamiTakamiya:
Changes for the new VS Code Extension release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -858,7 +858,7 @@
     "coverage-all": "TEST_LIGHTSPEED_URL='http://127.0.0.1:3000' yarn _coverage-all",
     "als-compile": "yarn workspace @ansible/ansible-language-server compile"
   },
-  "version": "2.13.0",
+  "version": "24.4.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -858,7 +858,7 @@
     "coverage-all": "TEST_LIGHTSPEED_URL='http://127.0.0.1:3000' yarn _coverage-all",
     "als-compile": "yarn workspace @ansible/ansible-language-server compile"
   },
-  "version": "2.13.0",
+  "version": "24.4.0",
```
goneri, line 861: This format sounds wrong to me, "24.4.0", for instance "24.2.0" > "24.12.0". It would be better to use 20240403 instead.

goneri (changes requested):

package.json
goneri: Sorry, I'm wrong:

```
$ rpmdev-vercmp 25.02 25.12
25.2 < 25.12
```

goneri (approved):

PR #1214, opened 2024-04-05
Improve usability for token expiration scenario

TamiTakamiya:
For [AAP-20239](https://issues.redhat.com/browse/AAP-20239).

It is challenging to write test cases for this... I will spend some time before sending this for review. Until then, I would like to keep this as a draft.


TamiTakamiya:
This was not that simple. I will close this and re-consider the design next week.

PR #1223, opened 2024-04-11
Upgrade to eslint v9

ssbarnea:
- call eslint as a system hook instead of native pre-commit one
- migrate eslint config to flat format
- fix few new rules and temporary disable others

Related:
- https://eslint.org/docs/latest/use/migrate-to-9.0.0
- https://github.com/typescript-eslint/typescript-eslint/issues/8211
- https://github.com/ansible/vscode-ansible/issues/1225


src/features/lightspeed/inlineSuggestions.ts
```diff
@@ -876,9 +876,11 @@ export async function inlineSuggestionHideHandler(userAction?: UserAction) {
   switch (action) {
     case UserAction.REJECTED: {
       console.log("[inline-suggestions] User rejected the inline suggestion.");
+      break;
```
TamiTakamiya, line 879: Good catch!

src/features/lightspeed/inlineSuggestions.ts
```diff
@@ -876,9 +876,11 @@ export async function inlineSuggestionHideHandler(userAction?: UserAction) {
...
     }
     case UserAction.IGNORED: {
       console.log("[inline-suggestions] User ignored the inline suggestion.");
+      break;
```
TamiTakamiya, line 883: This one ,too!

src/features/lightspeed/inlineSuggestions.ts
ssbarnea: Need to send kudos to eslint folks, not me. There are also lots of other changes that we will need to make, but we can do them later, i marked them as warnings only, so we will be nagged on PRs about them, at least until we can mark them as errors.

priyamsahoo (approved):

TamiTakamiya (approved):
As far as Lightspeed-related files concern, all looked good. Thank you.

PR #1230, opened 2024-04-12
Support wait window setting for Lightspeed inline suggestion

TamiTakamiya:
For [AAP-19633](https://issues.redhat.com/browse/AAP-19633):

Add a wait window after a trigger (user pressing enter on matching the condition). If the user continues to type within the wait window the extension shouldn't send completion request.

The length of the wait window is configured using a new setting `ansible.lightspeed.suggestions.waitWindow` in milli-seconds. The default is set to 0, which let the extension send completion request immediately (=same as the current behavior).


package.json
```diff
@@ -500,17 +500,24 @@
             "markdownDescription": "Enable Ansible Lightspeed with watsonx Code Assistant inline suggestions.",
             "order": 2
           },
+          "ansible.lightspeed.suggestions.waitWindow": {
+            "scope": "resource",
+            "type": "number",
+            "default": 0,
```
ganeshrn, line 506: Do we want to set the default to a small value to start with?

TamiTakamiya: > Do we want to set the default to a small value to start with?

I brought this on our team's Slack channel and Marty and Robin agreed to use 0 as the default value. I think it is safe not to change the current default behavior and making the response time longer by default does not seem to be a good idea even if it is a few hundred milli-seconds.

goneri:
@TamiTakamiya for the record, I've just done a rebase of your branch.

goneri (approved):

priyamsahoo (approved):

PR #1236, opened 2024-04-15
LightspeedExplorer view as a Webview and new button

goneri:
Convert the "Ansible Lighspeed Login" panel from a TreeView to a WebView. The goal
is to be able to expose an addition button icon.

This commit also adds a new "Explain the current playbook" button to this view to
trigger a playbook explanation.
Convert the "Ansible Lighspeed Login" panel from a TreeView to a WebView. The goal
is to be able to expose an addition button icon.

This commit also adds a new "Explain the current playbook" button to this view to
trigger a playbook explanation

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -25,6 +25,7 @@
 !media/playbookGeneration/style.css
 !out/client/src/extension.js
 !out/client/webview/apps/lightspeed/main.js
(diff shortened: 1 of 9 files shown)


TamiTakamiya (approved):
LGTM!

PR #1244, opened 2024-04-16
lightspeedExplorer: minor UI adjustments

goneri:
- Add a tooltip to explain why we cannot explain a playbook
- Use a dedicated CSS file
- Align the view content slightly on the right


src/features/lightspeed/utils/explorerView.ts
```diff
@@ -29,15 +33,17 @@ export function getWebviewContentWithLoginForm(
...
-        <section class="component-section">
-          <vscode-button id="lightspeed-explorer-connect">Connect</vscode-button>
+    <div id="lightspeedExplorerView">
+      Welcome to Ansible Lightspeed for Visual Studio Code.\nExperience smarter automation using Ansible Lightspeed with watsonx Code Assistant solutions for your playbook. <a href="https://www.redhat.com/en/engage/project-wisdom">Learn more</a><br />Let's simplify your workflow by connecting VS Code with Ansible Lightspeed.<br />
+      <form id="playbook-explanation-form">
+        <div class="button-container">
+          <section class="component-section">
+            <vscode-button id="lightspeed-explorer-connect">Connect</vscode-button>
+          </section>
         </section>
```
TamiTakamiya, line 43: I think this `</section>` needs to be changed to `</div>`, which corresponds to line 39.

src/features/lightspeed/utils/explorerView.ts
```diff
@@ -29,15 +33,17 @@ export function getWebviewContentWithLoginForm(
       <title>Ansible Lightspeed Explorer!</title>
     </head>
     <body>
-    Welcome to Ansible Lightspeed for Visual Studio Code.\nExperience smarter automation using Ansible Lightspeed with watsonx Code Assistant solutions for your playbook. <a href="https://www.redhat.com/en/engage/project-wisdom">Learn more</a><br />Let's simplify your workflow by connecting VS Code with Ansible Lightspeed.<br />
-    <form id="playbook-explanation-form">
-      <div class="component-container">
-        <section class="component-section">
-          <vscode-button id="lightspeed-explorer-connect">Connect</vscode-button>
+    <div id="lightspeedExplorerView">
```
TamiTakamiya, line 36: As `lightspeedExplorerView` is defined as a class in `style.css`, I think it would be used with `class=` instead of `id=`

src/features/lightspeed/utils/explorerView.ts
```diff
@@ -15,7 +15,11 @@ export function getWebviewContentWithLoginForm(
     "explorer",
     "main.js",
   ]);
-  const styleUri = getUri(webview, extensionUri, ["media", "style.css"]);
+  const styleUri = getUri(webview, extensionUri, [
+    "media",
+    "lightspeedExplorerView",
```
TamiTakamiya, line 20: :+1: 

I think `.vscodeignore` file also needs to be updated to include `media/lightspeedExplorerView/style.css` in the .vsix package.

TamiTakamiya (changes requested):
I have left three comments. Would you take a look? Thanks!

TamiTakamiya (approved):

PR #1258, opened 2024-04-19
lightspeed - Auth refactoring

goneri:
Improve the isolation
---------------------

- `lightSpeedOAuthProvider` doesn't handle the statusBar by itself
- `LightspeedExplorerWebviewViewProvider` is now managed by lightspeed/base.ts
- `lightSpeedOAuthProvider`: rename `isAuthenticated()` to `getLocalSession()`
  to reflect the fact it only check for a local session.
- `getLightSpeedStatusBarText()` doesn't take any parameter and generate the status bar text depending on the context.

New exceptions
--------------

- `LightspeedNoLocalSession`: raised when an error is caused by the absance of a
  local LightSpeed `AuthenticationSession` session.
- `LightspeedAccessDenied`: raised when the API /me end-point returns a 401 error.

This allow the simplification of `getApiInstance()`, `getData()`, `getLocalSession()`,
`rhOrgHasSubscription()`.

New attribute: `lightSpeedOAuthProvider.userIsConnected`
--------------------------------------------------------

This attribute is True only if the user has a local session and can reach the `/me`
end-point without getting a 401 (Unauthorized) error.

UI changes
----------

When the user is disconnected, a click on the Lightspeed bar will trigger the
authentication process.
image

### DIFF
--- a/src/features/lightspeed/base.ts
+++ b/src/features/lightspeed/base.ts
@@ -14,6 +14,7 @@ import {
 import {
   LIGHTSPEED_ME_AUTH_URL,
   AnsibleContentUploadTrigger,
+  LightSpeedCommands,
 } from "../../definitions/lightspeed";
 import { ContentMatchesWebview } from "./contentMatchesWebview";
 import {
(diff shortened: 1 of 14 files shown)


src/features/lightspeed/base.ts
```diff
@@ -106,6 +127,50 @@ export class LightSpeedManager {
...
+    try {
+      const data: UserResponse = await this.apiInstance.getData(
+        `${getBaseUri(this.settingsManager)}${LIGHTSPEED_ME_AUTH_URL}`,
+      );
+      console.log(`refreshUserInfo: ${data}`);
+      console.log(`userIsconnected!, lets'refresh`);
+      this.lightSpeedAuthenticationProvider.userIsConnected = true;
+      this.lightspeedExplorerProvider.refreshWebView();
+      this.statusBarProvider.updateLightSpeedStatusbar();
+      // NOTE: This lightspeedConnectReady doesn't seem to be actually used
```
TamiTakamiya, line 145: This `lightspeedConnectReady` context  was added with #1076, which was a timing issue caused because the initialization (`activate`) takes longer than usual in certain situation.  We need to make sure if the issue that was resolved by #1076 does not occur after this change.

PR #1259, opened 2024-04-19
support RHSSO authentication for Lightspeed

robinbobbitt:
Adds support for Red Hat Single Sign On authentication to the Lightspeed service. This is being added specifically to address scenarios where redirect URL would be invalid when using the Lightspeed OAuth provider, such as hosted devspaces. 

This PR attempts to avoid confusing the end user by NOT displaying both "Sign in with Ansible Lightspeed to use Ansible" and "Sign in with Red Hat to use Ansible". Instead, it prefers the Lightspeed auth unless we detect we're in an unsupported scenario, in which case it prefers RHSSO. It also adds commands to explicitly sign in with one or the other to allow for testing, preferences, etc. We can refine this approach over time as this gets use.

The Red Hat Authentication extension v0.2.0 needs to be installed in order to use the Sign in with Red Hat option. The extension is otherwise not required.


src/utils/logger.ts
```diff
@@ -0,0 +1,30 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from "vscode";
+
+export class Log {
+  private output: vscode.LogOutputChannel;
+
+  constructor() {
+    this.output = vscode.window.createOutputChannel("Ansible", { log: true });
```
robinbobbitt, line 12: @ganeshrn I see we already have an "Ansible Support" output channel. Should we use that for everything, or shall I go ahead with an "Ansible" one? Specifically here I'm logging the behavior around the two auth flavors.

TamiTakamiya:
@robinbobbitt I think this is too big for a "minor" label PR...

goneri (approved):

Issue #1225, opened 2024-04-12
Code quality checklist


ssbarnea:
This ticket is used to track progressive changes we need to make to the codebase in order to improve the code quality and prevent regression.

```[tasklist]
### Tasks
- [x] eslint: enforce "@typescript-eslint/no-non-null-assertion"
- [x] "strict": true in ALS tsconfig.js file [see](https://stackoverflow.com/a/58868185/99834) (might split it into multi-step)
- [x] eslint: enforce "no-empty-function"
- [x] eslint: enforce "no-case-declarations"
- [x] eslint: enforce "no-constant-condition"
- [x] eslint: enforce "no-control-regex"
- [x] eslint: enforce "no-prototype-builtins"
- [ ] typescript-eslint: adopt [strict-type-checked](https://typescript-eslint.io/users/configs/#strict-type-checked) (keep last)
```

Blocked ones:

- [ ] adopt ESM (due vscode extensions not supporting them yet)
- [x] adopt chai v5 (requires ESM codebase)

AAP-22902

PR #1262, opened 2024-04-22
Add type to smartFilter

ssbarnea:
Related: #1225


packages/ansible-language-server/test/helper.ts
```diff
@@ -108,7 +115,12 @@ export function smartFilter(completionList, triggerCharacter) {
   });
 
   let filteredCompletionList = triggerCharacter
-    ? searcher.search(triggerCharacter).slice(0, 5)
+    ? searcher
+        .search(triggerCharacter)
+        .slice(0, 5)
+        .map((completion) => {
+          return completion.item;
```
audgirka, line 122: @ssbarnea if it's a single item, we can just return it like
```
.map(completion => completion.item);
```

PR #1269, opened 2024-04-23
Add file icon

GeorchW:
I felt that it's a bit ugly to have no icon for Ansible files.

Before | After
--|--
image | image


ganeshrn (approved):

priyamsahoo:
@GeorchW, thanks for pointing it out. We already have an SVG located inside the `images` dir. Kindly update the PR to make use of that rather than adding a new SVG.

priyamsahoo (dismissed):
@GeorchW, thanks for pointing it out. We already have an SVG located at the `images` dir. Kindly update the PR to use that SVG rather than adding a new PNG.

GeorchW:
> We already have an SVG located at the images dir. Kindly update the PR to use that SVG rather than adding a new PNG.

Yes, I used that SVG file and adjusted the margins and the color slightly to fit with the other file icons in VS Code. The original icon is bigger and brighter and thus looks out-of-place. I had a look at it in both dark and light mode, and the adjusted color works for both. (I just had a look at the default dark and light themes, I didn't check every theme there is.)

GeorchW:
@priyamsahoo ping

goneri:
@GeorchW could you please rebase?

GeorchW:
@goneri done

goneri:
Thank you @GeorchW!

PR #1276, opened 2024-04-24
Make thumbUp/Down icons 'sticky' at bottom on explanation panel

TamiTakamiya:
Make thumbUp/Down icons 'sticky' at the bottom-right corner on playbook explanation panel so that it won't scroll up with the content.


goneri (approved):
Thank you @TamiTakamiya

PR #1277, opened 2024-04-25
Adjust ALS .js.map files for debugging

TamiTakamiya:
When `npm run compile` copies ALS js/d.ts/js.map files from packages/ansible-language-server to out directory, we need to adjust the source path in .js.map files so that we can set breakpoints on TypeScript source files in debugging.

PR #1282, opened 2024-04-26
Webpack fails with "File not under rootDir" error

TamiTakamiya:
This is an attempt to resolve ci build issues with webpack as shown in the following example:

```
  ERROR in /home/runner/work/vscode-ansible/vscode-ansible/src/extension.ts
  ./src/extension.ts 5:30-48
  [tsl] ERROR in /home/runner/work/vscode-ansible/vscode-ansible/src/extension.ts(5,31)
        TS6059: File '/home/runner/work/vscode-ansible/vscode-ansible/src/features/vault.ts' is not under 'rootDir' '/home/runner/work/vscode-ansible/vscode-ansible/packages/ansible-language-server'. 'rootDir' is expected to contain all source files.
    The file is in the program because:
      Imported via "./features/vault" from file '/home/runner/work/vscode-ansible/vscode-ansible/src/extension.ts'
      Root file specified for compilation
  ts-loader-default_02839d45964937a4
```

In my local test, these error occurred in 25% of probability (5 out of 20 executions of `yarn webpack`).

The fix contains following changes:

1. Add a new module rule to `webpack.config.ts`. Before this PR (after #1277), one rule was used for building sources for extension and ALS and unexpected compilations were done. For example, `out/src/client` directory contained `package` subdirectory, which contained outputs of ALS even though the same outputs were found in `/out/src/server` as well. Though it is not 100% sure why the error occurs only in 25% of time, build of extension and ALS should be clearly separated.
2. Removed `yarn als:compile` and `yarn compile` from `yarn webpack` because webpack will compile TypeScript sources we do not need to pre-compile before running webpack.  They generates unnecessary js files because webpack creates bundiled js files for extension.js and server.js.
3. Specify output directory of ALS in `webpack.config.ts` to `out/src/server` of project root. For debugging TypeScript, d.ts files need to be stored there.
4. Update Taskfile.yml to run `yarn test-compile` before running e2e/UI test.  Since we no longer run `yarn compile`, which executed `tsc -p ./` that compiled test files, we need to explicitly build test sources before running tests.

One remaining issue is that the source maps are generated in production build as well. It is possible to create separate webpack configurations for production and development, but it seems that we are using a single configuration for both. We may want to have separate webpack configuraitons in the near future.


TamiTakamiya:
I am going to close this without merging for now.

PR #1284, opened 2024-04-29
eslint: address more exceptions

ssbarnea:
Related: #1225

### DIFF
--- a/packages/ansible-language-server/src/providers/completionProviderUtils.ts
+++ b/packages/ansible-language-server/src/providers/completionProviderUtils.ts
@@ -45,7 +45,7 @@ export function getVarsCompletion(
       ) {
         path = parentKeyPath;
         const scopedNode = path[path.length - 3].toJSON();
-        if (scopedNode.hasOwnProperty("vars")) {
+        if (Object.prototype.hasOwnProperty.call(scopedNode, "vars")) {
           const varsObject = scopedNode["vars"];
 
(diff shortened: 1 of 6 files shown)


packages/ansible-language-server/src/providers/completionProviderUtils.ts
```diff
@@ -45,7 +45,7 @@ export function getVarsCompletion(
       ) {
         path = parentKeyPath;
         const scopedNode = path[path.length - 3].toJSON();
-        if (scopedNode.hasOwnProperty("vars")) {
+        if (Object.prototype.hasOwnProperty.call(scopedNode, "vars")) {
```
audgirka, line 48: @ssbarnea there is already a function in `utils/misc.ts` called `hasOwnProperty`. We can use that here at least for language-server related checks

audgirka: Also is there a way we can share this with extension code?

audgirka (changes requested):

packages/ansible-language-server/src/providers/completionProviderUtils.ts
ssbarnea: It seems that using that function confuses ts compiler and gives errors. If you can manage to make it work in a follow-up it would be happy to use it as it would look less verbose.

PR #1299, opened 2024-05-03
Allow use of ${workspaceFolder} in more commands

garymm:


packages/ansible-language-server/src/utils/execPath.ts
```diff
@@ -11,8 +13,22 @@ import { ExtensionSettings } from "../interfaces/extensionSettings";
...
+  context: WorkspaceFolderContext,
 ): string {
-  return name === "ansible-lint"
-    ? settings.validation.lint.path
-    : path.join(path.dirname(settings.ansible.path), name);
+  let pathFromSettings =
+    name === "ansible-lint"
+      ? settings.validation.lint.path
+      : path.join(path.dirname(settings.ansible.path), name);
+  return replaceWorkspaceFolderInPath(pathFromSettings, context);
```
ssbarnea, line 22: I am sure the vscode api has generic support for replacing all variables in the settings, please use that instead of trying to do it DIY.

ssbarnea (changes requested):

packages/ansible-language-server/src/utils/execPath.ts
garymm: I have been unable to find it.
This suggests it doesn't exist: https://github.com/microsoft/vscode/issues/46471

garymm: However if you can find it I'd be happy to use it!

ssbarnea:
@garymm Thanks for pointing to this upstream issue, see my comment from https://github.com/microsoft/vscode/issues/46471#issuecomment-2463087254 -- if nobody steps in to ensure that is maintained, we might want to add this as package under our extension and reuse the code from there. I do not fancy adding it as an npm dependency considering that the code was not touched in 4 years.

garymm:
@ssbarnea  If you want to do that instead feel free to close this PR and open an issue for that. I still think this PR is on net helpful but I don't maintain this code base so obviously up to you / other maintainers.

audgirka:
@garymm looks like library https://github.com/DominicVonk/vscode-variables has a new release and we can use it.

garymm:
@audgirka that'd be great if you could do that. I'm personally unlikely to do that any time soon.

garymm:
I think this is still an issue but I don't have time to fix it the way the maintainers have requested.

PR #1301, opened 2024-05-06
Telemetry for Playbook Summary API (ThumbsUp/Down)

TamiTakamiya:
"Schema 1" telemetry implementation for summary API call (ThumbsUp/Down).

### DIFF
--- a/src/interfaces/lightspeed.ts
+++ b/src/interfaces/lightspeed.ts
@@ -72,8 +72,8 @@ export interface IssueFeedbackEvent {
 }
 
 export interface PlaybookOutlineEvent {
-  action?: ThumbsUpDownAction;
-  outlineId?: string;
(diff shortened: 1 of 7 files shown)


goneri (approved):

src/interfaces/lightspeed.ts
```diff
@@ -67,11 +71,17 @@ export interface IssueFeedbackEvent {
   description: string;
 }
 
+export interface PlaybookOutlineEvent {
+  action?: ThumbsUpDownAction;
```
goneri, line 75: I cannot find a case where we would want these parameters to be optional. We can just flag them both as mandatory.

TamiTakamiya: Changed these two required parameters in the latest commit

PR #1305, opened 2024-05-06
Playbook Generation UI update

TamiTakamiya:
A few updates on Playbook Generation UI:

On the second page:
  1. display the prompt input on the first page as a static text.
  2. Add the "Edit" link next to the prompt static text. The link work in the same way as the "Back" button.
  3. Enable the "Reset" button only the summary text was changed from the original text generated by the API.

On the first page:
  1. Display examples in gray color.

### DIFF
--- a/src/webview/apps/lightspeed/playbookGeneration/main.ts
+++ b/src/webview/apps/lightspeed/playbookGeneration/main.ts
@@ -61,7 +61,7 @@ window.addEventListener("message", (event) => {
       changeDisplay("promptContainer", "block");
 
       const element = document.getElementById("playbook-text-area") as TextArea;
-      savedSummary = element.value = message.summary.content;
+      savedSummary = element.value = message.summary;
       resetTextAreaHeight();
 
       const prompt = document.getElementById("prompt") as HTMLSpanElement;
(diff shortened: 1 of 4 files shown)


src/webview/apps/lightspeed/playbookGeneration/main.ts
```diff
@@ -57,11 +58,16 @@ window.addEventListener("message", (event) => {
       changeDisplay("firstMessage", "none");
       changeDisplay("secondMessage", "block");
       changeDisplay("generatePlaybookContainer", "block");
+      changeDisplay("promptContainer", "block");
 
       const element = document.getElementById("playbook-text-area") as TextArea;
-      savedSummary = element.value = message.summary;
+      savedSummary = element.value = message.summary.content;
```
goneri, line 64: There is no `.content` key. `message.summary` is a string.

TamiTakamiya: You're right. It was a part of #1301... I will get rid of that. Thanks.

goneri (approved):
Thank you @TamiTakamiya <3.

PR #1309, opened 2024-05-09
Update msg for WCA model not found error and create a new error no org linked

TamiTakamiya:
Update the error message displayed for the error that WCA model was not found.  Old msg was:

`Could not find a Model Id for IBM watsonx Code Assistant. Please contact your administrator.`

and the new message is

`Your organization does not have an IBM watsonx Code Assistant model configured. Contact your Red Hat organization administrator to configure a model, or provide a model id override in your VSCode Ansible extension settings`

With this change, a new error is added for the case that the user is not linked to an organization because the `error__wca_model_id_not_found` error was issued for that case as well.  The error is separated with the new message:

`Your user Id is not linked to an organization Id. Please contact your administrator.`

This requires the change on the server side.  The server side PR will be created separately.


manstis (approved):
LGTM 👍

src/features/lightspeed/errors.ts
```diff
@@ -184,11 +184,18 @@ ERRORS.addError(
     "Could not find an API Key for IBM watsonx Code Assistant. Please contact your administrator.",
   ),
 );
+ERRORS.addError(
+  403,
+  new Error(
+    "error__wca_organization_not_linked",
+    "Your user Id is not linked to an organization Id. Please contact your administrator.",
```
goneri, line 191: This situation should never happen by design, we should instead ask the user to report a bug.

goneri (changes requested):

src/features/lightspeed/errors.ts
TamiTakamiya: @goneri I think end users need to contant their adminitrator to open a ticket.  Does the following message sound reasonable?
```
"Your user Id is not linked to an organization Id. Please contact your administrator to open a ticket with Red Hat support."
```

robinbobbitt: Agree with @goneri 's comment. If this ever happened with our saas Lightspeed service, it's a bug somewhere, and if it happens with on-prem, it's a misconfiguration of the service. How about something like "Ansible Lightspeed is not properly configured to determine your model id. Please contact your administrator." to cover either scenario?

src/features/lightspeed/errors.ts
```diff
@@ -184,11 +184,18 @@ ERRORS.addError(
...
+  ),
+);
 ERRORS.addError(
   403,
   new Error(
     "error__wca_model_id_not_found",
-    "Could not find a Model Id for IBM watsonx Code Assistant. Please contact your administrator.",
+    "Your organization does not have an IBM watsonx Code Assistant model configured. Contact your Red Hat organization administrator to configure a model, or provide a model id override in your VSCode Ansible extension settings",
```
robinbobbitt, line 198: I know this sentence came straight from the issue I wrote, but... (sorry!)
I'm thinking "VSCode" is redundant and maybe we can make the specific setting more clear. How about "or specify a model in your Ansible extension settings under Lightspeed: Model Id Override."?

robinbobbitt (changes requested):

src/features/lightspeed/errors.ts
robinbobbitt: Thinking on it more, I think wisdom-service should be updated to allow a custom model override even when a default model is not configured with ANSIBLE_AI_MODEL_MESH_MODEL_NAME. Then this message could be something like "Ansible Lightspeed does not have a model configured. Contact your Ansible administrator to configure a model, or provide a model id override in your VSCode Ansible extension settings." This is a very on-prem message, with the expectation we'd never hit this particular error with saas Lightspeed. wdyt?

TamiTakamiya: As I wrote in the PR on the service part, if it is possible to have a user without a link to an organization, the exception thrown will be changed. Let me work on the service part first and change the extension later.

src/features/lightspeed/errors.ts
TamiTakamiya: I agree with you.  I will work on the changes on service part first and then change extension.

src/features/lightspeed/errors.ts
```diff
@@ -184,11 +184,18 @@ ERRORS.addError(
     "Could not find an API Key for IBM watsonx Code Assistant. Please contact your administrator.",
   ),
 );
+ERRORS.addError(
+  403,
+  new Error(
+    "error__no_default_model_id",
+    "Ansible Lightspeed does not have a model configured. Contact your Ansible administrator to configure a model, or provide a model id override in your Ansible extension settings.",
```
robinbobbitt, line 191: Do we want to use the same wording here as the other message, `or specify a model in your Ansible extension settings under Lightspeed: Model Id Override.`?

TamiTakamiya: The message was updated in the latest commit.

src/features/lightspeed/errors.ts
TamiTakamiya: The message was updated in the latest commit, which corresponds to the changes on the service side to add WcaNoDefaultModelId exception.

goneri (approved):

manstis (approved):
LGTM 👍

robinbobbitt (approved):

PR #1312, opened 2024-05-10
test-setup.sh: just ignore the podman-docker shim if it's installed

goneri:


ssbarnea (commented):
Any reasoning for this change? The less verbose behavior may make it harder to debug or prevent regressions.

ssbarnea (changes requested):
This does not work as intended because docker shim also defines DOCKER_HOST variable, so execution will still fail.

goneri:
> This does not work as intended because docker shim also defines DOCKER_HOST variable, so execution will still fail.

Hum. It was working fine on Fedora 40 with this change and docker shim installed.

PR #1318, opened 2024-05-15
Add unit tests for content creator utility functions

priyamsahoo:
The PR will add unit tests for the utility functions written for the content creator features in the extension.


audgirka (approved):

package.json
```diff
@@ -956,7 +956,7 @@
     "package": "./tools/package.sh",
     "preinstall": "",
     "pretest": "yarn run compile",
-    "unit-tests": "nyc mocha --recursive test/units",
+    "unit-tests": "nyc mocha --recursive test/units/lightspeed",
```
goneri, line 959: How do you trigger these new tests?

priyamsahoo: These tests are triggered using the `test-e2e` script.
These new tests depend on the `vscode` engine, and thus, it is not possible to trigger them using `mocha`.

goneri: This is actually the reason why we created the tests/units directory originally. The idea was that unit-tests should not import `vscode` at all and just depend on a regular Javascript environment. If your tests are depending on the vscode run-time, maybe  you can move your tests in a different directory?

ssbarnea (approved):

PR #1328, opened 2024-05-24
Playbook Generation content match support

TamiTakamiya:
Playbook Generation content match support

### DIFF
--- a/src/features/lightspeed/playbookGeneration.ts
+++ b/src/features/lightspeed/playbookGeneration.ts
@@ -4,9 +4,10 @@ import { Webview, Uri, WebviewPanel } from "vscode";
 import { getNonce } from "../utils/getNonce";
 import { getUri } from "../utils/getUri";
 import { SettingsManager } from "../../settings";
-import { isLightspeedEnabled } from "../../extension";
+import { isLightspeedEnabled, lightSpeedManager } from "../../extension";
 import { LightspeedUser } from "./lightspeedUser";
 import { GenerationResponse } from "@ansible/ansible-language-server/src/interfaces/lightspeedApi";
+import { LightSpeedCommands } from "../../definitions/lightspeed";
 
 let currentPanel: WebviewPanel | undefined;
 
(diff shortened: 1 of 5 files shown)


src/features/lightspeed/playbookGeneration.ts
```diff
@@ -22,10 +23,23 @@ async function openNewPlaybookEditor(playbook: string) {
...
+
+  await editor.edit(editBuilder, {
     undoStopBefore: true,
     undoStopAfter: false,
   });
+
+  lightSpeedManager.contentMatchesProvider.suggestionDetails = [
+    {
+      suggestionId: generationId,
+      suggestion: playbook,
```
goneri, line 35: I think we should instead loop on all the tasks of the playbook and add an entry in this list. This would:
- allow us to continue to use the regular contentMatch end-point
- and reduce the complexity of the code base  by avoiding the `isPlaybook` parameter, on having just one branch in the code logic.

TamiTakamiya: So I prefer to sending a playbook to WCA.
Because Lightspeed (WCA) gives a playbook as a suggestion, I think it is more natural to set the whole playbook to the `suggestion` parameter to call contentMatch. 

This PR also parses the playbook (in lines 226-239 of contentMatchesWebView.ts) since the response provides one dimensional array of content match object per task.   However, I saw WCA generated a playbook that contained two plays. If we send an array of tasks collected from those two plays, some structural information is lost.


I guess probably the current WCA implementation is reusing the code match code implemetned for single-/multi-task inference suggestion.  So it is very likely that we get the same results even if we send a list of tasks instead of a playbook.

goneri:
@TamiTakamiya could you rebase the PR?

goneri (approved):

src/features/lightspeed/playbookGeneration.ts
TamiTakamiya: For the record: I will the code to send a playbook as the input. If issues arise with it, I will update the code.

PR #1333, opened 2024-05-27
lightspeed/telemetry: trigger even when explanation is opened

goneri:
Also minor refactoring of the interfaces:

- PlaybookExplanationEvent: When the explanation panel is requested
- PlaybookExplanationFeedbackEvent: When the user votes on an explanation
- PlaybookOutlineFeedbackEvent: When the user votes on an outline


src/extension.ts
```diff
@@ -69,8 +69,8 @@ import {
   AuthProviderType,
 } from "./features/lightspeed/lightspeedUser";
 import {
-  PlaybookOutlineEvent,
-  PlaybookExplanationEvent,
+  PlaybookOutlineFeedbackEvent,
+  PlaybookExplanationFeedbackEvent,
```
TamiTakamiya, line 73: Because we are going to the thumbs up/down buttons from page 2 to page 3, I think we'd better call the event as `PlaybookGenerationFeedbackEvent`. wdyt?

TamiTakamiya (commented):
I left one comment. Thanks!

TamiTakamiya (approved):

PR #1334, opened 2024-05-28
Allow more than one official vscode-ansible extension releases

TamiTakamiya:
This is an attempt to allow more than one official VS Code Marketplace releases, which is outlined in the comment lines of `get-marketplace-version.sh`:

```
# In order to allow more than one official releases in one month, we will add
#     100,000,000 * (patch segment in the version string in package.json)
# to the number of second passed between last tag and last commit. This logic
# will generate marketplace version numbers from the version string in
# package.json as shown in the following examples:
#
# Version in package.json  Official/Pre-release  Marketplace version number
# -------------------------------------------------------------------------
# 24.6.0                   Official release      24.6.0
# 24.6.0                   Pre-release           24.6.nnnnnnnn
# 24.6.1                   Official release      24.6.100000000
# 24.6.1                   Pre-release           24.6.1nnnnnnnn
# 24.6.2                   Official release      24.6.200000000
# 24.6.2                   Pre-release           24.6.2nnnnnnnn
#
# where nnnnnnnn represents a number of seconds passed from the last official
# release
```


tools/package.sh
```diff
@@ -6,7 +6,7 @@ rm -f ./*.vsix
 yarn run webpack
 VERSION="$(./tools/get-marketplace-version.sh)"
 vsce_package_args=(--no-dependencies --no-git-tag-version --no-update-package-json --readme-path docs/README.md)
-if [[ "$VERSION" != *.0 ]]; then
+if [[ "$VERSION" != *.0 && "$VERSION" != *00000000 ]]; then
```
goneri, line 9: Could we replace this long random digit with just the date? e.g: `date -u +"%Y%m%dT%H%M%S"`

TamiTakamiya: As long as a new version number is greater than any of existing version number, it will work.  So just using numbers based on the current date/time should work, but my intention here is to 

(1) make distinction between pre-releases and official releases, and 
(2) provide an easy way to co-relate official releases and GitHub releases

i.e. we can easily identify if a version is official by looking last digits of the patch version and checking release notes on the GitHub page is also easy, i.e. look at the release note for GitHub version `24.6.1` for Marketplace version `24.6.100000000`.

TamiTakamiya:
After discussions, we decided to use version numbers in the following way:

-  `YY.(even month number).n` for statble releases. `n` will be a sequential number, `0`, `1`, `2`, ...
- `YY.(odd month number).nnnnnn` for pre-releases.  The logic to determine the patch version `nnnnnnn` will be implemented later.

I am going to close this PR without merging. Thanks.

PR #1338, opened 2024-06-03
AAP-24192: Action field in Recommendation events have a blank/null value

romartin:
Hey @robinbobbitt @TamiTakamiya 

See https://issues.redhat.com/browse/AAP-24192

**Goal**

The goal is to guarantee the correlation between `completion`/`Recommendation Generated` events, along with the `inlineSuggestionFeeadback`/`Recommendation Action` ones, from the VSCode (client) perspective. It means this fix applies to both `schema1` and `schema2` (although "seated" users actually don't send schema1 feedback events ...) 

**Steps to reproduce**

The actual vscode plugin behavior works as:
1- User writes a prompt and press Enter, so a completion request starts
2- Meanwhile the completion request is in progress (so suggestion not yet displayed to the user), if f.i. a WHITESPACE key is pressed, it does not send any feedback event. But completion event has been already sent.
3- But once pressing the BACKSPACE key (before completion req. above is in progress), the cursor is back to the original position, and another completion request is started, so another completion event is sent, which may end with some suggestion displayed, and the user accepting/rejecting/igrnoing the suggestion (feedback event)
4- Instead of step 3, the user, while the completion request above is in progress, can also press another key again, so the iteration starts again

**Solution**

Assuming once a completion is in progress, and the user press some other key that may cause completion to abort, at least send a IGNORED feedback event for each. This solution may result in an increase of the number of feedback events, but at least guarantees that once a copmletion req. (so event as well) is in progress, it send an IGNORED event once the user does something that "cancels" current completion

I understand I am changing the actual "desired" behavior , as some tests where not expecting that, so feel free to let me know your concerns or feedback please, or if I miss something. :)

Thanks!

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -41,8 +41,8 @@ import {
   inlineSuggestionTextDocumentChangeHandler,
   inlineSuggestionTriggerHandler,
   LightSpeedInlineSuggestionProvider,
-  rejectPendingSuggestion,
   setDocumentChanged,
+  processInProgessSuggestion,
 } from "./features/lightspeed/inlineSuggestions";
 import { playbookExplanation } from "./features/lightspeed/playbookExplanation";
 import { ContentMatchesWebview } from "./features/lightspeed/contentMatchesWebview";
@@ -260,7 +260,7 @@ export async function activate(context: ExtensionContext): Promise<void> {
   // Listen for text selection changes
   context.subscriptions.push(
     vscode.window.onDidChangeTextEditorSelection(async () => {
-      rejectPendingSuggestion();
+      processInProgessSuggestion();
     }),
   );
 
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -42,7 +42,7 @@ import {
   inlineSuggestionTriggerHandler,
   LightSpeedInlineSuggestionProvider,
   setDocumentChanged,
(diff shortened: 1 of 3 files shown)


TamiTakamiya:
@romartin With #1129, we changed the code not to send feedback if the completion API returns a non-200 status. May I understand you are changing that with this PR?

romartin:
Good catch @TamiTakamiya , 

This implementation is not really caring the completion response error codes, it just ensuring that if some completion event was send, and some other key change happens, which "ignores" the completion request, it send the corresponding feedback event. So maybe this brokes https://github.com/ansible/vscode-ansible/pull/1129,  not sure, let me double check how it works with the non-200 repsonse error codes....  

Will let you know something ASAP! Thanks!

romartin:
Hey @TamiTakamiya , you were right, I fixed one issue but introduced another, once the response is not an HTTP 200. I am looking for a proper fix, and trying not to break anything else, anyway hope tests will help.

BTW Anything else I should take into account during this fix? (A part from checking response codes for sending feedback)

romartin:
Closing. No loner applies. Working on a new PR. Sorry for any inconveniences.

romartin:
Replaced by https://github.com/ansible/vscode-ansible/pull/1381

PR #1344, opened 2024-06-04
Playbook Generation wizardId support

TamiTakamiya:
Support wizardId for Playbook Generation feature to provide a unique identifier that belong to a specific Playbook Generation UI session.  

This also changed the behavior of generationId, which was re-used for multiple the generations API calls from a single Playbook Generation UI session, not to be re-used, i.e. always a unique ID will be provided for each generations API call.

### DIFF
--- a/test/ui-test/lightspeedUiTest.ts
+++ b/test/ui-test/lightspeedUiTest.ts
@@ -590,7 +590,7 @@ export function lightspeedUIAssetsTest(): void {
       const notification = notifications[0];
       expect(await notification.getMessage()).equals(
         "Enable lightspeed services from settings to use the feature.",
-      ).to.be.true;
+      );
     });
   });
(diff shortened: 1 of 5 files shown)


test/ui-test/lightspeedUiTest.ts
```diff
@@ -590,7 +590,7 @@ export function lightspeedUIAssetsTest(): void {
       const notification = notifications[0];
       expect(await notification.getMessage()).equals(
         "Enable lightspeed services from settings to use the feature.",
-      ).to.be.true;
```
TamiTakamiya, line 593: This is not directly related to the wizardId support, but was found when running the UI tests for verifying this PR.  I am not sure how it was successfully completed before...

goneri (approved):

PR #1345, opened 2024-06-04
Capture user operations on Playbook Generation UI

TamiTakamiya:
Related: https://issues.redhat.com/browse/AAP-24906

### DIFF
--- a/src/definitions/lightspeed.ts
+++ b/src/definitions/lightspeed.ts
@@ -24,6 +24,12 @@ export enum ThumbsUpDownAction {
   DOWN = 1, //Thumbs Down
 }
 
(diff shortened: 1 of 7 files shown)


ssbarnea (changes requested):
Do not merge this until @ganeshrn approves because it appears to work against reducing amount of analytics data. @ganeshrn remove my -1 if you think is ok.

See https://issues.redhat.com/browse/AAP-24906

ganeshrn (approved):
The analytics for lightspeed is send from the lightspeed service unlike the other features of extension that directly send the events to segement. The change in the PR looks good to me.

PR #1347, opened 2024-06-07
Playbook Explanation Thumbs button visibility issue

TamiTakamiya:
Improve the visibility of Thumbs Up/Down buttons on the playbook explanation panel when the explanation text overwraps the button.

**Without this PR:**
image
image

**With this PR:**
image
image

### DIFF
--- a/media/playbookGeneration/style.css
+++ b/media/playbookGeneration/style.css
@@ -165,10 +165,12 @@ a:active {
 }
 
 .feedbackContainer {
+  background-color: var(--vscode-editor-background, #ffffff80);
   position: absolute;
   right: 0px;
   bottom: 0px;
   display: flex;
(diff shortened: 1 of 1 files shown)


manstis (approved):
I'd preferred to have seen them move down so as not to overlap the text at all (I don't know if `padding-top` or `margin-top` would help or whether the icons end up getting clipped).

But, as it stands, this does improve their visibility so approving 👍

TamiTakamiya:
> I'd preferred to have seen them move down so as not to overlap the text at all (I don't know if `padding-top` or `margin-top` would help or whether the icons end up getting clipped).

Yes, that's my preference as well. However, I wanted to make the changes minimum for this release.

> But, as it stands, this does improve their visibility so approving 👍

Thanks!

PR #1348, opened 2024-06-10
Enable sending only ansibleMetadata event

ganeshrn:
*  Enable sending only ansibleMetadata event if telemetry is enabled
*  Add python version in ansibleMetadata event payload
*  Store the ansibleMetadata event globally to send the event only once across user seessions if any of the parameters within the payload is changed
*  Additionally send the event only when there is change in the extension settings or a new Asible file is opened with different ansibleMetadata payload.

Related: https://issues.redhat.com/browse/AAP-24906


src/utils/telemetryUtils.ts
```diff
@@ -134,10 +158,11 @@ export class TelemetryErrorHandler implements ErrorHandler {
   }
 
   error(error: Error, message: Message, count: number): ErrorHandlerResult {
-    this.telemetry.send({
-      name: "ansible.lsp.error",
-      properties: { jsonrpc: message.jsonrpc, error: error.message },
+    sendTelemetry(this.telemetry, true, "ansible.lsp.error", {
```
TamiTakamiya, line 161: The second parameter is set to `true`. Does it mean that this telemetry event can be sent even if the telemetry service is not initialized?

src/utils/telemetryUtils.ts
```diff
@@ -214,9 +239,8 @@ export class TelemetryOutputChannel implements vscode.OutputChannel {
       }
       this.errors.push(value);
       const timeoutHandle = setTimeout(() => {
-        this.telemetry.send({
-          name: "ansible.server.error",
-          properties: { error: this.createErrorMessage() },
+        sendTelemetry(this.telemetry, true, "ansible.server.error", {
```
TamiTakamiya, line 242: Same as on line 161.

TamiTakamiya (commented):
I left two comments (both on the same topic). Thanks.

src/utils/telemetryUtils.ts
ganeshrn: Have added `sendTelemetry` function to provide a single to call the telemetry send API. In the earlier code TelemetryErrorHandler call we are invoking the send method directly without checking for initiliased however there is a check added to ensure telemetryService object is initialised.

TamiTakamiya: Thanks for the explanation!

TamiTakamiya (approved):

ssbarnea (approved):
LGTM, assuming that coverage does not go down.

PR #1359, opened 2024-06-12
Always enable the remove content suggestion header

TamiTakamiya:
We found the content suggestion header
```
# Content suggestion provided by Ansible Lightspeed
```
sometimes causes unexpected results when other extensions read them insered in playbooks.

For dealing with those issues, we are going to remove them when an inline suggestion is inserted into a playbook.  We already have the option to control that behavior `ansible.lightspeed.disableContentSuggestionHeader`, whose default was `false` (=disabled).  We are going to remove the header always and remove the option.


src/features/lightspeed/inlineSuggestions.ts
```diff
@@ -819,37 +819,33 @@ export async function inlineSuggestionReplaceMarker(position: vscode.Position) {
...
+  const line = position.line;
+
+  const text = editor.document.getText(
+    new vscode.Range(
+      new vscode.Position(line, 0),
+      new vscode.Position(line + 1, 0),
+    ),
+  );
+  // Remove the prepended text
+  const commentPosition = text.indexOf(
```
goneri, line 832: I don't understand what is this `commentPosition` variable for.

TamiTakamiya: Though these are shown as new codes, they are actually just re-formatted ones (encluded in an if-block, which was removed with this PR).

I understand the variable is used just to check if the pre-defined comment exists or not.

goneri (approved):

PR #1362, opened 2024-06-12
lightspeed: remove an unused block

goneri:
The block was not actually never triggered.


TamiTakamiya:
@robinbobbitt would you take a look at this?  It seems to me the text always contains the suggestion header and those lines are not needed, but I might be missing some possibilities. Thanks.

robinbobbitt (approved):
Honestly I can't recall if I put those lines there out of paranoia or because I'd discovered some weird edge case while testing this. Testing today I'm not able to trigger this code in any way that would cause commentPosition to be -1.

ssbarnea (approved):

PR #1366, opened 2024-06-14
Gracefully handle playbook generation and explanation not available

manstis:
This PR moves the _generic_ handling of error `code`'s from the backend service in to the Language Server.

Exceptions thrown by the Language Server are handled by JSON-RPC and converted into messages following a standard format.

We, however, want to preserve the `code` from the backend service and show _meaningful_ messages to users.

The response from the Language Server is now always considered to be _successful_ to avoid JSON-RPC handling _exceptions_ but the payload represents either success or failure state. This is comparable to how shell scripting, C etc and Go handle exceptions: The function calls always _succeed_ however the result may indicate success (`rc=0`) or failure (`rc>0`) states.


manstis:
Hi @ssbarnea @TamiTakamiya @goneri 

I wonder if you can advise... when running
```
task build
task package
task test
```
The [changes](https://github.com/ansible/vscode-ansible/pull/1366/files#diff-25542c177e5427ebb65cd56c18fea2c82916ad16217afa28cc85ea7331afce21) I've made moving a utility function from the extension to LSP are not found and the extension shows an error message:
> Activating extension 'redhat.ansible' failed: Cannot find module '@ansible/ansible-language-server/src/utils/handleApiError'
Require stack:
- /home/manstis/workspaces/git/manstis/forks/vscode-ansible/out/client/src/features/lightspeed/api.js
- /home/manstis/workspaces/git/manstis/forks/vscode-ansible/out/client/src/features/lightspeed/base.js
- /home/manstis/workspaces/git/manstis/forks/vscode-ansible/out/client/src/extension.js
- /home/manstis/workspaces/git/manstis/forks/vscode-ansible/out/test-resources/VSCode-linux-x64/resources/app/out/vs/loader.js
- /home/manstis/workspaces/git/manstis/forks/vscode-ansible/out/test-resources/VSCode-linux-x64/resources/app/out/bootstrap-amd.js
- /home/manstis/workspaces/git/manstis/forks/vscode-ansible/out/test-resources/VSCode-linux-x64/resources/app/out/bootstrap-fork.js.

However if I install the `.vsix` built everything runs fine.

Is there some magic I need to perform (especially for this PR?)

TamiTakamiya:
I will run test on my machine.  I guess the failure occurred in "E2E" test, i.e. even though

```
yarn als-compile
yarn compile
yarn webpack-dev
yarn package
yarn coverage-ui-with-mock-lightspeed-server
```
passes,
```
yarn als-compile
yarn compile
yarn webpack-dev
yarn package
yarn coverage-all
```
fails.  Let me try it on my machine.

TamiTakamiya:
Yes, `yarn coverage-all` failed after `yarn package`

```
   :
Error: Cannot find module '@ansible/ansible-language-server/src/utils/handleApiError'
Require stack:
- /home/ttakamiy/git/ansible/vscode-ansible/out/client/src/features/lightspeed/api.js
- /home/ttakamiy/git/ansible/vscode-ansible/out/client/src/features/lightspeed/base.js
- /home/ttakamiy/git/ansible/vscode-ansible/out/client/src/extension.js
  :
```

ansible-language-server (ALS) runs in a separate process (previously it was in a separate repository).  When we package it in a .vsix file, we stores ALS files under `out/server` and configure the server uses the files in that directory.  So originally it was not designed to share files in ALS in VS Code extension.

When we package .vsix file, we run webpack and it resolves dependencies and creates bundled js files.  On client (VS Code extension) part, it is provided as `extension.js` file.  I believe that bundled `extension.js` contains the `handleApiError.js`. So it works with .vsix.

To resolve this isssue, I think we can create a js file, which is webpacked even when E2E test is running like what I did for "SyntaxHighlighter", which is used for Playbook Generation. I will describe it later (after standup).

TamiTakamiya:
> To resolve this isssue, I think we can create a js file, which is webpacked even when E2E test is running like what I did for "SyntaxHighlighter", which is used for Playbook Generation. I will describe it later (after standup).

In `webpack.config.ts`, I added a new entry:

```
  syntaxHighlighter: "./src/features/utils/syntaxHighlighter.ts",
```
We define output location as
```
"[name]/src/[name].js"
```
so `yarn webpack` or `yarn webpack-dev` will generate the bundled js as `out/syntaxHighlighter/src/syntaxHighlighter.js`

`src/features/lightspeed/playbookGeneration.ts` has the following ugly code to load `syntaxHighlighter.js`:
```
        // eslint-disable-next-line @typescript-eslint/no-explicit-any
        let syntaxHighlighter: any;
        try {
          syntaxHighlighter =
            await require(/* webpackIgnore: true */ "../../syntaxHighlighter/src/syntaxHighlighter");
        } catch (error) {
          syntaxHighlighter =
            await require(/* webpackIgnore: true */ "../../../../syntaxHighlighter/src/syntaxHighlighter");
        }
```
In the production env, the dynamic loading occurs from (bundled) `src/extension.js`. That's why I need to have two dynamic import calls...

So if we do

1. Add an entry to webpack.config.ts
2. Statically link the entry to handleApiError.js
3. Dynamically link the webpacked js where it is needed

it should work... but it may be too much for this...

For this case, simply creating a symlink might work.  Let me try that approach.

TamiTakamiya:
@manstis Still trying to find a better solution... For running E2E test, we are using "testRunner", which is descrbed [here](https://code.visualstudio.com/api/working-with-extensions/testing-extension).  If the testRunner could resolve the dependencies, we do not have to change the runtime code because .vsix runs w/o issues. I think there should be a way to specify additional path and checking the doc.

src/features/lightspeed/playbookExplanation.ts
```diff
@@ -43,21 +44,27 @@ export const playbookExplanation = async (
...
+      content,
+      explanationId,
+      client,
+      lightspeedAuthenticatedUser,
+      settingsManager,
+    ).then((response: ExplanationResponse) => {
+      if (isError(response)) {
+        vscode.window.showErrorMessage(response.message ?? UNKNOWN_ERROR);
+        currentPanel.setContent(
+          `<p><span class="codicon codicon-error"></span>Cannot load the explanation: <code>${response.code}</code></p>`,
```
goneri, line 60: Since is a message that is visible by the user. Something like: `The operation has failed:` would be better.

packages/ansible-language-server/src/errors.ts
```diff
@@ -246,6 +246,13 @@ ERRORS.addError(
     },
   ),
 );
+ERRORS.addError(
+  404,
+  new Error(
+    "feature_not_available",
+    "The requested action is not yet supported in your environment.",
```
goneri, line 253: `yet` makes it sounds like it's a matter of time. It's not necessarily the case. I would go with a more neutral: `The requested action is not available in your environment.`

TamiTakamiya:
@mantis  [THIS](https://github.com/ansible/vscode-ansible/commit/711a99945b9d526111cde2529d160c00ea41bf7f) is my atttempt to fix this with a webpacked js. Since it's too ugly, I will spend some more time on searching better solutions after fixing the Playbook Explanation issues with task files/role-based playbooks.

TamiTakamiya:
@manstis I have simplified [my test fix](https://github.com/ansible/vscode-ansible/commit/83410d5793101820bea66d823b36f9c6af5a4704). Would you take a look?

src/features/lightspeed/api.ts
```diff
@@ -20,12 +19,28 @@ import {
...
 import { ANSIBLE_LIGHTSPEED_API_TIMEOUT } from "../../definitions/constants";
 import { UserAction } from "../../definitions/lightspeed";
-import { mapError } from "./handleApiError";
+import { IError } from "@ansible/ansible-language-server/src/interfaces/lightspeedApi";
 import { lightSpeedManager } from "../../extension";
 import { LightspeedUser } from "./lightspeedUser";
 
 const UNKNOWN_ERROR: string = "An unknown error occurred.";
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
manstis, line 28: Thanks @TamiTakamiya :love_you_gesture:

goneri (approved):

goneri:
@manstis thank you Michael!

PR #1370, opened 2024-06-14
Last lines on Playbook Explanation are hard to read

TamiTakamiya:
This is the continuation of #1347.

After #1347, the visibility of ThumbsUp/Down icons were improved, but the readability of last lines of playbook explanation has been a problem. This PR inserts a 30-px height div element at the bottom of the explanation and resolve the overlapping issue when the window is scrolled down to the bottom.

image

image

### DIFF
--- a/media/playbookGeneration/style.css
+++ b/media/playbookGeneration/style.css
@@ -159,6 +159,10 @@ a:active {
   display: flow;
 }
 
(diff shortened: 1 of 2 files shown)


manstis (approved):
LGTM 👍

goneri (approved):

PR #1379, opened 2024-06-19
Playbook explanation for task files and playbooks w/o tasks

TamiTakamiya:
Modifications of the playbook explanation feature for 
1. Task files, and
2. Playbooks that do not contain tasks, for example, playbooks that have roles only

For 1. We do not enable playbook explanation. The "Explain the current playbook" button on the sidebar will be grayed out.

For 2. We will display an informational message: `Explaining a playbook with no tasks in the playbook is not supported.`

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -296,6 +296,7 @@ export async function activate(context: ExtensionContext): Promise<void> {
         lightSpeedManager,
         pythonInterpreterManager,
       );
+      lightSpeedManager.lightspeedExplorerProvider.refreshWebView();
       metaData.sendAnsibleMetadataTelemetry();
     }),
   );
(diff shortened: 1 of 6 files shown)


src/extension.ts
```diff
@@ -296,6 +296,7 @@ export async function activate(context: ExtensionContext): Promise<void> {
         lightSpeedManager,
         pythonInterpreterManager,
       );
+      await lightSpeedManager.lightspeedExplorerProvider.refreshWebView();
```
TamiTakamiya, line 299: This was needed for the case that a playbook was not initially recognized as an Ansible file and the user manually set the Ansible language code.  The webview needs to be refreshed at that timing.

goneri: Do we actually need to `await` here?

TamiTakamiya: Maybe not.  Let me try w/o await.

goneri (approved):
I did the rebase manually and everything works fine. Nice job @TamiTakamiya <3.

src/extension.ts
TamiTakamiya: It worked without `await`. I'll get rid of that.

goneri (approved):

goneri:
Tests are failing because of `TypeError: Cannot read properties of undefined (reading 'openEditor')` which was removed.

PR #1381, opened 2024-06-19
Action field in Recommendation events have a blank/null value

romartin:
See https://issues.redhat.com/browse/AAP-24192

**Description**

The goal is to provide some quick fix some more common use cases, where action feedback is being handled in a wrong way, without a need for a bigger refactoring.

This PR improves the lifecycle of completion / feedback events , by assuming that **several completion requests** can be sent, **before the suggestion is displayed to the user**, this way:
- Guarantees that for each successful (`HTTP 200`) completion request, it will end up on an action feedback event, can be accepted, rejected or ignored
- Guarantees that for failed (non-200) completion requests, no action feedback is sent at all

**Tested use cases**
- Accepting/rejecting a suggestion both by clicking on the buttons (`accept`, `acceptKey`) also by using keys (`Tab`, `Esc`)
- Ignored action feedback sent, before suggestion presented, but user pressed some key (eg: whitespace)
- Ignored action feedback sent, once suggestion presented, but user moved to another window, or lost the focus
- Rejected action feedback sent, once suggestion presented, but user pressed some key (eg: whitespace)
- Tested he fix for the steps to reproduce described in https://issues.redhat.com/browse/AAP-24192

**TODO: Update/add unit tests**

### DIFF
--- a/src/features/lightspeed/api.ts
+++ b/src/features/lightspeed/api.ts
@@ -150,22 +150,10 @@ export class LightSpeedAPI {
       vscode.window.showErrorMessage(mappedError.message ?? UNKNOWN_ERROR);
       return {} as CompletionResponseParams;
     } finally {
-      if (
-        isCompletionSuccess &&
-        !this.completeSuggestionFeedback(suggestionId)
-      ) {
+      if (isCompletionSuccess && !this.cancelSuggestionFeedback(suggestionId)) {
         await inlineSuggestionHideHandler(UserAction.IGNORED, suggestionId);
       }
     }
-  }
-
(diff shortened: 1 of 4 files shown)


TamiTakamiya (approved):
LGTM. I like the idea to store multiple pending feedbacks in an array. Thanks for the efforts.

romartin:
Thanks @TamiTakamiya !! All existing tests are passing, so sounds good, also I am working on increasing the coverage now , once done I will update the PR.

BTW You had a chance to test it?

Thanks!

romartin:
Increased test coverage!! also rebased :+1:

PR #1386, opened 2024-06-20
lightspeed: report 502 error (e.g uwsgi Harakiri)

goneri:


robinbobbitt (commented):
@goneri I'm not sure how this changes any behavior since without these lines it would pick up the same error on line 49?

goneri:
Yes, you're right this is not the right fix.

PR #1395, opened 2024-06-21
Lightspeed exp/gen: trigger a timeout after 28s

goneri:
Give up to 28s for the operation to succeed and cancel the call after that.
28s is below our hard limit of 30s that we have server side.


TamiTakamiya (approved):

TamiTakamiya:
@ssbarnea We do not need to include this in today's (Jun 21) release. I've checked with @goneri.

PR #1398, opened 2024-06-21
package.json: remove the compile-withserver target

goneri:
We don't need to build ansible-language-server from a different directory anymore.


TamiTakamiya (approved):

audgirka (dismissed):
@goneri this change is good but it is being used in multiple places. Please update the same

PR #1405, opened 2024-06-24
Lightspeed Authentication UI Tests (Linux only)

TamiTakamiya:
Implement Lightspeed Authentication UI Tests based on [the idea posted here.](https://github.com/redhat-developer/vscode-extension-tester/issues/518#issuecomment-1265205344)

This PR is for Linux only.  I will address Mac support later.

### DIFF
--- a/tools/run-ui-test-with-mock-lightspeed-server.sh
+++ b/tools/run-ui-test-with-mock-lightspeed-server.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# For invoking VS Code executable to open "vscode:" URL
+export UI_TEST=1
+
 if [ "$1" == "--coverage" ]; then
   TEST_SCRIPT=coverage-ui-current
 else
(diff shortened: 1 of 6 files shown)


tools/run-ui-test-with-mock-lightspeed-server.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# For invoking VS Code executable to open "vscode:" URL
+export UI_TEST=1
```
manstis, line 4: Are you confident `1` equates to [`true`](https://github.com/ansible/vscode-ansible/pull/1405/files#diff-ff27543dbc8fdf6d075f959784b5efa93483e056a7dd470da27ba87d23b45bf6R8) in TypeScript?

I know this is a common Python pattern... but since TypeScript is type-safe using a `boolean` may be better?

manstis (approved):
I only skimmed the tests; but their scope seems reasonable.

tools/run-ui-test-with-mock-lightspeed-server.sh
TamiTakamiya: This is a bash shell script and I think using `1` as `true` is a convention in shell programming.

Besides the convention, using an environment variable here itself may not be a good programming practice... Probably I should implement a command option for mock lightspeed server. I will consider that when I implement MacOS support. Thanks.

PR #1415, opened 2024-07-05
Adopt community-ansible-dev-tools image

priyamsahoo:
The PR adopts the new community-ansible-dev-tools (adt) image:
1. changes the default EE to be adt from creator-ee.
2. updates the test fixtures to test with the adt image.
3. updates the CI files to test the extension using adt in the CI.


tools/test-setup.sh
```diff
@@ -442,9 +443,12 @@ if [[ "${PODMAN_VERSION}" != 'null' ]] && [[ "${SKIP_PODMAN:-}" != '1' ]]; then
     log notice "Retrieving ansible-lint version from ee"
     EE_ANSIBLE_LINT_VERSION=$(get_version \
         podman run --rm "${IMAGE}" ansible-lint --nocolor --version)
+    log notice "ansible: ${EE_ANSIBLE_VERSION}, ansible-lint: ${EE_ANSIBLE_LINT_VERSION}"
     log notice "Test podman ability to mount current folder with write access, default mount options"
-    podman run --rm -v "$PWD:$PWD" ghcr.io/ansible/creator-ee:latest \
-        bash -c "[ -w $PWD ] && echo 'Mounts working' || { echo 'Mounts not working. You might need to either disable or make selinux permissive.'; exit 1; }"
+    podman run -v "$PWD:$PWD" \
```
cidrblock, line 448: Try setting user as root here, that might get past the filesystem issue

tools/test-setup.sh
```diff
@@ -379,10 +379,11 @@ if [[ "${DOCKER_VERSION}" != 'null' ]] && [[ "${SKIP_DOCKER:-}" != '1' ]]; then
...
         docker run --rm "${IMAGE}" ansible --version)
     EE_ANSIBLE_LINT_VERSION=$(get_version \
-        docker run --rm "${IMAGE}" ansible-lint --nocolor --version)
-    # Test podman ability to mount current folder with write access, default mount options
-    docker run --rm -v "$PWD:$PWD" ghcr.io/ansible/creator-ee:latest \
-        bash -c "[ -w $PWD ] && echo 'Mounts working' || { echo 'Mounts not working. You might need to either disable or make selinux permissive.'; exit 1; }"
+        docker run "${IMAGE}" ansible-lint --nocolor --version)
+    log notice "ansible: ${EE_ANSIBLE_VERSION}, ansible-lint: ${EE_ANSIBLE_LINT_VERSION}"
+    # Test docker ability to mount current folder with write access, default mount options
+    docker run -v "$PWD:$PWD" ghcr.io/ansible/community-ansible-dev-tools:latest \
```
cidrblock, line 385: although this could work, I think long term we'd want a writable mount.  The image should default to the "podman" user but that could be checked.  It might help to ls -l the dir here to see the perms

cidrblock: interesting, the file for docker are owned by 1001, wehich tells be the runner user is not 1000.  Maybe try a user-podman for docker.

(which is odd thinking about it now https://github.com/ansible/ansible-dev-tools/issues/334)

audgirka: keeping this as is per discussion on weekly call

tools/test-setup.sh
audgirka: keeping this as is per disscussion

cidrblock (approved):

PR #1421, opened 2024-07-09
[DRAFT] AAP-26574 - edit user status view based on changes to /me endpoint

SiddharthRajaraman:
**Jira Ticket:** https://issues.redhat.com/browse/AAP-26574

### DIFF
--- a/src/features/lightspeed/explorerWebviewViewProvider.ts
+++ b/src/features/lightspeed/explorerWebviewViewProvider.ts
@@ -71,7 +71,18 @@ export class LightspeedExplorerWebviewViewProvider
   private async _getWebviewContent(webview: Webview, extensionUri: Uri) {
     const userDetails =
       await this.lightspeedAuthenticatedUser.getLightspeedUserDetails(false);
-    if (userDetails) {
+    let markdownUserDetails =
+      await this.lightspeedAuthenticatedUser.getMarkdownLightspeedUserDetails(
+        false,
+      );
+    markdownUserDetails = String(markdownUserDetails);
+    console.log("==========markdownuserdetails==========");
+    console.log(markdownUserDetails);
+    console.log("=======================================");
+    let content = "";
+    if (markdownUserDetails !== "") {
+      content = markdownUserDetails;
+    } else if (userDetails) {
       const sessionInfo = getLoggedInUserDetails(userDetails);
       const userName = userDetails.displayNameWithUserType;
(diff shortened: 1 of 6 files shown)


src/features/lightspeed/explorerWebviewViewProvider.ts
```diff
@@ -71,25 +71,40 @@ export class LightspeedExplorerWebviewViewProvider
...
           : "";
-      return getWebviewContentWithActiveSession(
-        webview,
-        extensionUri,
-        userName,
-        userType,
-        userRole,
-        this.hasPlaybookOpened(),
-      );
+      content = `
```
goneri, line 93: It would be better to handle the fallback scenario inside `getMarkdownLightspeedUserDetails()`.

src/features/lightspeed/lightspeedUser.ts
```diff
@@ -135,7 +138,44 @@ export class LightspeedUser {
       }
     }
   }
+  /* Get the user info from server */
+  public async getMarkdownUserInfo(token: string) {
```
goneri, line 142: Can you return a HTML string instead? This way you can rename the function `getUserdetailsSnippet()`.

src/features/lightspeed/lightspeedUser.ts
```diff
@@ -135,7 +138,44 @@ export class LightspeedUser {
...
+        `${getBaseUri(this._settingsManager)}${LIGHTSPEED_MARKDOWN_ME_AUTH_URL}`,
+        {
+          headers: {
+            Authorization: `Bearer ${token}`,
+          },
+        },
+      );
 
+      const markdownData = marked.parse(data.content) as string;
+      return markdownData;
```
goneri, line 158: I believe this is actually already a HTML string :-).

src/features/lightspeed/lightspeedUser.ts
```diff
@@ -299,6 +348,25 @@ export class LightspeedUser {
 
     return this._userDetails;
   }
+  public async getMarkdownLightspeedUserDetails(
+    createIfNone: boolean,
+    useProviderType: AuthProviderType | undefined = undefined,
+  ) {
+    // Ensure we don't try to get a lightspeed auth session when the provider is not initialized
+    if (!this._settingsManager.settings.lightSpeedService.enabled) {
+      return undefined;
```
goneri, line 357: To be fair, if we reach this case, it means we are already in serious trouble because half of the extension is not properly initialized.

PR #1425, opened 2024-07-12
AAP 26574 - expose markdown content

SiddharthRajaraman:


src/features/lightspeed/lightspeedUser.ts
```diff
@@ -300,6 +355,61 @@ export class LightspeedUser {
...
+        const userRole =
+          sessionInfo.userInfo?.role !== undefined
+            ? sessionInfo.userInfo?.role
+            : "";
+        content = `
+          <p><strong>Logged in as</strong>: ${userName}</p>
+          <p><strong>User Type</strong>: ${userType}</p>
+          ${userRole ? "Role: " + userRole : ""}
+        `;
+      } else {
```
goneri, line 405: I would initialize `content` as `content = "undefined";` and drop this else block.

src/features/lightspeed/lightspeedUser.ts
```diff
@@ -136,6 +140,47 @@ export class LightspeedUser {
...
+        `${getBaseUri(this._settingsManager)}${LIGHTSPEED_MARKDOWN_ME_AUTH_URL}`,
+        {
+          headers: {
+            Authorization: `Bearer ${token}`,
+          },
+        },
+      );
+
+      const markdownData = marked.parse(data.content) as string;
+      console.log("markdownData");
```
TamiTakamiya, line 159: Are these two lines are for debug only?

SiddharthRajaraman: yep... forgot to delete 🤦.

TamiTakamiya:
In `lightspeedUser.ts`, there are two places where `this._userDetails` is reset to `undefined`. I think `this._markdownUserDetails` also needs to be reset to `undefined` there.

SiddharthRajaraman:
> In `lightspeedUser.ts`, there are two places where `this._userDetails` is reset to `undefined`. I think `this._markdownUserDetails` also needs to be reset to `undefined` there.

Would that be lines 326 and 332?

TamiTakamiya (approved):
LGTM. Thanks!

PR #1426, opened 2024-07-13
Support permission_denied__can_apply_for_trial error code

TamiTakamiya:


src/features/lightspeed/utils/oneClickTrial.ts
```diff
@@ -3,23 +3,43 @@ import { IError } from "@ansible/ansible-language-server/src/interfaces/lightspe
...
-    );
-    if (selection === buttonLabel) {
-      vscode.commands.executeCommand(
-        LightSpeedCommands.LIGHTSPEED_OPEN_TRIAL_PAGE,
+export class OneClickTrialProvider {
+  public mapError(mappedError: IError): IError {
+    // Don't show "Apply for Trial" message if experimental features are not enabled
+    if (
+      mappedError?.code === "permission_denied__can_apply_for_trial" &&
+      !lightSpeedManager.lightspeedExplorerProvider
```
goneri, line 11: We don't need the experimental check.

goneri (approved):

src/features/lightspeed/utils/oneClickTrial.ts
goneri: Ok. We can keep it for now, but since the feature should be enabled server side first and we want the clients to be ready as quickly as possible.

BTW, we must also avoid pushing a release of this without having a fully functional `permission_denied__can_apply_for_trial` support behind.

PR #1432, opened 2024-07-18
Handle HAP detected in playbook gen and explanation

manstis:


packages/ansible-language-server/src/errors.ts
```diff
@@ -129,6 +129,13 @@ ERRORS.addError(
     "Cloudflare rejected the request. Please contact your administrator.",
   ),
 );
+ERRORS.addError(
+  400,
+  new Error(
+    "error__wca_hap_filter_rejection",
+    "IBM watsonx Code Assistant Hate, Abuse, and Profanity filter rejected your request. Please try again.",
```
TamiTakamiya, line 136: If the input is unchanged, re-submitting the playbook generation/explanation request will be rejected again...  I guess we need to tell the user something like "Please eliminate any doubtful phrases and try again".

manstis: @TamiTakamiya Agreed.. I had thought that too. I'll update.

packages/ansible-language-server/src/errors.ts
```diff
@@ -129,6 +129,13 @@ ERRORS.addError(
     "Cloudflare rejected the request. Please contact your administrator.",
   ),
 );
+ERRORS.addError(
+  400,
+  new Error(
+    "error__wca_hap_filter_rejection",
+    "IBM watsonx Code Assistant Hate, Abuse, and Profanity filter rejected your request. Please re-phrase and try again.",
```
robinbobbitt, line 136: Given that the response from WCA doesn't spell out HAP, I'm wondering if we should be presenting this like an official entity with all caps etc. The message from WCA is:
`Our filters detected a potential problem with entities in your input. Please check your input and try again.`
What do we think about something like "Potentially harmful language was detected in your request. Please check your input and try again."

robinbobbitt (changes requested):

packages/ansible-language-server/src/errors.ts
manstis: @robinbobbitt Updated.. Happy to go with your recommendation.

robinbobbitt (approved):
LGTM. Thanks @manstis !

PR #1436, opened 2024-07-19
Message for no access error is displayed at every other attempt for code completion

manstis:
Ensure the "[feedback] in progress" flag is reset in all scenarios; success and error.

### DIFF
--- a/src/features/lightspeed/api.ts
+++ b/src/features/lightspeed/api.ts
@@ -160,7 +160,8 @@ export class LightSpeedAPI {
       }
       return {} as CompletionResponseParams;
     } finally {
-      if (isCompletionSuccess && !this.cancelSuggestionFeedback(suggestionId)) {
+      const cancelled = this.cancelSuggestionFeedback(suggestionId);
+      if (isCompletionSuccess && !cancelled) {
         await inlineSuggestionHideHandler(UserAction.IGNORED, suggestionId);
       }
     }
--- a/test/testScripts/lightspeed/testLightSpeedFunctions.test.ts
+++ b/test/testScripts/lightspeed/testLightSpeedFunctions.test.ts
@@ -204,6 +204,7 @@ function testFeedbackCompletionAPI(): void {
         "SuggestionId assertion successful",
       );
 
+      assert.equal(cancelSuggestionFeedbackSpy.called, true);
       assert.equal(inlineSuggestionHideHandlerSpy.called, false);
     });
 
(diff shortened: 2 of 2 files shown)


src/features/lightspeed/api.ts
```diff
@@ -160,7 +160,12 @@ export class LightSpeedAPI {
       }
       return {} as CompletionResponseParams;
     } finally {
-      if (isCompletionSuccess && !this.cancelSuggestionFeedback(suggestionId)) {
```
TamiTakamiya, line 163: I see... Thank you for finding this out!!

TamiTakamiya (approved):
LGTM. Thanks!

hasys:
Looks like tests do not have 

```
assert.equal(cancelSuggestionFeedbackSpy.called, false);
```

scenario only `true` cases. Is it intended?

manstis:
@hasys 

> Looks like tests do not have
> 
> ```
> assert.equal(cancelSuggestionFeedbackSpy.called, false);
> ```
> 
> scenario only `true` cases. Is it intended?

Yes, see https://github.com/ansible/vscode-ansible/blob/71915d564deaf8420eb8a052f09f34a22dc38905/src/features/lightspeed/api.ts#L163

We *always* ensure the in-flight/in-progress feedback is reset.

We [used to](https://github.com/ansible/vscode-ansible/blob/main/src/features/lightspeed/api.ts#L163) only reset it if the "completion" call was successful; which led to the problem @TamiTakamiya reported.

hasys:
> @hasys
> 
> > Looks like tests do not have
> > ```
> > assert.equal(cancelSuggestionFeedbackSpy.called, false);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > scenario only `true` cases. Is it intended?
> 
> Yes, see
> 
> https://github.com/ansible/vscode-ansible/blob/71915d564deaf8420eb8a052f09f34a22dc38905/src/features/lightspeed/api.ts#L163
> 
> We _always_ ensure the in-flight/in-progress feedback is reset.
> 
> We [used to](https://github.com/ansible/vscode-ansible/blob/main/src/features/lightspeed/api.ts#L163) only reset it if the "completion" call was successful; which led to the problem @TamiTakamiya reported.

If it's always true, should it be removed entirely instead of ensuring it is always true?

manstis:
@hasys 

> If it's always true, should it be removed entirely instead of ensuring it is always true?

The problem @TamiTakamiya found was caused by it *not* being called for both (success and failed) scenarios.

The tests are now asserting that the method was called. Isn't this the point of the test, to check something happened?

PR #1437, opened 2024-07-19
log response data to console when /me request fails

robinbobbitt:
We have had several users report problems connecting to Lightspeed through a proxy. Debugging the response body for requests to our /me endpoint when they fail help us quickly pinpoint if the request is failing when it hits our service or if it is being rejected by the proxy.


manstis (approved):
LGTM 👍

PR #1441, opened 2024-07-22
Add ansible-playbook argument setting option

### LINKED ISSUE #1042: Support command arguments to ansible-playbook
### Problem

Apologies if my workflow was never intended. But I am NOT using an execution environment (so I don't check this in the settings). I AM trying to setup a bunch of local machines (not cloud, not containers).  I need `sudo` with ssh.  Usually I would run something like `ansible-playbook -kK`.  However because the arguments are not exposed to the config, I can't add these flags and running `ansible-playbook` against `playbook.yaml` I get `"msg": "Missing sudo password"`. 

Again maybe my workflow isn't supported and this extension is only meant for (paying?) users against cloud resources and containers.  It's hard to tell because there's no real great starting guide. There's a blog post, a youtube video, and a Readme.md that doesn't even give any information on how to even run `ansible-playbook`, so I assume the target audience is seasoned DevOps not those of us that want to automate our own setups.

### Solution

Expose arguments to `ansible-playbook` in the config (similar to `ansible-lint`)

### Alternatives

I'll have to manually run from terminal or vscode terminal.

### Additional context

_No response_

alisonlhart:
Adds an extension setting under `ansible.playbook.arguments` to add arguments to `ansible-playbook` runs.

Demo:

`ansible-playbook` run,  without arguments set:

Without setting

How to set the argument in settings, and then running `ansible-playbook`:

With setting

Additional changes:

- Made a minor bugfix in `src/features/runner.ts`. The `SettingsManager` class now replaces `ExtensionSettings`. 
- Refactored the `getFilePath()` helper function (now `getFixturePath()` )to be more generic , and adjusted function calls as needed. 
- Added tests for `ansible-playbook` and `ansible-navigator` commands.

Closes #1042.

### DIFF
--- a/.config/requirements.in
+++ b/.config/requirements.in
@@ -1,6 +1,7 @@
 cython>=3.0.5; python_version >= "3.12"
 ansible-core>=2.13.0
 ansible-lint>=6.8.7
+ansible-navigator
 mypy # used by vscode
 pip>=22.3.1
 pip-tools>=6.12.1
--- a/docs/README.md
+++ b/docs/README.md
@@ -231,6 +231,8 @@ any level (User, Remote, Workspace and/or Folder).
 - `ansible.lightspeed.modelIdOverride`: Model ID to override your organization's
   default model. This setting is only applicable to commercial users with an
   Ansible Lightspeed seat assignment.
+- `ansible.playbook.arguments`: Specify additional arguments to append to
+  ansible-playbook invocation. e.g. `--syntax-check`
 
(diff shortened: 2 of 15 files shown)


priyamsahoo (changes requested):
Everything looks good. Just one last thing is required. Documentation :)

alisonlhart:
@priyamsahoo fixed! Added the setting to the list in README.md.

priyamsahoo (approved):

PR #1449, opened 2024-07-23
Remove one of two examples for Lightspeed completion tests

TamiTakamiya:
Right now each of Lightspeed completions API tests is executed twice for `taskName: "Print hello world"` and `taskName: "Create a file foo.txt"`.  When we used a real Lightspeed instance, running the test twice using different samples were somehow meaningful, but now we are using a mock server and running the test twice just increases the possibility of the failure on timing-dependent test cases.  This PR remove one of the two examples to improve the stability of completions API tests.


hasys:
Probably, it should be removed instead of commenting out. If needed the commit can be reverted in the future.

hasys (approved):
Looks good, thank you @TamiTakamiya! 👍

PR #1450, opened 2024-07-24
Refresh icon for Lightspeed Explorer view

TamiTakamiya:
Screenshot from 2024-07-24 08-58-55

### DIFF
--- a/src/features/lightspeed/lightspeedUser.ts
+++ b/src/features/lightspeed/lightspeedUser.ts
@@ -404,8 +404,8 @@ export class LightspeedUser {
       await this.getMarkdownLightspeedUserDetails(false);
     const userDetails = await this.getLightspeedUserDetails(false);
 
-    let content = "undefined";
-    if (markdownUserDetails !== "") {
+    let content;
+    if (markdownUserDetails) {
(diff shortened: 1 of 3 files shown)


src/features/lightspeed/lightspeedUser.ts
```diff
@@ -332,6 +286,67 @@ export class LightspeedUser {
...
+    createIfNone: boolean,
+    session: vscode.AuthenticationSession,
+  ): Promise<boolean> {
+    try {
+      const userinfo: LoggedInUserInfo = await this.getUserInfo(
+        session.accessToken,
+      );
+      this._session = session;
+
+      let markdownUserInfo: string = "";
```
goneri, line 305: I believe we can avoid the try inside an existing try block by moving the block above.

goneri (approved):

src/features/lightspeed/lightspeedUser.ts
TamiTakamiya: I also do not like nested try block, but calling getUserInfo before calling getUserInfoFromMarkdown seems more natural than the reversed order... Let me merge this version and consider refactoring later.

PR #1477, opened 2024-07-31
Limiting vscode-extension-tester dep to ~8.3.0

alisonlhart:
See npm/node-semver docs: https://github.com/npm/node-semver?tab=readme-ov-file#tilde-ranges-123-12-1.


alisonlhart:
@ssbarnea With this versioning, it allows us to move to vscode-extension-tester 8.3.1. Would it be better to just temporarily have a hard req of `8.3.0`?

ssbarnea (approved):

PR #1487, opened 2024-08-02
Remove redundant activation events

ssbarnea:


audgirka (approved):

ssbarnea:
@audgirka Apparently this is a sign that these activation events are needed. I am bit confused because vscode reports them as no longer needed, as being implicit:



See https://stackoverflow.com/questions/68870703/why-should-i-declare-the-command-again-in-activationevents-filed-of-package-js

Any ideas?

Issue #1481, opened 2024-08-01
test: console output is polluted with output from lightspeed server


ssbarnea:
Our console logs when runnings tests is polluted with output from others tools and services that run in parallel, making it harder to debug. We need to address these by ensuring that they end-up in specific log files that are collected.

## output pollution with: test unit

Please note that this pollution happens with passing tests! If this would be visible on screen on failed tests, it would have not being a problem.

Only mocha output should be visible for passing tests.

```
✔ suggestion has no matches
{ contentmatches: [ { contentmatch: [Array] } ] }
YAMLParseError: Unexpected scalar at node end at line 2, column 1:

- name: foo
I
^

    at Composer.onError (/home/runner/work/vscode-ansible/vscode-ansible/node_modules/yaml/dist/compose/composer.js:69:34)
    at Object.resolveEnd (/home/runner/work/vscode-ansible/vscode-ansible/node_modules/yaml/dist/compose/resolve-end.js:31:21)
    at Object.composeDoc (/home/runner/work/vscode-ansible/vscode-ansible/node_modules/yaml/dist/compose/compose-doc.js:37:27)
...    ✔ suggest has invalid YAML
```

## lightspeed server pollution with: test ui
```
{"level":"info","message":"{\n  \"playbookGenerationAction\": {\n    \"wizardId\": \"d406325e-dcf8-4741-afe5-0181e6571209\",\n    \"action\": 0,\n    \"toPage\": 1\n  },\n  \"metadata\": {\n    \"ansibleExtensionVersion\": \"0.0.0\"\n  }\n}"}
```

## pollution with: test e2e

Spamming messages such below must go in specific log file, not test console:
```
Ansible found in the workspace
[file-inspection] language set by file inspection
[inline-suggestions] ...
[ansible-lightspeed] ...
```

## Remediation

```[tasklist]
### Tasks
- [ ] test ui: have clean mocha output, other services should log to file
- [ ] test e2e: have clean mocha output, other services should log to file
- [ ] test unit: have clean mocha output, other services should log to file
- [x] test als: no action needed, output seems clean
```

PR #1490, opened 2024-08-02
Additional logging when proxy block requests to lightspeed

robinbobbitt:
Follow up to https://github.com/ansible/vscode-ansible/pull/1437 for token exchange blocked by proxy. Debugging the response body for requests to Lightspeed when they fail help us quickly pinpoint if the request is failing when it hits our service or if it is being rejected by the proxy.

### DIFF
--- a/src/features/lightspeed/lightSpeedOAuthProvider.ts
+++ b/src/features/lightspeed/lightSpeedOAuthProvider.ts
@@ -400,6 +400,10 @@ export class LightSpeedAuthenticationProvider
           "[ansible-lightspeed-oauth] error message: ",
           error.message,
         );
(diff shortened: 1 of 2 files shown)


ssbarnea (changes requested):
This change works against https://github.com/ansible/vscode-ansible/issues/1481 -- we need to use a different logging mechanism than console log for the service, specialized file logger?

robinbobbitt:
@ssbarnea I made the istanbul edits you recommend. Ok if we move forward with this and address the logging recommendations as part of https://github.com/ansible/vscode-ansible/issues/1481?

ssbarnea (approved):

PR #1492, opened 2024-08-02
lightspeed: ensure the LSP can access the http_proxy envvars

goneri:
Ensure the LSP has access to the root process envvars, this way Axios can honor the
local proxy configuration.


ssbarnea:
Is this still needed? If so it will need a rebase. I made it draft, make it ready for review when is passing CI, or close it if no longer needed.

PR #1498, opened 2024-08-05
lightspeed: rely on electron's fetch

goneri:
- `fetch()` in the main process works well with the proxy
- this is not the case from within the LSP server

TODO:

- Move away from the LSP server
- Error handling

### DIFF
--- a/packages/ansible-language-server/package.json
+++ b/packages/ansible-language-server/package.json
@@ -47,7 +47,6 @@
   "dependencies": {
     "@flatten-js/interval-tree": "^1.1.3",
     "antsibull-docs": "^1.0.2",
-    "axios": "^1.7.2",
     "glob": "^10.4.5",
     "ini": "^4.1.3",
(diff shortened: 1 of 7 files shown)


TamiTakamiya:
I also verified that the `loadFetch()` worked well on the extension side (in `base.ts`), but the `loadFetch()` on LSP (in `ansibleLanguageService.ts`) failed.  `require("electron")` throws an error (`Cannot find module 'electron'`)

goneri:
> I also verified that the `loadFetch()` worked well on the extension side (in `base.ts`), but the `loadFetch()` on LSP (in `ansibleLanguageService.ts`) failed. `require("electron")` throws an error (`Cannot find module 'electron'`)

Yes, I also faced the issue. This is the main reason why I decided to move away from LSP. It was going to make our life miserable with the HTTP proxy when NTLM/Kerberos is in the picture.

PR #1499, opened 2024-08-06
[DRAFT] AAP-27334 | removing 'explain current playbook' when trial UI page appears

SiddharthRajaraman:


goneri (changes requested):
I'm not a super fan of this change because:
- once the users will have started the Trial, they will be forced to refresh the panel to see the "Explain current playbook" button and this is counter intuitive.
- the match on the string is likely to be a source of pain in the future if we need to change it, especially if we have different versions of the server in the wide (e.g OnPrem).

I feel like it's better to stay like we are today, if the users don't want to see the button, they can still turn Lightspeed off.

I believe a better design would imply the user of a polling to refresh the user status and only display the button when the status say that the user configuration is operational.

PR #1500, opened 2024-08-06
AAP-28376: lightspeed: move playbook exp/gen away from the LSP server

goneri:
It's tricky to get the LSP to send the requests through a HTTP server and we don't
want to deal with to different network configuration/stack.

In addition, we would like to migrate away from Axios and use Electron's `fetch` which
is already configured to use the VS Code proxy configuration.


TamiTakamiya (approved):
LGTM. It was good that we could remove the tricky logic to get `mapError()` in `api.ts` :-)

Issue #1140, opened 2023-12-14
hardcoded python3 is used as executable instead of ansible.python.interpreterPath


zalent:
### Summary

There is **_ansible.python.interpreterPath_** setting but it's not getting used. There are multiple hardcoded python3 executables launches which should not be used when **_ansible.python.interpreterPath_** is set.
Few examples:
https://github.com/ansible/ansible-language-server/blob/75a39bb16eb3b8767d7d719d61fb13b2c86f6215/src/services/ansibleConfig.ts#L85
https://github.com/ansible/ansible-language-server/blob/75a39bb16eb3b8767d7d719d61fb13b2c86f6215/src/utils/getAnsibleMetaData.ts#L114
https://github.com/ansible/ansible-language-server/blob/75a39bb16eb3b8767d7d719d61fb13b2c86f6215/src/utils/getAnsibleMetaData.ts#L128
https://github.com/ansible/ansible-language-server/blob/75a39bb16eb3b8767d7d719d61fb13b2c86f6215/src/utils/commandRunner.ts#L37

### Extension version

2.9.118

### VS Code version

1.86.0

### Ansible Version

```shell
❯ ansible --version
ansible [core 2.16.0]
  config file = None
  configured module search path = ['/home/ans/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/ans/.local/lib/python3.11/site-packages/ansible
  ansible collection location = /home/ans/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/ans/.local/bin/ansible
  python version = 3.11.5 (main, Sep  7 2023, 00:00:00) [GCC 11.4.1 20230605 (Red Hat 11.4.1-2)] (/usr/bin/python3.11)
  jinja version = 3.1.2
  libyaml = True
```

### OS / Environment

Win 11 WSL2 Centos 9 Stream Python 3.11, default system python3 points to python3.9

### Relevant log output

```shell
Ansible meta data
ansible information:

core version: 2.16.0

location: ~/.local/lib/python3.11/site-packages/ansible

config file path: /mnt/c/Users/ans/AppData/Local/Programs/Microsoft VS Code Insiders/None

collections location:

~/.ansible/collections
/usr/share/ansible/collections
/usr/lib64/python39.zip
/usr/lib64/python3.9
/usr/lib64/python3.9/lib-dynload
~/.local/lib/python3.9/site-packages
/usr/lib64/python3.9/site-packages
[...]

PR #1502, opened 2024-08-08
Use `ansible.python.interpreterPath` as executable instead of hardcoded `python3`

shatakshiiii:
Fixes: #1140


packages/ansible-language-server/src/utils/getAnsibleMetaData.ts
```diff
@@ -138,8 +138,12 @@ async function getAnsibleInfo() {
 async function getPythonInfo() {
   const pythonInfo: ansibleMetaDataEntryType = {};
 
+  const settings = await context.documentSettings.get(
+    context.workspaceFolder.uri,
+  );
+  const interpreterPathFromConfig = settings.python.interpreterPath;
   const pythonVersionResult = await getResultsThroughCommandRunner(
-    "python3",
+    `${interpreterPathFromConfig}/python3`,
```
ganeshrn, line 146: Is `python3` required here?

audgirka:
@ganeshrn @alisonlhart we don't need this PR as it is already been taken care of [here](https://github.com/ansible/vscode-ansible/blob/main/packages/ansible-language-server/src/utils/commandRunner.ts#L36)

Since the we just get the python path from config we need to send the executable name that is the reason why we are just sending `python3`

PR #1503, opened 2024-08-10
Unmount volumes when execution environment is disabled

audgirka:
This PR removes volume mounts when we disable execution environments.
Also, the volumes will be read-only when mounted


cidrblock (approved):
TY

ganeshrn (approved):

PR #1504, opened 2024-08-10
Update collection path for some als tests

audgirka:
This PR updates the incorrect collection path 
(from `/usr/share/ansible` to `/usr/share/ansible/collections`) 
which is set before running some als tests.


shatakshiiii (approved):

cidrblock (approved):
TY

PR #1508, opened 2024-08-13
Delete ALS EE cache before and after E2E and ALS tests

### LINKED ISSUE #1473: test: als tests are not executed on linux and they fail on GHA
Due to an unfortunate mistake in build scripts (taskfile) we did not run als tests under linux for a very long time. Luckily for us we did run them on windows runner, so we have some limited coverage (non ee ones only).

See https://github.com/ansible/vscode-ansible/actions/runs/10179594927/job/28155656130?pr=1446

We need to address these failures ASAP, other changes have less priority, especially features.

Hint: I have the impression that switching the order of the 4 tests would make the als test pass if they run before the unit or e2e tests but causing one of them to fails. If confirmed it basically means that there are some leftovers that affect their executions. Keep in mind that I was not able to reproduce this locally because locally I already have executed the tests at some point. I tried removing the `out` and `node_modules` and the behavior is still the same, so these leftovers are likely to reside at different level, like user home directory or container engine state (missing image or extra present).

alisonlhart:
This resolves an issue where EE volume mounts were incorrect after running EE tests in either e2e or ALS test suites.

Fixes #1473.

### DIFF
--- a/.config/dictionary.txt
+++ b/.config/dictionary.txt
@@ -35,6 +35,7 @@ WSLENV
 Webviews
 YOLO
 aarch
+afterall
 ajinkyau
 akira
 alefragnani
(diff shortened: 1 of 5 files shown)


cidrblock (approved):
awesome, ty for sticking with it to the end

PR #1509, opened 2024-08-14
Require ALS tests

cidrblock:
Require ALS test run moving forward since #1508


cidrblock:
I think this was the intent of #1508 , merging

PR #1517, opened 2024-08-21
Disable Enhanced Context in VS Code

TamiTakamiya:
Per [AAP-28766](https://issues.redhat.com/browse/AAP-28766)


omaciel (approved):
/lgtm

Side note: I wish we had test/coverage as we're removing code without a clear way to see if we broke anything

PR #1518, opened 2024-08-22
Make playbook generation work even with an empty outline

TamiTakamiya:
This is an attempt to make the Playbook Generation UI work even when WCA return an empty outline for the first API call made from the first page of the Playbook Generation UI.


goneri:
Outdated.

PR #1548, opened 2024-09-25
Clean up test-resources & ext dirs before running UI test

TamiTakamiya:
AAP-31519

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -156,9 +156,6 @@ tasks:
   test-ui:
     desc: Run UI tests {{.XVFB}}
     aliases: [ui]
-    env:
-      MOCHA_GREP: Test One Click Trial feature
-      MOCHA_INVERT: true
     cmds:
       - yarn webpack-dev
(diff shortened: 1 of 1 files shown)


ssbarnea (approved):

ssbarnea (approved):

Taskfile.yml
```diff
@@ -156,12 +156,10 @@ tasks:
   test-ui:
     desc: Run UI tests {{.XVFB}}
     aliases: [ui]
-    env:
-      MOCHA_GREP: Test One Click Trial feature
-      MOCHA_INVERT: true
     cmds:
       - yarn webpack-dev
       - "{{.VIRTUAL_ENV}}/bin/python3 --version"
+      - bash -c 'rm -rf ./out/test-resources ./out/ext'
```
ssbarnea, line 162: there is no need to run bash here, it would work without too.

TamiTakamiya: Thanks for the information!

PR #1554, opened 2024-09-27
AAP-32136: VS Code Extension: Playbook custom_prompt titivation

manstis:
Change settings description and add validation.


TamiTakamiya (approved):
LGTM. Thanks.

btw, I've read that [multiline settings are supported](https://code.visualstudio.com/updates/v1_59#_multiline-string-settings) and thought it would be good for those options, but it did not work as far as I tried...

TamiTakamiya:
https://github.com/microsoft/vscode/issues/183025#issuecomment-1557347569

It looks like the `multilineText` support was mentioned in a release note, but it was not actually implemented....

manstis:
@TamiTakamiya 

> It looks like the `multilineText` support was mentioned in a release note

It does work.. but it does not render a `TextArea`.. you can however keep typing and the widget height grows (unlike single line text settings).. I've made the change on this PR. I gleaned this from looking at the source code for it!

PR #1558, opened 2024-10-02
AAP-32429: VS Code Extension: Remove custom_prompt fields from Settings page

manstis:
Remove "Custom Prompt" related settings.

PR #1559, opened 2024-10-02
AAP-32429: VS Code Extension: Remove custom_prompt fields from Settings page

TamiTakamiya:
Same as #1558, but from an ansible/vscode-ansible branch.


manstis (approved):
LGTM :+1: (but it would.... as I'm approving my own PR).

PR #1568, opened 2024-10-07
AAP-32403: Lightspeed: Playbook generation: Handle "warnings" in response

manstis:
VSCode changes to support display of "warnings" resulting from Playbook Generation requests.

### DIFF
--- a/src/features/lightspeed/playbookGeneration.ts
+++ b/src/features/lightspeed/playbookGeneration.ts
@@ -172,6 +172,7 @@ export async function showPlaybookGenerationPage(
               outline: {
                 playbook: message.playbook,
                 outline: message.outline,
+                warnings: message.warnings,
                 generationId: message.generationId,
               },
             });
@@ -281,6 +282,11 @@ export function getWebviewContent(webview: Webview, extensionUri: Uri) {
     "playbookGeneration",
     "style.css",
   ]);
+  const playbookGenerationUri = getUri(webview, extensionUri, [
+    "media",
+    "playbookGeneration",
+    "playbookGeneration.css",
+  ]);
   const codiconsUri = getUri(webview, extensionUri, [
     "media",
     "codicons",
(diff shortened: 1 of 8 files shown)


manstis:
SonarCloud complaints are about duplicate test code.. 

Personally, the duplication makes individual tests easier to read.

src/features/lightspeed/playbookGeneration.ts
```diff
@@ -332,6 +339,23 @@ export function getWebviewContent(webview: Webview, extensionUri: Uri) {
                 <li></li>
               </ol>
             </div>
+            <div class="warningsContainer">
```
TamiTakamiya, line 342: This implementation looks very nice. Thanks for the efforts on this.  

Followng is just an informative comment:

So far th errors found in Playbook Generation are shown using VS Code's showErrorMessage API ([example](https://github.com/ansible/vscode-ansible/blob/main/src/features/lightspeed/playbookGeneration.ts#L180-L182)), which opens a small pop-up window at the right-bottom of the workbench. I believe it is a standard way and nothing wrong, but showing an error/warning message in the Playbook Generation UI would be more intuitive (btw, Playbook Explanation also displays error messages in its own window). I understand this PR won't be merged into the main branch for now, but when it it is merged to main, we want to re-consider the way to display error messages occured in Playbook Generation.

test/mockLightspeedServer/generations.ts
```diff
@@ -118,9 +119,17 @@ export function generations(
     outline += `\n6. model=${model}`;
   }
 
+  let warnings: GenerationResponseWarning[] = [];
+  if (text === "Create a playbook with warnings") {
```
TamiTakamiya, line 123: :smile:

TamiTakamiya (approved):
LGTM. I left a comment. It's just informative and no action is required for that. Thanks!

src/features/lightspeed/playbookGeneration.ts
manstis: Ah yes indeed @TamiTakamiya 

The "warnings" however are not errors...

The WCA call returns with HTTP200, a `playbook` and _sometimes_ an `outline` and _sometimes_ a `warnings` list.

It is these `warnings` I am propagating to the UI.. Playbook generation can proceed but may give unexpect results.

TamiTakamiya: jfyi. VS Code API has [showWarningMessage](https://code.visualstudio.com/api/references/vscode-api#window.showWarningMessage).  Though I am not a big fan of those APIs, it's an easy route...

PR #1581, opened 2024-10-11
Invoke contentmatch immediately after user accepted a suggestion

TamiTakamiya:
**NOTE: This PR is to be merged into the "custom_propmpts" branch, not into the main branch.**

A customer reported an issue that content match is not invoked occasionally after a task suggestion is accepted.
Their log shows a REJECTED user response is sent after a task suggestion is accepted.  The REJECTED user response can be sent by a text change, which was not caused by accepting a suggestion. Usually, when a text suggestion is accepted, the extension receives the text change caused by the acceptance, but there seem some cases that cause this error.

This PR will immediately after user accepted a suggestion within the `inlineSuggestionCommitHandler`, which is called when a user hit the TAB key.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -231,6 +231,10 @@ any level (User, Remote, Workspace and/or Folder).
 - `ansible.lightspeed.modelIdOverride`: Model ID to override your organization's
   default model. This setting is only applicable to commercial users with an
   Ansible Lightspeed seat assignment.
+- `ansible.lightspeed.includeModelIdOverrideInPlaybookGenExpRequests`:
+  (experimental) When set to `true`, the overridden model ID is included in
+  playbook generation and explanation requests in addition to inline suggestion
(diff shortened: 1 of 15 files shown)


TamiTakamiya:
@manstis 
> It doesn't relate to the work for custom_prompts and appears to be bug that would exist in main.
Maybe so. My thought was that these codes were developed for supporting the Copilot co-existence scenario and I want to spend some more time before merging this change into main.  At least I want to find a more reliable problem recreation procedure and add a unit test for this.

manstis (approved):
LGTM :+1:

manstis:
@TamiTakamiya Was you planning on actually merging this PR!?!?

I'd like to rebase `custom_prompts` with `main` at some point and it'd be nice to "be up to date" first.

TamiTakamiya:
@manstis I will close this and open another PR that is for merging into main.

PR #1585, opened 2024-10-14
Invoke contentmatch immediately after user accepted a suggestion

TamiTakamiya:
This PR contains the same change as the one included in PR #1581.

I was thinking that I would add a test case before merging this change to the main branch.  However, the test case would be similar to the one that #1458 temporary disabled, which was causing instabilities in CI until it was disabled.  I re-thought adding such a test was not a good idea.

A .vsix file that contained PR #1581 fix was sent to a customer and they found it worked. With that manual verification with the customer, I think it is safe to merge this fix to the main branch.


TamiTakamiya:
The E2E test case failing in the one written for Copilot co-exsistence scenario. This means that the change that this PR is about to make may break the scenario. We need to examine this more carefully before merging.  I will close this for now.

PR #1599, opened 2024-10-22
lightspeed auth: forceNewSession by default

goneri:


src/features/lightspeed/lightspeedUser.ts
```diff
@@ -246,7 +246,7 @@ export class LightspeedUser {
       session = await vscode.authentication.getSession(
         useProviderType,
         this.getScopesForAuthProviderType(useProviderType),
-        { createIfNone },
+        { forceNewSession: true },
```
manstis, line 249: I'm not sure it's so simple.

The wizard sends events to and from the view, from and to the controller. If you have multiple tabs open I suspect they'll all receive events unless you begin to filter by/include something like `wizardId`.. I could be wrong.

goneri: It's a draft/hack, I was trying to understand a problem @anshulbehl is facing.

anshulbehl:
Looks like a bug, we should re-open this issue if we think this will resolve the problem I am facing.

goneri:
I think your local cache is corrupted for some reason. I'm not sure if we can recover from this state and if this is a consequence of some problem with the extension itself.

PR #1613, opened 2024-10-31
Rework EE volume mount setting so default value will populate

### LINKED ISSUE #1524: Mounting local volume is not working when using Execution Environment
### Summary

Using a local volume by providing its name like documented in  https://github.com/ansible/vscode-ansible/tree/main#configuration is not possible.

This code https://github.com/ansible/vscode-ansible/blob/main/packages/ansible-language-server/src/services/executionEnvironment.ts#L224-L229 only checks if path exists in filesystem.

Would it be possible to improve this code ? Maybe by checking first with `volume inspect ${mountPath}`.

### Extension version

24.8.4

### VS Code version

1.92.2

### Ansible Version

```shell
$ docker run --rm -it ansible-ee:2.16-dev ansible --version
ansible [core 2.16.10]
  config file = None
  configured module search path = ['/home/ansible/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.12/site-packages/ansible
  ansible collection location = /home/ansible/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.12.3 (main, Aug 23 2024, 06:10:48) [GCC 13.2.1 20240309] (/usr/bin/python3)
  jinja version = 3.1.4
  libyaml = True
```

### OS / Environment

Ubuntu 24.04 using a custom Execution Environment image (`ansible-ee:2.16-dev`)

Using these workspace settings :

```json
"ansible.executionEnvironment.enabled": true,
"ansible.executionEnvironment.containerEngine": "docker",
"ansible.executionEnvironment.image": "ansible-ee:2.16-dev",
"ansible.executionEnvironment.pull.policy": "tag",
"ansible.executionEnvironment.containerOptions": "--network host",
"ansible.executionEnvironment.volumeMounts": [
    {
        "src": "ansible-roles",
        "dest": "/home/ansible/.ansible/roles"
    }
],
```

### Relevant log output

```shell
[Error - 4:28:54 PM] Volume mount source path 'ansible-roles' does not exist. Ignoring this volume mount entry.
```

alisonlhart:
This PR reworks the EE volume mount setting so a default value of `[]` for the volume mounts array is assigned correctly. 

Reason for the change:

Our EE volume mount settings didn't set default values of `""` to the `src`, `dest`, and `options` fields due to apparent lack of vscode detection for settings default values at that nested depth. 

This means the `src`, `dest`, and `options` values were set to their dictionary object values, for example:

```
          src: {
            default: "",
            description:
              "The name of the local volume or path to be mounted within execution environment.",
          },
```
The type would be `object`, and podman would either throw an error or the error would be output into the log (even if the setting was not used at all). 

This PR also removes custom error handling so that podman can throw errors more accurately and display them through toast notifications rather than the log. 

There is also a change to update the `ANSIBLE_ADJACENT_COLLECTIONS__PATH` value in ALS tests, since this path was being skipped previously in tests because the path didn't exist. 

This also updates the docs generation script for ALS so empty arrays as default settings values aren't skipped. 

Fixes #1524.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -202,7 +202,7 @@ any level (User, Remote, Workspace and/or Folder).
   the image and setting tag will always pull if the image tag is 'latest',
   otherwise pull if not locally available.
 - `ansible.executionEnvironment.volumeMounts`: The setting contains volume mount
-  information for each entry in the list. Individual entry consist of a
+  information for each dict entry in the list. Individual entry consists of
   - `src`: The name of the local volume or path to be mounted within execution
     environment.
   - `dest`: The path where the file or directory are mounted in the container.
--- a/docs/als/settings.md
+++ b/docs/als/settings.md
@@ -49,6 +49,11 @@ Specify any additional parameters that should be added to the pull command when 
 . Default value:
 ``""``
 
+## [`ansible.executionEnvironment.volumeMounts`](#executionEnvironment.volumeMounts) { #executionEnvironment.volumeMounts data-toc-label=executionEnvironment.volumeMounts }
+Add a dictionary entry to the array with the volume mount source path (key: &#x27;src&#x27;), destination (key: &#x27;dest&#x27;), and options (key: &#x27;options&#x27;)
+. Default value:
+``""``
+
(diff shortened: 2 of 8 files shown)


cidrblock (approved):
nice digging

PR #1635, opened 2024-11-04
updates for eslint strict type check

audgirka:
This PR adds updates for eslint strict type check complaints on the `lint` CI


audgirka:
@alisonlhart ^^

PR #1641, opened 2024-11-05
AAP-34738: Lightspeed Identify the collections available in the current workspace

goneri:
With this change, the extension builds a list of the local collections of the Workspace
and uses it for the Role Generation wizard.

### DIFF
--- a/src/features/lightspeed/roleGeneration.ts
+++ b/src/features/lightspeed/roleGeneration.ts
@@ -1,8 +1,9 @@
 import * as vscode from "vscode";
 import { v4 as uuidv4 } from "uuid";
-import { Webview, Uri, WebviewPanel } from "vscode";
+import { Webview, Uri, WebviewPanel, workspace } from "vscode";
 import { getNonce } from "../utils/getNonce";
 import { getUri } from "../utils/getUri";
+import { CollectionFinder } from "./utils/scanner";
 import { isLightspeedEnabled, lightSpeedManager } from "../../extension";
 import { IError } from "./utils/errors";
 import { GenerationResponseParams } from "../../interfaces/lightspeed";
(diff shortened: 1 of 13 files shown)


TamiTakamiya (commented):
It's a general comment on the approach: Can't we just execute `ansible-galaxy collection list` command and parse the command output?

goneri:
> It's a general comment on the approach: Can't we just execute `ansible-galaxy collection list` command and parse the command output?

The right command would now be `ansible-creator init collection foo.bar`, but it won't create the legacy `MANIFEST.json` or broken files. I prefer to avoid the complexity of having to deal with the venv command availability/execution of the command/clean up.

src/features/lightspeed/roleGeneration.ts
```diff
@@ -93,6 +94,20 @@ async function generateRole(
   }
 }
 
+async function getCollectionsFromWorkspace() {
+  const workspaceFolders = workspace.workspaceFolders;
+  let collectionsFound: string[] = [];
+  if (workspaceFolders) {
+    const workspaceDirectories = workspaceFolders.map((f) => f.uri.fsPath);
+    const collectionFinder = new CollectionFinder(workspaceDirectories[0]);
```
TamiTakamiya, line 102: I read ["Multi-root workspaces"](https://code.visualstudio.com/docs/editor/workspaces#_multiroot-workspaces) and wondered there might be more than workspaceDirectories.  I think 99.99% of users are not using it wit Ansible extension and do not think we need to support those cases, but we may want to add `console.debug()` when `workspaceDirectories.length > 1`

TamiTakamiya (approved):
LGTM.  Left one minor comment.

ssbarnea (dismissed):
Please ensure that change is test covered, codecov should pass.

src/features/lightspeed/roleGeneration.ts
goneri: Ah thanks! I forgot about this one. I just adjusted the code to accept a list of workspace directories.

PR #1646, opened 2024-11-07
Bump vscode-extension-tester from 8.3.0 to 8.9.0

alisonlhart:
Bumps the vscode-extension-tester dependency from 8.3.0 to 8.8.2.

PR #1648, opened 2024-11-07
dep: bump vscode-extension-tester to 8.8.2

goneri:
Upgrade `vscode-extension-tester` to get a `max` version of VSCode that is supported by Marketplace's `ms-python.python`.


goneri:
See: https://github.com/ansible/vscode-ansible/pull/1646

PR #1649, opened 2024-11-07
test: use test-launch.sh to run the tests

goneri:
Use one single script called `./tools/test-launch.sh` to trigger all the different tests:

- `COVERAGE=1` or `-c` to enable the code coverage
- `COVE_VERSION` to set the vscode-extension-tester parameter (e.g `max` or `min`)
- `MOCK_LIGHTSPEED_API` (`1` or ``): set it to `1` to dynamically start the mocked server and set `TEST_LIGHTSPEED_URL`
- `TEST_TYPE`: `e2e` or `ui` (default)

The goal is to reduce the amount of reentrant `yarn` calls and provide a more linar execution flow.
Bonus points:
- we don't need `start-server-and-test` anymore
- we avoid a lot of complexity with how we deal with `TEST_LIGHTSPEED_URL` on MacOS.


ssbarnea (commented):
I love the direction. My only question is why the yarn.lock file is modified when you did not modify package.json in any way?

package.json
```diff
@@ -917,7 +917,6 @@
     "marked": "^14.1.3",
     "minimatch": "^10.0.1",
     "semver": "^7.6.3",
-    "start-server-and-test": "^2.0.8",
```
goneri, line 920: @ssbarnea The change is here ^

goneri:
> I love the direction. My only question is why the yarn.lock file is modified when you did not modify package.json in any way?

https://github.com/ansible/vscode-ansible/pull/1649/files#r1834505006

ssbarnea (approved):

ssbarnea (approved):

TamiTakamiya (commented):
Thank you @goneri!  This was long due change that I should have made months ago...

PR #1650, opened 2024-11-10
** DO NOT MERGE ** Tami takamiya/chore/ubuntu 24.04

TamiTakamiya:
Try https://github.com/redhat-developer/vscode-extension-tester/pull/1599 to see if it resolve our CI issue with Ubuntu 24.04 runner.


TamiTakamiya:
I realized that the fix was already in but still a ui test was failing.... it should be another issue from the original one.  closing.

PR #1657, opened 2024-11-13
test-launcher: rebuild the .vsix on the fly if it's outdated

goneri:
Rebuild the vsix before installing it if:
- there is no .vsix
- the sources are newer that the .vsix file


alisonlhart (approved):

alisonlhart:
This is great :100:

PR #1670, opened 2024-11-21
Ansible Lightspeed Verizon RFE - Open multiple Playbook Gen wizards

manstis:
Support opening multiple "Playbook Generation" wizards.

A (very) basic test is included too :smiley_cat:

### DIFF
--- a/src/features/lightspeed/playbookGeneration.ts
+++ b/src/features/lightspeed/playbookGeneration.ts
@@ -1,6 +1,6 @@
 import * as vscode from "vscode";
 import { v4 as uuidv4 } from "uuid";
-import { Webview, Uri, WebviewPanel } from "vscode";
+import { Webview, Uri } from "vscode";
 import { getNonce } from "../utils/getNonce";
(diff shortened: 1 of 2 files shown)


src/features/lightspeed/playbookGeneration.ts
```diff
@@ -14,7 +14,6 @@ import { isError, UNKNOWN_ERROR } from "./utils/errors";
 import { getOneClickTrialProvider } from "./utils/oneClickTrial";
 import { LightSpeedAPI } from "./api";
 
-let currentPanel: WebviewPanel | undefined;
 let wizardId: string | undefined;
```
goneri, line 17: `wizardId` also needs to be unique for each instance of the Wizard.

manstis: [It still is](https://github.com/ansible/vscode-ansible/blob/e04570aa163da35b658512d4366614c0ad49111d/src/features/lightspeed/playbookGeneration.ts#L123).

PR #1676, opened 2024-11-25
AAP-35216: lightspeed: Rename PlaybookGenerationActionType

goneri:
Rename `PlaybookGenerationActionType` as `WizardGenerationActionType` since the
interface will also be used for the role generation.


goneri:
Reopened as https://github.com/ansible/vscode-ansible/pull/1683

PR #1681, opened 2024-11-26
Add a webview to generate a devfile without ansible-creator

alisonlhart:
This PR adds a new webview for generating a devfile from a template, designed for Red Hat OpenShift Dev Spaces. 

 ### New `Create Devfile` Webview
Adds a view where devfiles can be generated. The options are: `Destination Path`, `Ansible Project Name`, `Container Image`, and `Override`. 
- Destination Path: defaults to the first folder open in the current workspace, or an empty path. If the path is empty, a path must be entered into the text box before the devfile can be generated. 
- Ansible Project Name: Ansible project the devfile is being added to
- Container Image: Image to use in the devfile
- Override: If a file named `devfile.yaml` already exists at the destination path, create a new one with the form details. 

### Logging
Adds log output to the webview form showing that the devfile was created or that an error occurred. 

### Open file options
Shows a toast notification if the devfile was created successfully, and highlights an option to open it in the editor. '

### Testing
Adds tests for checking the webview elements and the create functionality. 
Adds a pre-commit hook (local) to check if the extension devfile template matches the ansible-creator devfile template. 

### Webview

image

image

### DIFF
--- a/.config/dictionary.txt
+++ b/.config/dictionary.txt
@@ -3,6 +3,8 @@ Ansible
 Anson
 CFLAGS
 Containerfile
+Devfile
+Devfiles
 Dockal
 Dpkg
 Elio
@@ -13,6 +15,7 @@ Ganesh
 HORIZONTALLINE
 IAM
 Jenkinsfile
+KUBEDOCK
 Launay
 Maciążek
 Nalawade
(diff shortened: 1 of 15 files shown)


cidrblock (approved):
looks great.

PR #1687, opened 2024-11-28
dirty.sh: silently exit 0 when run from within VSCode

goneri:
Ignore `dirty.sh` outcomes when it's called from within VSCode.

### DIFF
--- a/tools/dirty.sh
+++ b/tools/dirty.sh
@@ -6,6 +6,9 @@
 RED='\033[0;31m'
 NC='\033[0m' # No Color
 
+if [ "${VSCODE_INJECTION}" = "1" ]; then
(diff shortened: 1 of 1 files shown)


manstis (approved):
LGTM :+1: 

I couldn't find much about the _magic_ environment variable `VSCODE_INJECTION`.

What I did find suggests it does what your code suggests it does :smile_cat:

goneri:
> LGTM 👍
> 
> I couldn't find much about the _magic_ environment variable `VSCODE_INJECTION`.
> 
> What I did find suggests it does what your code suggests it does 😸

I picked the first variable that sounded reasonably safe :-).

PR #1689, opened 2024-11-29
Playbook Generation: Add "Are you sure?" Confirmation for Reset Button

manstis:
Add a prompt for Users to confirm reset of Outline in Playbook Generation.

Example screen shot:

AAP-34643

### DIFF
--- a/src/features/lightspeed/playbookGeneration.ts
+++ b/src/features/lightspeed/playbookGeneration.ts
@@ -238,6 +238,25 @@ export async function showPlaybookGenerationPage(extensionUri: vscode.Uri) {
         panel.dispose();
         break;
       }
+      case "resetOutline": {
+        vscode.window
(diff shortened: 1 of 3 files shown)


src/features/lightspeed/playbookGeneration.ts
```diff
@@ -238,6 +238,25 @@ export async function showPlaybookGenerationPage(extensionUri: vscode.Uri) {
         panel.dispose();
         break;
       }
+      case "resetOutline": {
+        vscode.window
```
manstis, line 242: You can't show a modal dialog from the _view_ so we need to send an event from the _view_ to the _controller_.

The _controller_ then shows the dialog and, if the user confirms, sends another event to reset the Outline.

ldjebran (approved):
LGTM

PR #1702, opened 2024-12-03
Label images in the devfile webview

alisonlhart:
Labels the devfile image with an "Upstream" descriptor. This pattern will be followed for the downstream image when it is added in the future. 

Creates a constant in `src/definitions/constants.ts` that defines devfile images. 

Adds a function to get the devfile image based on the webview dropdown input.


cidrblock (approved):
awesome, ty

PR #1707, opened 2024-12-05
tools/test-launcher.sh: drop the -x flag

goneri:
The script works as intended and ve we can reduce the verbosity level now.


manstis (approved):
LGTM :+1:

ssbarnea (approved):

PR #1715, opened 2024-12-06
Disable explanation thumbs up and thumbs down buttons after selection

mabashian:
The ask here was to disable the thumbs up and thumbs down buttons after a user interacts with one of them in the playbook explanation panel.  Inside the same function(s) that update the button class, I went ahead and set the `disabled` attribute which the button component seems to handle correctly.

disabled_thumbs_up_down


manstis (approved):
LGTM 👍 Sweet.

PR #1718, opened 2024-12-09
Update name of ui test upload from unit to ui

mabashian:
Speculative... I noticed when looking at the codecov interface that both the ui coverage and the unit coverage where both called "unit" - It's unclear to me if this is a bug or problematic in any way.  This could just be housekeeping.

Screenshot 2024-12-09 at 11 58 27 AM


alisonlhart (approved):
Good catch.

PR #1723, opened 2024-12-13
Add Webview to scaffold sample execution-environment.yml file using creator add subcommand

abhikdps:
### Summary

This PR adds a new webview for adding a sample execution-environment.yml file to an existing destination path.

#### Available fields:
- `Destination Path`: The path where the sample execution-environment.yml file gets added. If the destination path does not exist, an error message comes up.

### Webview

image

#### Related JIRA: [AAP-36883](https://issues.redhat.com/browse/AAP-36883)


alisonlhart:
I'm wondering if we need to keep the log output to file, copy logs, and open log file options - similar to how the devfile webview currently is. 
@audgirka What do you think?

abhikdps:
Reopening this PR on a branch off the upstream project so we can check code coverage - [#1727](https://github.com/ansible/vscode-ansible/pull/1727)

PR #1724, opened 2024-12-13
AAP-37422: Empty line at the top of the file break completion

goneri:
When the first line of the document is empty, the line is removed from the `documentLines`
array because of the early `trim()` call.
As a consequence, the `promptLine` value is incorrect and not completion is never triggered.
This PR address the problem.

See: c5b208d0985228110864dde9774c85cc68b09fe6

### DIFF
--- a/src/features/lightspeed/utils/multiLinePromptForMultiTasks.ts
+++ b/src/features/lightspeed/utils/multiLinePromptForMultiTasks.ts
@@ -2,9 +2,9 @@ export function shouldRequestForPromptPosition(
   documentContent: string,
   promptLine: number,
 ): boolean {
-  const documentLines = documentContent.trim().split("\n");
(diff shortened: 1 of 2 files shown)


manstis (approved):
LGTM 👍

src/features/lightspeed/utils/multiLinePromptForMultiTasks.ts
```diff
@@ -2,9 +2,9 @@ export function shouldRequestForPromptPosition(
   documentContent: string,
   promptLine: number,
 ): boolean {
-  const documentLines = documentContent.trim().split("\n");
+  const documentLines = documentContent.split("\n").map((l) => l.trim());
```
manstis, line 5: 🥳

PR #1741, opened 2024-12-23
Ensure NODE_OPTIONS is the same on local and CI

ssbarnea:


shatakshiiii (approved):
Looks good!

ssbarnea:
We will force merge it because the current failures are unrelated and this on fixes the OOM issue with webpack.

PR #1756, opened 2025-01-09
Prevent Feedback Panel from Opening on Status Bar Click

manstis:
Remove the code that launches the Feedback view when already logged in.


manstis:
I believe the `codecov` issue is being addressed.

PR #1757, opened 2025-01-09
tools/helper: accept minor == '0' in release version

goneri:
We have been pushing releases with minor version set to 0 since October.


manstis (approved):
From what I understand; LGTM :+1:

PR #1758, opened 2025-01-10
[Webview] Replace Collection Project webview to use `@vscode-elements` package

audgirka:
This PR updates Ansible collection project webview to use @vscode-elements/elements package 
insted of now deprecated @vscode/webview-ui-toolkit.

Added this new package, updated related code and styles.

For @vscode-elements package info refer: https://vscode-elements.github.io/getting-started/

Related: [AAP-37621](https://issues.redhat.com/browse/AAP-37621)


alisonlhart (commented):
It looks like the text for the ansible-creator log output is slightly less centered than it used to be. Comparing it to the original seems like the original was also off center, but it's a bit more noticeable now. If it could be centered in the logs text area completely, that would be ideal IMO.

alisonlhart (approved):

PR #1762, opened 2025-01-13
AAP-29950: Swap axios for electron fetch

mabashian:
The basic gist here is that we'll attempt to use electron's fetch when it's available.  See https://github.com/Microsoft/vscode/issues/12588 for the entire proxy support saga.  It looks like some additional options may be coming in the future but for now this seems like our best approach.  In environments (browser) where electron isn't available, we fall back to native `fetch`.  This is also true in testing environments.

Resolves https://issues.redhat.com/browse/AAP-29950


goneri (approved):

goneri:
Could you add `test/units/lightspeed/utils/handleApiError.test.ts` in https://github.com/ansible/vscode-ansible/blob/main/.sonarcloud.properties and I believe we will be able to land the PR.

alisonlhart (approved):

alisonlhart:
Great work @mabashian !

PR #1765, opened 2025-01-15
AAP-38438: roleGen: show a message if no collection found

goneri:
Display a message when no collection was found in the user workspace.


mabashian (approved):
lgtm

PR #1783, opened 2025-01-23
Update collection project webview to show divider and Logs label

audgirka:
Update collection project webview to show divider above logs.
Added Logs label above textarea which show's logs.


shatakshiiii (approved):
Looks great!

PR #1788, opened 2025-01-27
AAP-38977: test-launcher.sh: save the lcov.info file

goneri:
When TEST_TYPE="ui" and COVERAGE=1, save the lcov file associated with the test itself.


goneri:
Continuing with https://github.com/ansible/vscode-ansible/pull/1791

PR #1789, opened 2025-01-27
Automatically set lightspeed and inline suggestion checkboxes to enabled when user clicks Connect

mabashian:
https://issues.redhat.com/browse/AAP-31064

The basic gist of this change is that when we catch the "connect" message and before we initiate the login flow, we update ansible.lightspeed.enabled and ansible.lightspeed.suggestions.enabled to be true (if they're not already true).  

I found that I needed to update the status bar logic since that gets triggered by the programatic settings update and the scenario where a user is not yet logged in wasn't explicitly handled.  Without this change, the modal never appears after clicking `Connect`

As far as the tests go, I extended what was already there in the ui tests to include two scenarios:

1. Clicking Connect when both checkboxes are already checked
2. Clicking Connect when both checkboxes are unchecked

In both cases the tests confirm that the checkboxes are checked after login

### DIFF
--- a/src/features/lightspeed/utils/explorerView.ts
+++ b/src/features/lightspeed/utils/explorerView.ts
@@ -1,4 +1,11 @@
-import { Disposable, Webview, Uri, commands } from "vscode";
+import {
+  ConfigurationTarget,
+  Disposable,
+  Webview,
+  Uri,
+  commands,
+  workspace,
+} from "vscode";
 import { getUri } from "../../utils/getUri";
 import { getNonce } from "../../utils/getNonce";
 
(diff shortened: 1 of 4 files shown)


src/features/lightspeed/utils/explorerView.ts
```diff
@@ -117,16 +124,45 @@ export function setWebviewMessageListener(
...
+            await ansibleSettings.update(
+              "lightspeed.enabled",
+              true,
+              workspace.workspaceFolders
+                ? ConfigurationTarget.Workspace
+                : ConfigurationTarget.Global,
+            );
+          }
+
+          if (!lightSpeedSuggestionsEnabled) {
```
goneri, line 154: IMO, if the value is already set to `true`, nothing bad will happen.

I don't think we need to read `lightSpeedSuggestionsEnabled` first. This increases the complexity of the code, and reduce the test coverage, without bringing any tangible value IMO.

goneri (changes requested):

mabashian:
Closing in favor of https://github.com/ansible/vscode-ansible/pull/1797

PR #1791, opened 2025-01-28
Upload all coverage reports to codecov

abhikdps:
- Add some filters to remove the Webpack generated files that should not be part of the final coverage anyway.
- Generate and upload one .info file per UI test

Related JIRAs: [AAP-37628](https://issues.redhat.com/browse/AAP-37628), [AAP-38977](https://issues.redhat.com/browse/AAP-38977)

### DIFF
--- a/tools/test-launcher.sh
+++ b/tools/test-launcher.sh
@@ -126,7 +126,7 @@ if [[ "${TEST_TYPE}" == "ui" ]]; then
         fi
         refresh_settings "${test_file}"
 
-        TEST_COVERAGE_FILE=./out/coverage/ui/lcov.$(basename "${test_file}").info
+        TEST_COVERAGE_FILE=./out/coverage/ui/lcov.${test_file##*/}.info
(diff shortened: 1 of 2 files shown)


tools/test-launcher.sh
```diff
@@ -126,7 +126,13 @@ if [[ "${TEST_TYPE}" == "ui" ]]; then
         fi
         refresh_settings "${test_file}"
 
+        TEST_COVERAGE_FILE=./out/coverage/ui/lcov.$(basename "${test_file}").info
```
goneri, line 129: You can avoid the `basename` call with something like:

    mv ./out/coverage/ui/lcov.info "./out/coverage/ui/lcov.${test_file//\//.}.info"

abhikdps:
Built upon @goneri ’s PR #1788 . Thanks @goneri , for laying the groundwork with the initial changes!

tools/test-launcher.sh
abhikdps: Made a small change to your suggested method so that we only use the base filename of the test file and not append the whole path of the test file to the coverage filename.

goneri (approved):

PR #1792, opened 2025-01-29
lightspeed: remove the client attribute from the LightSpeedManager class

goneri:
Since ab17f51051b56e1cd6e35edd44863b2646b756c3 we don't delegate our requests to the LSP server.
We don't need to keep track of the client object anymore.


manstis (approved):
LGTM :+1:

PR #1794, opened 2025-01-29
Update Clear All buttons to work with vscode-elements limitations

alisonlhart:
Due to what seems like a bug in the vscode-elements package, the VSCodeTextField `.value` property does not edit the text inside the text field once the field has been edited. 

This PR introduces a workaround by accessing the shadow DOM within the VSCodeTextField that contains an `input` element, which then is saved and the `.value` property is accessed and set to an empty string when the "Clear All" button is clicked. 

Because of this same issue, the text fields for folder selection (both init path and log file path) were not updated when a folder was selected. This is resolved here. 

This also resolves an issue where the logToFileDiv would not be collapsed when the logToFile checkbox was unchecked during the clear function.

### DIFF
--- a/src/webview/apps/contentCreator/createAnsibleCollectionPageApp.ts
+++ b/src/webview/apps/contentCreator/createAnsibleCollectionPageApp.ts
@@ -19,6 +19,12 @@ window.addEventListener("load", main);
 
 let initNamespaceNameTextField: VscodeTextfield;
 let initCollectionNameTextField: VscodeTextfield;
+
+let namespaceInputField: HTMLInputElement;
+let collectionInputField: HTMLInputElement;
+let logFilePathInputField: HTMLInputElement;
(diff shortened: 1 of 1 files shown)


abhikdps (changes requested):
@alisonlhart I tested this and the changes for clearing the text fields and collapsing the logToFileDiv are working as expected. However, the issue with logFilePath and initPath not updating correctly still persists.
When selecting a logFilePath, it also updates the initPath and vice versa - again, updating the initPath changes the logFilePath as well. Please take a look at this when you have a chance.

shatakshiiii (approved):
Looks great @alisonlhart! Tested locally and `Clear All` button is working as expected. 
Approving these changes, as log file and init path issues are unrelated to `vscode-elements` change.

abhikdps (approved):

PR #1796, opened 2025-01-30
Add new form options with builder capabilities to Execution Environment webview

tanwigeetika1618:
## Description
This PR updates the Execution Environment (EE) webview by introducing new form options with Builder capabilities. It creates an execution-environment.yml file without using `ansible-creator` command and it modifies the workflow to use `ansible-builder` commands, specifically, `ansible-builder create` to generate the context and `ansible-builder build` to build the image after the `execution-environment.yml` file is created. Additionally, the implementation has been improved by structuring form data as JSON and converting it to YAML using the `js-yaml` package, eliminating the direct use of `ansible-creator`. And it updates the webview form to use `vscode-elements` instead of `vscode-webview-toolkit` UI.

### **1. New Form Options in VSCode EE Webview**
The following options have been added to the Execution Environment form:
- **Base image** (dropdown)
- **Custom Base image** (text input)
- **Suggested collections** (predefined options in form of checkboxes)
- **Additional collections** (user input)
- **System packages** (user input)
- **Additional Python packages** (user input)
- **Tag** (user input for specifying image tags)
- **Create context** (checkbox)
- **Build image** (checkbox)

### **2. Leaving `ansible-creator` command**
- The EE webview form now collects data into a structured JSON format.
- The JSON structure is converted into a properly formatted YAML file using the `js-yaml` package.
- The `ansible-builder create` command is used to generate the build context.
- The `ansible-builder build` command is executed after creating the `execution-environment.yml` file.

### **3. UI Enhancement: Migration to `vscode-elements`**
- The webview now uses `vscode-elements` instead of `vscode-webview-toolkit` for improved compatibility and maintainability.

### **4. Testing**
- Implements relevant tests to validate webview elements and ensure full coverage of the new functionality. 

This PR enhances the flexibility and usability of the Execution Environment webview while aligning with `vscode-elements` ui enhancement for best practices.

Screenshot from 2025-01-30 01-55-46
Screenshot from 2025-01-30 01-56-19


test/ui-test/contentCreatorUiTestNoCreatorTest.ts
```diff
@@ -121,3 +152,133 @@ describe("Test devcontainer generation webview (without creator)", () => {
...
+      By.xpath("//vscode-checkbox[@id='overwrite-checkbox']"),
+    );
+    await overwriteCheckbox.click();
+    await clickButtonAndCheckEnabled(eeWebview, "create-button");
+    await sleep(1000);
+
+    await clickButtonAndCheckEnabled(eeWebview, "clear-logs-button");
+    await clickButtonAndCheckEnabled(eeWebview, "clear-button");
+
+    await checkAndInteractWithEEField(eeWebview, "path-url", "~/ee-new-test");
```
abhikdps, line 209: I don't think this path will be present on the runners and since there are no assertions or expect in this test, it will pass. Please change this path. Also, try using : `os.homedir()` instead of "~" to specify the home directory.

abhikdps: Fixed this :white_check_mark:

tanwigeetika1618: Cool.. Thanks!

test/ui-test/contentCreatorUiTestNoCreatorTest.ts
```diff
@@ -121,3 +153,189 @@ describe("Test devcontainer generation webview (without creator)", () => {
...
+    );
+
+    const descriptionText = await (
+      await eeWebview.findWebElement(By.xpath("//div[@class='title-div']/h1"))
+    ).getText();
+    expect(descriptionText).to.contain(
+      "Create an Ansible execution environment",
+    );
+
+    await checkAndInteractWithEEField(eeWebview, "path-url", "~");
```
alisonlhart, line 171: This looks like it may have accidentally been reverted to `~` instead of `os.homedir()`.

alisonlhart (changes requested):
A few things I noticed:

The `Build` button doesn't activate unless text is present in the destination path, though the destination path isn't listed as a required field (since the workspace folder should be used as the default path). 

Also, I received a JSON schema error when creating context or building the image when using any builder version earlier than 3.1.0, since tags were not supported before then. Should we have a check for that version? 
```
Execution environment file created at /home/ali/code/vscode-extension/vscode-ansible/examples/
jsonschema.exceptions.ValidationError: Additional properties are not allowed ('tags' was unexpected)
```

I also think we need to disable the "Build" button while the EE builds, and maybe display a toast notification or log message indicating that the build is in process; otherwise from the user's perspective, nothing happened and they may try to reclick the button several times. 

On the same log note, I regularly received an error message along with a successful build message. For example, I got:
```
---------------------------------------- Execution environment generation logs ---------------------------------------
Execution environment file created at /home/ali/code/vscode-extension/vscode-ansible/examples
Complete! The build context can be found at: /home/ali/code/vscode-extension/vscode-ansible/examples/context

Error occurred while running the command.
``
I think this may be caused by the context/ folder already being present.

File: test/ui-test/contentCreatorUiTestNoCreatorTest.ts (hunk 2 above)
Comment (tanwigeetika1618) [reply test/ui-test/contentCreatorUiTestNoCreatorTest.ts:171]: Have fixed it!

File: test/ui-test/contentCreatorUiTest.ts [hunk 3]
```diff
@@ -95,144 +86,6 @@ describe("Test Ansible playbook and collection project scaffolding", () => {
...
-describe("Test Ansible sample execution environment file scaffolding", () => {
-  let createEEButton: WebElement;
-  let editorView: EditorView;
-  let output: WebElement;
-
-  before(async () => {
-    // Install ansible-creator
-    await workbenchExecuteCommand("Install Ansible Content Creator");
-    await sleep(2000);
-  });
```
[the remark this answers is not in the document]
tanwigeetika1618: Tried that @alisonlhart .. it seems it's not entirely providing the coverage.. cause even after adding it, the coverage went up by 0.03%.

abhikdps (approved):
Looks Awesome! Approving this because the codecov failure is unrelated/not directly related to this PR. Great work @tanwigeetika1618 ! :tada:

yarn.lock
```diff
@@ -1206,6 +1206,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@types/js-yaml@npm:4.0.9":
```
cidrblock, line 1209: WE depend on the yaml extension and it depends on the yaml language server which uses the yaml npm package.  Could we use that instead here?  https://www.npmjs.com/package/yaml
https://github.com/redhat-developer/yaml-language-server/blob/main/package.json#L43

yarn.lock
```diff
@@ -2166,7 +2173,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"ansi-regex@npm:^6.0.1":
+"ansi-regex@npm:^6.0.1, ansi-regex@npm:^6.1.0":
```
cidrblock, line 2176: I think there is a regex that can be used here rather than pulling an additional dependency.

https://github.com/chalk/ansi-regex/blob/main/index.js#L1

cidrblock (changes requested):
Couple question about the new dep, otherwise awesome.

alisonlhart:
This looks fantastic @tgeetika!!! Amazing work, especially the logic in createExecutionEnvPage :star-struck: 

I only have one concern with some of the error handling for the builder exec command - I was getting an error when building the image and even with output verbosity on high, the only error output given is "Error occurred while running the command". Maybe we could surface some of the builder error output in those cases?

yarn.lock
tanwigeetika1618: Have updated the code to use "yaml" and removed the dependency from js-yaml.

yarn.lock
tanwigeetika1618: Yes this works completely fine and there's no need to pull the ansi-regex ^6.1.0 version dependency do have removed that and updated the code accordingly.

tanwigeetika1618:
> This looks fantastic @tgeetika!!! Amazing work, especially the logic in createExecutionEnvPage 🥇
> 
> I only have one concern with some of the error handling for the builder exec command - I was getting an error when building the image and even with output verbosity on high, the only error output given is "Error occurred while running the command". Maybe we could surface some of the builder error output in those cases?

cidrblock (approved):

PR #1803, opened 2025-02-04
Lightspeed: update ligtspeed api paths version to v1

ldjebran:
Lightspeed: update ligtspeed api paths version to v1
REF: https://issues.redhat.com/browse/AAP-37996

### DIFF
--- a/src/definitions/lightspeed.ts
+++ b/src/definitions/lightspeed.ts
@@ -64,9 +64,10 @@ export namespace LightSpeedCommands {
 }
 
 export const LIGHTSPEED_API_VERSION = "v0";
+export const LIGHTSPEED_API_VERSION_V1 = "v1";
 export const LIGHTSPEED_PLAYBOOK_EXPLANATION_URL = `${LIGHTSPEED_API_VERSION}/ai/explanations/`;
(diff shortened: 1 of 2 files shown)


TamiTakamiya:
@ldjebran I guess we need to update [the API version of the "mock server"](https://github.com/ansible/vscode-ansible/blob/55672357ed923ac8591512d289831210d253f1e6/test/mockLightspeedServer/server.ts#L39), which is used in E2E/UI testing.

manstis (approved):
LGTM :+1: 

I tested this PR against our staging environment (that has `v1` support deployed) and everything (completions, contentmatch, Playbook Generation and Playbook Explanation) worked :+1:

TamiTakamiya (approved):

ldjebran:
Put this to not merge as need more discussion about compatibility with different envs.

goneri (changes requested):
This breaks the Lightspeed onPrem support \o/.

goneri:
> I agree this will break newer versions of our extension with older versions of Lightspeed but wanted to understand your concern.

Yes, this is the problem. VSCode will upgrade the extension for these users and this will break their local Lightspeed service.

ldjebran:
updated only the role end point to v1 for the moment 
@goneri please revisit

PR #1804, opened 2025-02-05
Update plugin webview: Sort and capitalise pluginType dropdown entries

shatakshiiii:


alisonlhart:
The coverage issue appears to be occurring with all content creator pages. It looks like these pages have extremely low coverage compared to what the tests should provide. The Lightspeed team is planning to overhaul our UI test process soon, and with the introduction of Vite over Webpack, we will have the opportunity to improve a lot of our tests. 

For this reason, I'm approving with the patch coverage failure.

alisonlhart (approved):

PR #1810, opened 2025-02-06
Refactor vscode-elements textfield value workaround

alisonlhart:
This PR removes workarounds related to https://github.com/vscode-elements/elements/issues/271, as that was resolved in vscode-elements 1.12.0. 

Now the VSCodeTextField `.value` attribute can be accessed and utilized correctly to update the inner <input> element. 

This PR also resolves a small issue related to this workaround in the Collection and Playbook Project webviews, where when selecting a folder for the init path, the init path would not be updated. 

An import for webview-ui-toolkit in the Quick Links webview is also removed in this change, since none of those elements are utilized in that webview. 

The collection project webview also now displays the correct init collection path rather than only showing the default path.

### DIFF
--- a/package.json
+++ b/package.json
@@ -938,7 +938,7 @@
     "@redhat-developer/vscode-redhat-telemetry": "^0.9.1",
     "@shikijs/core": "2.2.0",
     "@types/ini": "^4.1.1",
-    "@vscode-elements/elements": "^1.11.0",
+    "@vscode-elements/elements": "^1.12.0",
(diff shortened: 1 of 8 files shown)


tanwigeetika1618 (approved):

audgirka (approved):

shatakshiiii (approved):

abhikdps (approved):
lgtm! 🎉

PR #1824, opened 2025-02-13
Fix lightspeed status bar test

audgirka:
Related: https://github.com/ansible/vscode-ansible/actions/runs/13309992534/job/37169851038#step:13:949


ssbarnea (approved):
We should avoid using sleep but this fix is still better than no fix.

goneri (approved):

PR #1826, opened 2025-02-13
Bump mocha timeout to 40s as some tests occasionally run longer than 30

mabashian:
`lightspeedUiTestPlaybookExpTestNoExpTest.ts` is periodically taking longer than 30 seconds to run and is resulting in failures.  This is not ideal but is a mitigation for now until we can spend some time improving the performance of our tests.

### DIFF
--- a/.mocharc.yml
+++ b/.mocharc.yml
@@ -1,7 +1,7 @@
 extension: ["ts"]
 require: ts-node/register
 package: "./package.json"
-timeout: 30003 # default is 2000
+timeout: 40003 # default is 2000
 # most UI tests are >22s due to our current wait times and we do not want
 # red slow marker to distract us until we sort that part yet. Red is expected
 # to appear on unexpected long tests, not on an expected duration.


ssbarnea (changes requested):
Our test durations already become unreasonable long and that is affecting the maintenance and any other contributions.

We really need to implement check loops for slow tests with individual timeouts.

Changing the default timeout for all test is a no as it does have a waterfall effect. it is already insanely high (should be 2s!).

I am inclined to even vote against introducing a requirements that would not allow addition of any code that does not have tests that are **polling for expected result and fail on timeout**.

If we have polling, it would be fine to increase the timeout for a specific test that proves to take more time.

mabashian:
@ssbarnea totally fine with pushing back here, I'm not personally in a rush.  That said, I doubt there's an easy solution to be found and to do this properly will (for me) likely take significant time.  The moment we introduce sleep(s) into selenium based ui tests we're asking for instability with timeouts, etc.  I won't be back around until Monday so if we're waiting for me to find a solution it won't be until next week at the very earliest and even then I have other priorities.  Totally agree that this is not a solution but an acknowledgment of the problem and kicks the can down the road.  Just trying to help unblock PRs in the short term.

ssbarnea (approved):

Issue #1830, opened 2025-02-17
Usage of sh instead of bash is failing on some systems


noonedeadpunk:
### Summary

With a change of usage `sh` intead of `bash` which was introduced with #1823 is failing on operating systems, like Ubuntu.

As `source` is a `bash` internal command, which simply does not exist with `sh`.  So `which` will not provide any path for the `source`, as there's no standalone binary today.

### Extension version

25.2.0

### VS Code version

1.97.2

### Ansible Version

```shell
ansible [core 2.17.1]
  config file = None
  configured module search path = ['/home/dr5005/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/dr5005/.virtualenvs/ansible/lib/python3.11/site-packages/ansible
  ansible collection location = /home/dr5005/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/dr5005/.virtualenvs/ansible/bin/ansible
  python version = 3.11.11 (main, Dec  4 2024, 08:55:08) [GCC 13.2.0] (/home/dr5005/.virtualenvs/ansible/bin/python3.11)
  jinja version = 3.1.4
  libyaml = True
```

### OS / Environment

Ubuntu 24.04

### Relevant log output

```shell
$ which source
$ apt install source^C
$ source --help
source: source filename [arguments]
    Execute commands from a file in the current shell.
    
    Read and execute commands from FILENAME in the current shell.  The
    entries in $PATH are used to find the directory containing FILENAME.
    If any ARGUMENTS are supplied, they become the positional parameters
    when FILENAME is executed.
    
    Exit Status:
    Returns the status of the last command executed in FILENAME; fails if
    FILENAME cannot be read.
$ sh -c 'source ~/.virtualenvs/ansible/bin/activate && ansible-config'
sh: 1: source: not found
$ bash -c 'source ~/.virtualenvs/ansible/bin/activate && ansible-config'
usage: ansible-config [-h] [--version] [-v] {list,dump,view,init} ...
ansible-config: error: the following arguments are required: action
 
usage: ansible-config [-h] [--version] [-v] {list,dump,view,init} ...

View ansible configuration.

positional arguments:
  {list,dump,view,init}
    list                Print all config options
    dump                Dump configuration
    view                View configuration file
    init                Create initial configuration

options:
  --version             show program's version number, config file location, configured module search path, module location, executable location and exit
  -h, --help            show this help message and exit
[...]

Issue #1827, opened 2025-02-14
Cannot activate venv in 25.2.0


CaKrome:
### Summary

cannot activate venv

I believe this is related to https://github.com/ansible/vscode-ansible/pull/1823

sh does not have a source command

### Extension version

25.2.0

### VS Code version

1.97.2

### Ansible Version

```shell
ansible [core 2.18.2]
  config file = None
  configured module search path = ['/home/xxx/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/xxx/xxx_venv/lib/python3.11/site-packages/ansible
  ansible collection location = /home/xxx/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/xxx/xxx_venv/bin/ansible
  python version = 3.11.2 (main, Nov 30 2024, 21:22:50) [GCC 12.2.0] (/home/xxx/xxx_venv/bin/python3)
  jinja version = 3.1.5
  libyaml = True
```

### OS / Environment

Windows 11, Debian 12 bookworm on WSL 2.4.11

### Relevant log output

```shell
Command failed: sh -c 'source /home/xxx/xxx_venv/bin/activate && ansible-config dump' sh: 1: source: not found
```

JasonGantner:
I just ran into the same issue on a fresh install.

I've located the origin of the command here : [src/features/utils/commandRunner.ts](https://github.com/ansible/vscode-ansible/blob/3f383956d82496f9acbb185071f91f9d1150a22e/src/features/utils/commandRunner.ts#L31)

I'd suggest either replacing `source` with `.` to follow POSIX standards or switching from `sh` to `bash` since it seems that it's what is used by the devs to test.

noonedeadpunk:
So `source` is a bash internal command, which is not available in `sh` on some systems.

So either `bash` should be kept to be leveraged, or, indeed, to use `.` instead of `source`.

thendricks0:
Why is there no v25.2.1 release for this issue? It breaks the latest release for anyone who uses .venvs.

PR #1831, opened 2025-02-17
Revert "Use default sh instead of bash (#1823)"

noonedeadpunk:
This reverts commit 15856a6c3dfab5da5da8b7003aae786f8ca8b10b due to the bug found after the change.

Closes-Bug: #1830 
Closes-Bug: #1827


ssbarnea (changes requested):
The correct fix for it is documented at https://stackoverflow.com/questions/13702425/source-command-not-found-in-sh-shell 

What worries me a lot of why this change did not fail any tests! It means that none of these lines had coverage!!!

PR #1841, opened 2025-02-18
roleGen: continue to use webview-ui-toolkit

goneri:


alisonlhart:
@goneri Is there a reason the vscode-elements package can't be used instead of webview-ui-toolkit?

mabashian (approved):

goneri:
Waiting for @mabashian instead to avoid any conflict.

PR #1842, opened 2025-02-18
roleGen: reset the outline if use go back on first page

goneri:
When the user goes back to the first page, we need to ignore the `outline` used
previously or we will mess up the next role generation request.


mabashian (approved):

goneri:
Closing, another PR is coming.

PR #1852, opened 2025-02-27
Remove most sleep() function calls from ui tests

mabashian:
Relates to https://issues.redhat.com/browse/AAP-40154

The suggestion from the team that maintains `vscode-extension-tester` was to utilize `VSBrowser.instance.driver.wait` rather than indiscriminate sleep calls.  Here's the documentation for that function:

https://github.com/SeleniumHQ/selenium/blob/94007cefdccfc9360f7d69a898406ec9c7b90853/javascript/node/selenium-webdriver/lib/webdriver.js#L419-L462

Attn: `In the event a condition returns a {@linkplain Promise}, the polling loop will wait for it to be resolved and use the resolved value for whether the condition has been satisfied.`. I think this is exactly what we want.  We'll only wait long enough for the statement to become true and then move forward.  We still need to figure out what an appropriate timeout looks like but for now I just copied what was previously there.

# Summary of Changes

- Replaced explicit sleep calls with a waitForCondition utility to improve test reliability and eliminate hardcoded delays.
  - waitForCondition will poll (on a pollInterval) for a set amount of time looking for a truthy evaluation of the passed condition.
- Introduced better handling of asynchronous operations in UI tests, such as waiting for elements and notifications.
- Improved test setup and teardown processes by dismissing notifications and cleaning up temporary files in certain test cases.
- Enhanced checks for UI elements and error notifications in tests.
- Removed lightspeedUiTestPlaybookGenTest.ts.  The test cases covered there are already covered by the PlaybookGenWebview part 1 and 2 tests.
- Consolidated duplicate code and streamlined functions across test files.
- Improved error handling in file operations (e.g., using fs.promises for asynchronous file deletion).

This should be resulting in a ~20-25% reduction in run times for the UI tests in both linux and macos env's


test/ui-test/walkthroughUiTest.ts
```diff
@@ -51,10 +61,15 @@ describe("Check walkthroughs, elements and associated commands", async () => {
...
-      await sleep(1000);
-
       // Select the editor window
-      const welcomeTab = await editorView.getTabByTitle("Walkthrough: Ansible");
+      const welcomeTab = await VSBrowser.instance.driver.wait(
+        async () => {
+          return await editorView.getTabByTitle("Walkthrough: Ansible");
+        },
+        undefined,
+        "Timed out waiting for walkthrough tab to open",
```
ssbarnea, line 70: Where is the timeout value? While removing the sleep is what we want, we still need a a maximum wait time and that is not visible inside the example. I suspect that the 'undefined' parameter might be the one but that hides that critical value.

I think that we should never rely on the default and make all the timeouts explicit. OK if we define them as variables and reuse them, but still not relying on 'undefined' as this is opaque and does not inform the 'random' developer looking at the code about its existence.

alisonlhart (approved):

PR #1866, opened 2025-03-06
Add support for lightspeed role explanation

mabashian:
Addresses https://issues.redhat.com/browse/AAP-40496

This PR adds UI support for role explanation.  It includes basic logic to detect whether a focused file is part of a "role" by traversing the file directory structure looking for the hallmarks of an Ansible role like a `/roles` directory along with one of the required sub-directories (see https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html#role-directory-structure).

Note that the right-click menu item shows up all the time.  I'll be filing a follow-up ticket to address that functionality in a later PR.

Also of note, the corresponding lightspeed API endpoint is currently being built.  Until that's available the request to actually fetch the explanation will fail

### DIFF
--- a/src/definitions/lightspeed.ts
+++ b/src/definitions/lightspeed.ts
@@ -54,6 +54,8 @@ export namespace LightSpeedCommands {
   export const LIGHTSPEED_PLAYBOOK_GENERATION =
     "ansible.lightspeed.playbookGeneration";
   export const LIGHTSPEED_ROLE_GENERATION = "ansible.lightspeed.roleGeneration";
+  export const LIGHTSPEED_ROLE_EXPLANATION =
+    "ansible.lightspeed.roleExplanation";
   export const LIGHTSPEED_SIGN_IN_WITH_REDHAT =
     "ansible.lightspeed.signInWithRedHat";
   export const LIGHTSPEED_SIGN_IN_WITH_LIGHTSPEED =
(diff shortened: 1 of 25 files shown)


src/features/lightspeed/utils/data.ts
```diff
@@ -503,3 +492,30 @@ function shouldTriggerMultiTaskSuggestionForPlaybook(
...
+    if (fs.existsSync(dirPath)) {
+      const yamlFiles = await fs
+        .readdirSync(dirPath)
+        .filter((file) => /\.(yml|yaml)$/.test(file));
+      yamlFiles.forEach((file) => {
+        const fileContents = readVarFiles(`${dirPath}/${file}`);
+        if (fileContents) {
+          files.push({
+            path: `${dirPath}/${file}`,
+            file_type: dir.slice(0, -1) as "default" | "task",
```
goneri, line 512: We already have the `GenerationFileType` enum that we can rename and reuse here.

src/definitions/lightspeed.ts
```diff
@@ -68,6 +70,7 @@ export const LIGHTSPEED_API_VERSION_V1 = "v1";
 export const LIGHTSPEED_PLAYBOOK_EXPLANATION_URL = `${LIGHTSPEED_API_VERSION}/ai/explanations/`;
 export const LIGHTSPEED_PLAYBOOK_GENERATION_URL = `${LIGHTSPEED_API_VERSION}/ai/generations/`;
 export const LIGHTSPEED_ROLE_GENERATION_URL = `${LIGHTSPEED_API_VERSION_V1}/ai/generations/role/`;
+export const LIGHTSPEED_ROLE_EXPLANATION_URL = `${LIGHTSPEED_API_VERSION_V1}/ai/explanations/role`;
```
goneri, line 73: There is a trailing `/` at the end of the URL.

src/features/lightspeed/roleExplanation.ts
```diff
@@ -0,0 +1,266 @@
...
+        markdown = response.content;
+        if (markdown.length === 0) {
+          markdown = "### No explanation provided.";
+        }
+        const html_snippet = marked.parse(markdown) as string;
+        currentPanel.setContent(html_snippet, true);
+      }
+    });
+  } catch (e) {
+    console.log(e);
```
goneri, line 118: debug print?

src/features/lightspeed/roleExplanation.ts
```diff
@@ -0,0 +1,266 @@
...
+  let markdown = "";
+  lightSpeedManager.statusBarProvider.statusBar.text = `$(loading~spin) ${lightSpeedStatusbarText}`;
+  try {
+    await generateRoleExplanation(
+      lightSpeedManager.apiInstance,
+      files,
+      roleName,
+      explanationId,
+    ).then(async (response: ExplanationResponseParams | IError) => {
+      console.log(response);
```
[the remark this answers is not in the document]
mabashian: Ah yea this is copy/paste from https://github.com/ansible/vscode-ansible/blob/main/src/features/lightspeed/playbookExplanation.ts#L99.  I'll remove

src/features/lightspeed/roleExplanation.ts
mabashian: Same with this - copy paste from https://github.com/ansible/vscode-ansible/blob/main/src/features/lightspeed/playbookExplanation.ts#L119.  I'll remove

goneri (approved):

PR #1871, opened 2025-03-07
primevue 4.3.1 with a patch for 7294

goneri:
The `AutoComplete` component is currently broken upstream, the patch is a backport of their fix.

This commit is a follow up to a5d283b515e11bd39c05204b3a9dd22031d4f7ca.


manstis (approved):
"Their fix" is https://github.com/primefaces/primevue/pull/7365

PR #1876, opened 2025-03-11
primevue: embed a local copy of AutoComplete

goneri:
Embed a local copy until Primevue 4.3.2 is released.

The patch imported with 6e5a936061fcd5e1deb3f5e560371e96d78e0c70 was not applied properly when building the .vsix.


goneri:
New PR https://github.com/ansible/vscode-ansible/pull/1882

PR #1877, opened 2025-03-11
lightspeed: remove experimental flag for roleExp/Gen

goneri:


src/features/lightspeed/utils/explorerView.ts
```diff
@@ -88,6 +88,12 @@ export function getWebviewContentWithActiveSession(
   <script type="module" nonce="${nonce}" src="${webviewUri}"></script>
   </form>
   </div>`;
+  const generateRoleForm = `<div class="button-container">
+  <form id="playbook-generation-form">
```
manstis, line 92: `id="role-generation-form"`?

manstis (approved):
LGTM :+1: 

(With one minor, almost cosmetic comment).

src/features/lightspeed/utils/explorerView.ts
goneri: oh good catch.

PR #1879, opened 2025-03-12
lightpseed gen: no not "No results found" message

goneri:
Don't display anymore the `No results found` message when the prompt
history is empty.


goneri:
Superseded by https://github.com/ansible/vscode-ansible/pull/1881

Issue #768, opened 2023-01-30
Support for ansible-vault when Ansible config is not in the workspace root


Tina-otoge:
### Problem

I'm editing an Ansible project that is in a subfolder of my VS Code workspace directory

My VS Code workspace looks like this
```
├── ansible-project1
│   ├── ansible.cfg
│   ├── docs
│   ├── files
│   ├── group_vars
│   ├── hosts
│   ├── molecule
│   ├── playbook_install.yml
│   ├── README.md
│   ├── requirements.txt
│   ├── roles
│   ├── tasks
│   ├── templates
│   └── .ansible_vault_key.txt
└── ansible-project2
    ├── ansible.cfg
    ├── group_vars
    ├── hosts
    ├── molecule
    ├── playbook_install.yml
    ├── README.md
    ├── requirements.txt
    ├── roles
    ├── tasks
    ├── templates
    └── .ansible_vault_key.txt
```
If I'm editing `ansible-project1/group_vars/prod/vault.yml`, and try to run the extension command `Ansible Vault: Encrypt/decrypt via 'ansible vault'` from VS Code, it fails with message:
> no valid ansible vault config found, cannot de/-encrypt

### Solution

I believe the extension could benefit from testing for any valid ansible config starting from the file we are trying to encrypt/decrypt, and escalate to the parent folder, until it reaches the workspace root.

### Alternatives

Currently, the alternative is either running ansible-vault manually from the command line, or re-opening a new VS Code instance in one of the subfolders before using the Ansible Vault extension features.

priyamsahoo:
Hey @Tina-otoge, I believe you might have opened the parent directory as the workspace's root. With the project structure you mentioned, if you open both the projects as two different workspaces, the extension would successfully pick the associated config files (which you can confirm with the status bar present below).

Tina-otoge:
Indeed, as stated in the "Alternatives" section of my ticket, it works fine if I open the ansible directory directly as the workspace root.

I was hoping the extension could support working on projects that are made of multiple Ansible sub-projects, by looking for an ansible config starting from the file you are editing and its parents, instead of by looking at the project root only, which would improve my workflow in my use case a lot.

andrea-berling:
Also interested in this

vrelk-net:
[...]

Issue #701, opened 2022-11-30
`ansible-vault: command not found`, perhaps unique to when running in Python venv


eengstrom:
### Summary

TLDR: `ansible-vault: command not found`

Details:

When editing a file that is encrypted with `ansible-vault` (e.g. a `group_vars/GROUP/vault.yml`, I try to issue the command **"Ansible Vault: encyrpt/decrypt via `ansible-vault`"**, but get in return a `/bin/sh: ansible-vault: command not found`.

Yes, the language on the yaml file that I'm trying to decrypt is indeed `Ansible`.

Note that I run my Ansible out of a Python venv, but that works fine for other tasks like `ansible-lint`. I've looked at the code base for this extension, but cannot find where calls to `ansible-lint` are passing through the `venv` and perhaps similar calls to `ansible-vault` are not.

### Extension version

v1.0.90

### VS Code version

1.73.1

### Ansible Version

```shell
ansible 2.10.17
  config file = $HOME/Projects/sysops/ansible/ansible.cfg
  configured module search path = ['$HOME/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = $HOME/Projects/sysops/ansible/.venv/lib/python3.8/site-packages/ansible
  executable location = $HOME/Projects/sysops/ansible/.venv/bin/ansible
  python version = 3.8.9 (default, Apr 13 2022, 08:48:06) [Clang 13.1.6 (clang-1316.0.21.2.5)]
```

### OS / Environment

macOS Monterey (12.6.1), MBP M1 Max

```
% ansible-vault --version
ansible-vault 2.10.17
  config file = $HOME/Projects/sysops/ansible/ansible.cfg
  configured module search path = ['$HOME/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = $HOME/Projects/sysops/ansible/.venv/lib/python3.8/site-packages/ansible
  executable location = $HOME/Projects/sysops/ansible/.venv/bin/ansible-vault
  python version = 3.8.9 (default, Apr 13 2022, 08:48:06) [Clang 13.1.6 (clang-1316.0.21.2.5)]
```

### Relevant log output

```shell
Decryption of $HOME/Projects/sysops/ansible/playbooks/group_vars/gitlab/vault.yml failed: Error: Command failed: ansible-vault decrypt "/Users/engstrom/Projects/sysops/ansible/playbooks/group_vars/gitlab/vault.yml"
[...]

PR #1885, opened 2025-03-15
Fix for vault command

pgalbraith:
The vault command is not working when using virtual environments, per #701.

This may address, in some measure, #768 but I'm not confident I fully understand the issue there.


pgalbraith:
Resubmitting from another branch...

Issue #1896, opened 2025-03-20
VSCode demands login for Ansible Lightspeed with watsonx Code Assistant


wolfgang-demeter:
### Summary

VSCode `1.98.2` updated the [Ansible extension](https://marketplace.visualstudio.com/items?itemName=redhat.ansible) to version `25.3.0`. Since that moment there is a demand to login to something called **Ansible Lightspeed with watsonx Code Assistant**. I don't want to use it. I don't intend to use it. **How can i get rid of this message/nudge?**

![Image](https://github.com/user-attachments/assets/9c7f8bc5-878a-49bf-9381-c8cee2d3ec39)

**Edit:**
* I already disabled `ansible.lightspeed.suggestions.enabled`.
* I removed the URL `ansible.lightspeed.URL` completely; funny enough the nudge tries to open an empty URL.

### Extension version

25.3.0

### VS Code version

1.98.2

### Ansible Version

```shell
$ ansible --version
ansible [core 2.17.9]
  config file = /home/wolfgang-demeter/GitLab/ansible-devops/ansible.cfg
  configured module search path = ['/home/wolfgang-demeter/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /home/wolfgang-demeter/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

### OS / Environment

Windows 11 24H2 with WSL

### Relevant log output

```shell

```

jmighion:
Happened before https://github.com/ansible/vscode-ansible/issues/859

goneri:
Hi @wolfgang-demeter!

Thank you for the bug report. We are working on two different adjustments that will help on this:
- first, we will simplify the access to the "Disable Lightspeed Suggestion" configuration check box https://github.com/ansible/vscode-ansible/pull/1899
- and we will also restore back the "Disable Lightspeed" option. When it will be check, we will hide the "Ansible Lightspeed" webview, and disable the Auth provider.

timway:
Additionally creates a highlighted *nag* in the bottom of UI when in a file the extension works with (like a playbook). This should respect Lightspeed being disabled and not be shown in those circumstances.

[...]

PR #1899, opened 2025-03-20
lightspeed: login popup now exposes a "Disable Lightspeed" option

goneri:
image

See: https://github.com/ansible/vscode-ansible/issues/1896


manstis (approved):
LGTM :+1 (As a way to route users to the option to disable Lightspeed).

mabashian (approved):
Nice!

PR #1903, opened 2025-03-24
lightspeed/auth: use vscode-redhat-account if Che env is detected

goneri:
Use the `redhat.vscode-redhat-account` and Oauth2 device Connect when
an Eclipse Che environment is detected (using the `CHE_DASHBOARD_URL`
envvar).


goneri:
Closing as this will likely break the combination of OpenShift Dev Spaces + onPrem Lightspeed + AAP.
At this stage, we can manually set https://github.com/ansible/vscode-ansible/pull/1905 to enable the Device Flow mode.

PR #1908, opened 2025-03-25
Add support to scaffold Test plugin in Collection plugin webview

tanwigeetika1618:
Related Jira issue : https://issues.redhat.com/browse/AAP-41976 

## Description
Add Support for Test Plugin Scaffolding in Collection Plugin Webview

## Summary
This Pull Request enhances the existing Collection Plugin webview by adding support for Test Plugin scaffolding. Users can now select `Test` as a plugin type from the existing plugin-type dropdown to scaffold a test plugin.

## Changes
- Updated the plugin-type dropdown to include the "Test" option.

## Testing
- Verified that the "Test" option is available in the plugin-type dropdown.
- Confirmed that selecting "Test" triggers the correct scaffold generation.

Screenshot 2025-03-25 at 9 39 57 PM

### DIFF
--- a/src/features/contentCreator/addPluginPage.ts
+++ b/src/features/contentCreator/addPluginPage.ts
@@ -142,6 +142,7 @@ export class AddPlugin {
                       <vscode-option>Filter</vscode-option>
                       <vscode-option>Lookup</vscode-option>
                       <vscode-option>Module</vscode-option>
+                      <vscode-option>Test</vscode-option>
                     </vscode-single-select>
                   </div>
                 </div>
@@ -359,6 +360,7 @@ export class AddPlugin {
       filter: "24.12.1",
       action: "25.0.0",
       module: "25.3.0",
+      test: "25.3.1",
     };
     const requiredCreatorVersion =
       minRequiredCreatorVersion[pluginType.toLowerCase()];
--- a/test/ui-test/contentCreatorUiTest.ts
+++ b/test/ui-test/contentCreatorUiTest.ts
@@ -458,6 +458,15 @@ describe("Test collection plugins scaffolding", () => {
       "Module",
     );
   });
+  it("Check add-plugin webview elements for test plugin", async () => {
+    await testWebViewElements(
+      "Ansible: Add a Plugin",
+      "~",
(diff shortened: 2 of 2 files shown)


src/features/contentCreator/addPluginPage.ts
```diff
@@ -359,6 +360,7 @@ export class AddPlugin {
       filter: "24.12.1",
       action: "25.0.0",
       module: "25.3.0",
```
shvenkat-rh, line 362: Can the module version also be changed to 25.3.1 since 25.3.0 has a bug with module scaffolding?

shvenkat-rh (approved):

shvenkat-rh:
LGTM 🚀

src/features/contentCreator/addPluginPage.ts
tanwigeetika1618: okay.. will update the version

test/ui-test/contentCreatorUiTest.ts
```diff
@@ -458,6 +458,15 @@ describe("Test collection plugins scaffolding", () => {
       "Module",
     );
   });
+  it("Check add-plugin webview elements for test plugin", async () => {
+    await testWebViewElements(
+      "Ansible: Add a Plugin",
+      "~",
```
abhikdps, line 464: can we use os.homedir() instead of "~"?

tanwigeetika1618: Yes definitely we can.. It's the same thing.. Shall I update "~" with os.homedir() at all places?

abhikdps: If it is being used at multiple places then instead of calling `os.homedir()` multiple times, let’s store it in a variable and use that throughout for better efficiency and readability. Using `os.homedir()` will also ensure compatibility across all platforms, as `~` might not expand correctly in all cases

tanwigeetika1618: Have replaced `"~"` with `os.homedir()` and stored the result in a variable for better efficiency. The change has been applied throughout the code.

PR #1928, opened 2025-04-09
Fix for vault command

pgalbraith:
The vault command is not working when using virtual environments, per #701.

This may address, in some measure, #768 but I'm not confident I fully understand the issue there.


src/extension.ts
```diff
@@ -88,6 +88,9 @@ export const globalFileSystemWatcher: IFileSystemWatchers = {};
 const lsName = "Ansible Support";
 let lsOutputChannel: vscode.OutputChannel;
 
+// Expose settings
+export const extSettings = new SettingsManager();
```
alisonlhart, line 92: The settingsManager should not be exported here. Look at how the src/features/runner.ts file obtains the settings to match that implementation.

alisonlhart (changes requested):
Thanks for submitting this! I have left a comment here with needed changes, and the team will re-review after that is complete.

src/extension.ts
pgalbraith: Thanks @alisonlhart.  I converted some (the bare minimum) of vaullt.ts to behave more like runner and use the SettingsManager in local context, I trust that's what you're aiming for (though I didn't put all of the vault functions into the class).  A more targeted alternative here that minimiizes changes might be to just hold extSettings as a local var in vault.ts and change that directly from extension.ts.  Not pretty but keeps this changes focused on fixing vault rather than mixing in class refactoring to the PR.  Let me know what you think.

alisonlhart: I think this is acceptable @pgalbraith.

PR #1931, opened 2025-04-14
Combine explanation webviews and convert to vue.js

mabashian:
Resolves https://issues.redhat.com/browse/AAP-42282

- Removes old playbook/role explanation webviews
- Removes old styling associated with old webviews
- Adds a new Explanation webview using vue.js which can handle both playbook and role explanations
- Lines up vite compiled webviews CSP with non-vite CSP's
- Solves for the codicons issue by leveraging `/media` as the configured public dir.  This means we can re-use the codicon assets available in that directory.  I tried many different ways to get codicons working w/ vite & vue using es6 module imports but could not get it working the way that I wanted to.  Perhaps someone else can figure this out in the future so we can move away from static codicon assets.
- Updates the tests to match new webview

Here it is in action:

explanation_demo


goneri (approved):

goneri:
Great work, You can list `src/features/lightspeed/vue/views/explanationPanel.ts` in the `.sonarcloud.properties` to avoid the code dup error.

ssbarnea:
@mabashian This still have some duplication that needs the addressed to merge it.

mabashian:
Addressed the sonarcloud stuff but i noticed some broken styling on the generation pages so working on that before merging

PR #1962, opened 2025-04-30
build: enable webpack caching

ssbarnea:
This change reduced the time required to perform webpack from ~80s to
~20s. Further changes can be implemented to improve it based on
https://webpack.js.org/guides/build-performance/


tanwigeetika1618 (approved):
Switching to filesystem caching makes a lot of sense here, especially with the config dependency tracked via __filename.

alisonlhart (approved):

PR #1975, opened 2025-05-06
Avoid potentially unnecessary webpack step inside task-ui

tanwigeetika1618:
### PR Description

The `yarn webpack-dev` command in the `task-ui` step was redundant, as it was already being executed during the `task package` step. So instead of that `task package` command is being run as the first step in `task-ui`. This update removes the duplication and streamlines the process. Instead 

#### Timings Improvement:
- **test(linux)**: From ~15 min 6 sec to ~13 min 50 sec
- **test(macos)**: From ~18 min 22 sec to ~16 min 17 sec

After this change it is observed that the timings slightly improve across both Linux and macOS environments.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -159,7 +159,7 @@ tasks:
     desc: Run UI tests {{.XVFB}}
     aliases: [ui]
     cmds:
-      - yarn webpack-dev
+      # - yarn webpack-dev
       - "{{.VIRTUAL_ENV}}/bin/python3 --version"
       - ' {{.XVFB}} bash -c ''. "{{.VIRTUAL_ENV}}/bin/activate" && COVERAGE=1 MOCK_LIGHTSPEED_API=1 ./tools/test-launcher.sh'''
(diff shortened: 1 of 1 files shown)


abhikdps (approved):

shatakshiiii (approved):

alisonlhart (approved):

tanwigeetika1618:
> Please explain the reasoning for doing this in the PR body.

Done!

ssbarnea (changes requested):
This is still wrong because I checked all the dependencies of `task-ui` and the packaging is not listed. This means that after this change, if someone clones the repo and run `task ui` they will get an error.

Is ok if you want to remove it from its current form, but you need to be sure you are doing it correctly so the operation I mentioned does not produce an error.


This can be achieve with either task dependencies or task inclusions.

abhikdps (approved):

tanwigeetika1618:
Have updated the PR description too with the latest change.

PR #1983, opened 2025-05-07
Add ADE `--im=cfg` flag for collection webview

alisonlhart:
Adds the `--isolation-mode=cfg` flag to the ADE command used in the collection project webview. 

This prevents issues with an existing environment that has collections, particularly the devcontainer setup. 

Wait message is printed to the log while the command is running:

image


cidrblock (approved):

cidrblock (approved):

ssbarnea (approved):
Assuming that it does not reduce coverage, everything else looks fine. Thanks.

PR #1994, opened 2025-05-08
test-launcher: call extest run-tests with the -c

goneri:
Pass the `-c MAX` parameter to be sure `extest run-tests` figure out what version is the right version to use. This version is important since it's used to build the path of the `chromedriver` binary.

### DIFF
--- a/tools/test-launcher.sh
+++ b/tools/test-launcher.sh
@@ -131,7 +131,12 @@ if [[ "${TEST_TYPE}" == "ui" ]]; then
             start_server
         fi
         refresh_settings "${test_file}"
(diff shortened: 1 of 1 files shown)


mabashian (approved):

ssbarnea:
> Pass the `-c MAX` parameter to be sure `extest run-tests` figure out what version is the right version to use. This version is important since it's used to build the path of the `chromedriver` binary.

In fact I want us to expose the vscode version used by relevant ui/e2e tests in:

- test job name in GHA
- junit test report (at least as a tag)
- test report filename

This should make it easier for us to spot when vscode version is causing tests to fail.

Thanks for adding it.

PR #2010, opened 2025-05-13
docs: clarify use of vite for development

goneri:


alisonlhart (approved):

ssbarnea (changes requested):
Remove the image, is not really essential and also is external to the repo.

ssbarnea (approved):

PR #2040, opened 2025-05-22
Checking : Migration from webpack to Vite +Vue JS for webviews

tanwigeetika1618:
[DO NOT MERGE]


src/features/lightspeed/vue/views/contentCreatorHelper.ts
```diff
@@ -0,0 +1,329 @@
+/* eslint-disable  @typescript-eslint/no-explicit-any */
+
```
goneri, line 2: I think it would make sense to continue to use one single helper.ts to handle the event/callback, this way it will become easier to encourage reusability between webviews. However, to improve maintainability, it would be nice to maintain the logic in dedicated modules/classes.

src/features/lightspeed/vue/views/contentCreatorHelper.ts
```diff
@@ -0,0 +1,329 @@
...
+  if (semver.gte(creatorVersion, ANSIBLE_CREATOR_COLLECTION_VERSION_MIN)) {
+    command = `ansible-creator init collection ${namespaceName}.${collectionName} ${initPathUrl} --no-ansi`;
+  } else {
+    command = `ansible-creator init ${namespaceName}.${collectionName} --init-path=${initPathUrl} --no-ansi`;
+  }
+  return command;
+}
+
+async function openLogFile(fileUrl: string) {
```
goneri, line 90: I believe this function can be renamed to become a generic `readFile`.

src/features/lightspeed/vue/views/contentCreatorHelper.ts
```diff
@@ -0,0 +1,329 @@
...
+  public static async setupWebviewHooks(
+    webview: Webview,
+    disposables: Disposable[],
+  ) {
+    webview.onDidReceiveMessage(
+      async (message: any) => {
+        const command = message.command;
+        let payload;
+
+        switch (command) {
```
goneri, line 288: This is the part that think should be moved in the existing `helper.ts`.

goneri (changes requested):
This is really cool :-). Just a minor comment regarding the way the `onDidReceiveMessage` callback should be handled.

PR #2083, opened 2025-06-21
Refactor common functions in vue migrated webviews

tanwigeetika1618:
**Related Jira issue**: [AAP-47613](https://issues.redhat.com/browse/AAP-47613)

## Description  
Refactor migrated vue-based webview utility functions into Centralized webviewUtils file

### Summary
This PR refactors duplicated logic from multiple Vue-based webviews (under content creator) into a centralized utility file: `webviewUtils.ts`. The primary goal is to improve **code modularity**, **maintainability**, and **readability** by consolidating reusable functions that were previously redefined in each individual `App.vue` file.

### Changes Introduced

- Extracted common utility functions like:
  - `openFolderExplorer`
  - `openFileExplorer`
  - `clearLogs`, `copyLogs`, `openLogFile`
  - `initializeUI`, `checkADEPresence`
- Introduced `webviewUtils.ts` as a shared utility module for all content-creator webviews.
- Replaced local function definitions in Vue components with imports from the centralized utils file.
- Separated logic clearly between:
  - Generic shared functions
  - Collection/project-specific logic
  - Role and plugin utilities
- Ensured all components integrate the shared functions properly.

### Testing

- Verified that the refactored functions behave the same across all affected Vue webviews.
- Manual testing done to ensure:
  - Folder and file explorers open correctly
  - Log operations function as expected
  - Project/collection creation logic still works without regression
  - No console or runtime errors on initialization or interaction

### Notes

This sets a foundation for future refactors and utility enhancements across all webviews. The same pattern can be extended to additional components like plugin creation for consistency.


src/features/contentCreator/webviewUtils.ts
```diff
@@ -0,0 +1,270 @@
...
+) {
+  vscodeApi.postMessage({
+    type: "init-open-scaffolded-folder-plugin",
+    payload: {
+      projectUrl,
+      pluginName: pluginName.trim(),
+      pluginType: pluginType.trim(),
+    },
+  });
+}
```
anusshukla: please check i think is not repeating function

src/features/contentCreator/webviewUtils.ts
```diff
@@ -0,0 +1,270 @@
...
+
+export function openRoleFolder(projectUrl: string, roleName: string) {
+  vscodeApi.postMessage({
+    type: "init-open-role-folder",
+    payload: {
+      projectUrl,
+      roleName: roleName.trim(),
+    },
+  });
+}
```
anusshukla: Same as above

src/features/contentCreator/webviewUtils.ts
tanwigeetika1618: Yep I agree.. removed the non-duplicated functions from webviewUtils.ts file

src/features/contentCreator/webviewUtils.ts
tanwigeetika1618: Done!

src/features/contentCreator/webviewUtils.ts
```diff
@@ -0,0 +1,245 @@
...
+        if (commonState?.logs) {
+          commonState.logs.value += message.data + "\n";
+        }
+        break;
+    }
+
+    switch (message.command) {
+      case "homedirAndTempdir":
+        if (config.onHomedirAndTempdir) {
+          config.onHomedirAndTempdir(message.homedir, message.tempdir);
```
anusshukla: Instead of using a switch-based approach, we can refactor this into a class-based design. This will make the code more maintainable and scalable in the long run, avoiding issues we might face as the number of message types grows.

For example, you can define handlers like this:

```
class MessageRouter {
  constructor(private config, private commonState) {}

  handle(message: Message) {
    switch (message.type) {
      case "homeDirectory":
        this.onHomeDirectory(message.data);
        break;
    }
  }

  private onHomeDirectory(data: string) {
    this.config?.onHomeDirectory?.(data);
    this.commonState?.homeDir && (this.commonState.homeDir.value = data);
  }
}
```

This makes each message type's logic modular and easier to test or extend individually. You can also refer to the official TypeScript documentation on classes for more guidance:
https://www.typescriptlang.org/docs/handbook/classes.html

tanwigeetika1618: Yep definitely.. Thank you for the thoughtful suggestion! I really liked the idea of moving to a class-based approach for better scalability and maintainability.

Based on your input, I went ahead and implemented a MessageRouter class and modularized all the message types and commands into dedicated private methods. This not only simplifies the handle() logic but also makes it much easier to extend and test individual message handlers moving forward.

I also added proper typing for the Message interface to improve type safety and readability. Let me know if there’s anything else I could refine further!

alisonlhart (dismissed):

src/features/contentCreator/webviewUtils.ts
```diff
@@ -0,0 +1,282 @@
...
+          break;
+        case "execution-log":
+          this.onExecutionLog(message.arguments);
+          break;
+        case "ADEPresence":
+          this.onADEPresence(message.arguments);
+          break;
+      }
+    }
+  }
```
anusshukla: Instead of using a switch block, we can simplify the structure using a handler map for better scalability and maintainability.This avoids growing switch statements and makes it easier to add or test individual handlers in the future. Let me know if you'd like help updating more of them!

webviews/AddPluginPageApp.vue
```diff
@@ -1,63 +1,38 @@
...
-      defaultPath: initPath.value || homeDir.value,
-    },
-  });
-}
-
-function handleInitOpenScaffoldedFolderClick() {
+const handleOpenFolderExplorer = () => {
+  openFolderExplorer(initPath.value || homeDir.value);
+};
+const handleClearLogs = () => clearLogs(commonState.logs);
```
anusshukla: Can we directly import the function from webviewUtils.ts and use it in the onClick handler, instead of assigning it to a constant and then importing that constant?

tanwigeetika1618: Yes definitely we can.. I did this thinking it would help keep things clean and easy to manage in the long run. By using wrapper functions with clear names like handle*, it keeps things consistent across the code, so anyone reading it can easily understand what's going on. It also makes it easier to add things like logs or checks later without changing the main template. When something goes wrong, having named functions helps us debug faster because we can see exactly where the problem happened. Plus, putting the logic in the script instead of the template makes the UI part simpler and easier to read. However can do that part for openFolderExplorer on 'onClick' handler like this 
 @click="openFolderExplorer(initPath || homeDir)" 
and 
 @click.prevent="clearLogs(commonState.logs)"
 in the template part

src/features/contentCreator/webviewUtils.ts
tanwigeetika1618: Hmm.. got it.. I am trying to fix this part.. I am raising a commit for the same where 
// we can add new handlers straightaway to the map.. like this
this.typeHandlers.newMessageType = (message) => this.onNewMessageType(message.data);

alisonlhart (approved):

webviews/AddPluginPageApp.vue
anusshukla: That’s definitely a thoughtful and forward-looking approach, and I really appreciate the consideration for consistency and maintainability.

That said, in this specific case, since the wrapper functions weren’t adding any extra logic, they seemed a bit unnecessary and added an extra layer of indirection. Using the utility functions directly keeps the code simpler and easier to follow for now.

If we find the need to add logging or validation in the future, we can always refactor as needed.

PR #2101, opened 2025-07-01
Add UI check for minimum required ansible-creator version

anusshukla:
Description - This PR adds a UI-level check to validate the minimum required version of ansible-creator for specific webviews such as Create Plugin and Create Collection.

Changes made -
- Added version validation in the UI for ansible-creator.
- If the installed version is lower than the required minimum.
image
- If installed version is correct it will work as accepted
image

- A warning banner is displayed at the top of the webview.
- The rest of the UI will be greyed out.
- All input fields and actions are disabled to prevent interaction.
- Webviews that do not rely on ansible-creator remain unaffected.


alisonlhart:
> We could do a new separate message in case `ansible-creator` is not installed at all:
> 
> > ```
> >     This feature requires ansible-creator version <b>{{ failures[0].required }}</b> or higher, but ansible-creator was not found. Please install it to use this feature.
> > ```
I like the idea of having two separate messages for upgrading vs installing. It communicates clearly what the resolution is. 

> 2. if someone installed ansible-dev-tools, are they expected to upgrade a single tool like ansible-creator?
Since the pypi ansible-dev-tools package is a meta-package including everything, it is completely fine for users to upgrade a single package rather than all of them. If there are any inter-dependencies that require a different version of the other packages, pip/uv will handle it and notify. 

> 3. And in the case of the devcontainer,  shouldn't they upgrade the image instead, or this will be done automatically when opening an existing Ansible devcontainer project with the `:latest` tag for the image?
This is a good question. I don't think the image will automatically update to a new "latest" if they already have the image pulled and cached. We should recommend upgrading the image, but I don't know how we'd detect that we're inside the devcontainer environment specifically.

PR #2110, opened 2025-07-07
Add webview for Ansible pattern scaffolding support

anusshukla:
**Description:**
Added the new Add Pattern webview to support pattern creation in Ansible collections. 
image

To prevent timeout issues during webview testing, introduced helper.ts and panelUtils.ts for better panel lifecycle management.

Previously, we were reusing the Lightspeed helper, which used the path:
`serverUrl: ${process.env.VITE_DEV_SERVER_URL}webviews/lightspeed/${name}.html`

However, since the Content Creator webviews are located directly under the webviews/ directory, we updated the path to:

`serverUrl: ${process.env.VITE_DEV_SERVER_URL}webviews/${name}.html`

This ensures correct URL resolution and improved stability during test execution.

**Related** - AAP-47363

### DIFF
--- a/webviews/AddPatternPageApp.vue
+++ b/webviews/AddPatternPageApp.vue
@@ -10,7 +10,7 @@ import {
   clearAllFields,
   createActionWrapper,
   createFormValidator} from './../src/features/contentCreator/webviewUtils';
-import "../media/contentCreator/addPluginPageStyle.css";
+import "../media/contentCreator/addPluginAndPatterPageStyle.css";
 import RequirementsBanner from './RequirementsBanner.vue';
 
 const commonState = useCommonWebviewState();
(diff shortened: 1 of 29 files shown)


webviews/AddPatternPageApp.vue
```diff
@@ -0,0 +1,227 @@
...
+});
+</script>
+
+<template>
+<body>
+  <RequirementsBanner v-if="!requirementsMet" :failures="requirementFailures" />
+  <div :class="{ 'disabled-content': !requirementsMet }">
+  <div class="title-div">
+    <h1>Add a pattern to an existing collection</h1>
+    <p class="subtitle">Extending automation with python</p>
```
alisonlhart, line 106:
```suggestion
    <p class="subtitle">Extending automation with python</p>
```
We should change this to something more relevant to patterns. We can ask @tima about this.

alisonlhart (changes requested):
Left one comment, everything looks great @anusshukla. Awesome work.

alisonlhart (approved):

PR #2116, opened 2025-07-14
Enhanced JSON Schema Support for Ansible Patterns

anusshukla:
**Description:-**
This feature introduces JSON Schema-based validation and intelligent autocompletion for Ansible patterns.

Key Enhancements:

- Context-aware autocompletion for host patterns (e.g., all, ~web.*) directly within the editor.
- Real-time validation to detect syntax errors and invalid host expressions early.
- Advanced pattern support, including regular expressions and group operators (:, !, &, ~) for precise and flexible host targeting.

related :- AAP-47364


cidrblock (changes requested):
Can you split out the pattern schema support into it's own PR seperate from the host completion?

It'll be easier to sort out issues in the future if we need to.

oh......

I think I may have not explained this well, "pattern" support for be a schema check using this

https://github.com/ansible/pattern-service/blob/main/specifications/pattern_specification.json

(thisPR is aslo cool, so maybe another for that)

ssbarnea (changes requested):
Please ask the pattern team to publish their JSON Schema on https://github.com/SchemaStore/schemastore/blob/12502c1442b85fa714a4cde33ddbe97e7bc7793f/src/api/json/catalog.json#L2689 catalog. Once is published there, we can rely on the values from there.

PR #2137, opened 2025-07-30
Dev-container Webview migration to vue

anusshukla:
Related: [AAP-41909](https://issues.redhat.com/browse/AAP-41909)
About: Migrating playbook project from web-view to Vite JS/Vue
image


src/features/lightspeed/vue/views/webviewMessageHandlers.ts
```diff
@@ -499,4 +535,132 @@ export class WebviewMessageHandlers {
...
+        "devcontainer.json",
+      );
+
+      // Create .devcontainer directory if it doesn't exist
+      if (!fs.existsSync(devcontainerDir)) {
+        fs.mkdirSync(devcontainerDir, { recursive: true });
+      }
+
+      // Create devcontainer.json content
+      const devcontainerConfig = {
```
alisonlhart: The devcontainers should be generated from the template files under resources/contentCreator/ for devcontainer. The whole .devcontainer directory should be scaffolded along with the folders/files inside in the structure they're stored in, then the substitutions for image and extensions will go into those template files before creation. So the `.devcontainer/` dir, and podman/, docker/, and devcontainer file at the root of the dir should all be generated.

See the original scaffold function here: https://github.com/ansible/vscode-ansible/pull/2137/files#diff-ecac48deb6c5ffcb0ec4fd219eb62e7dea0a92e108811b3e229ce89f9defbe0cL371

alisonlhart (changes requested):
This is a good start @anusshukla! There's a required change for scaffolding that I added a comment to.

alisonlhart (approved):
Awesome, LGTM!

PR #2167, opened 2025-09-08
EE Webview migration to vue and base image css correction

anusshukla:
Related - [AAP-41911](https://issues.redhat.com/browse/AAP-41911)
About: Migrated the Execution Environment (EE) webview from the old Webview setup to Vite.js + Vue.js for better performance and maintainability, and fixed the CSS dropdown issue in the Base Image selection.
image


execution-environment.yml
```diff
@@ -0,0 +1,16 @@
+version: 3
```
alisonlhart: This file should be removed from the base of the repo. Having the example execution-environment.yml in the examples/ directory is fine, though.

alisonlhart (changes requested):
The view looks even better than before and interacts well. 
Found one issue:
If the execution-environment.yml file already exists, the file will not be overwritten but the build still starts. 

Example:
```
---------------------------------------- Execution environment generation logs ------------------------
Error: Execution environment file already exists at /home/ali/code/vscode-ansible-main/examples and was not overwritten. Use the 'Overwrite' option to overwrite the existing file.
Complete! The build context can be found at: /home/ali/code/vscode-ansible-main/examples/context

Building execution environment, this may take a few minutes....

```

We should only start the build of context and image if the new file has been created. So if we have an existing file and don't overwrite, we should fail without running any other builder commands.

execution-environment.yml
anusshukla: Thanks for pointing that out . This file was pushed by mistake — I’ll remove it from the repo base

anusshukla:
I’ve updated the logic so that the build process will only start if a new execution-environment.yml file is actually created.
Now, if the file already exists and the overwrite option is not selected, the process exits cleanly without running any builder commands

alisonlhart (approved):

PR #2182, opened 2025-09-19
Add a new extension setting for MCP server

shatakshiiii:
### SUMMARY

- This PR adds a new extension setting to enable MCP server via command palette or extension settings.
- This setting is set to be disabled by default.
- Removed command palette entries for `ansible-creator create` and `ansible-creator sample` that were showing message- "This feature is coming soon. Stay tuned." Because sample won't be implemented anytime soon and create is recognised as 'add' and that's already implemented.

JIRA: [AAP-53787](https://issues.redhat.com/browse/AAP-53787)

Screenshot 2025-09-22 at 4 49 39 PM


src/extension.ts
```diff
@@ -1011,6 +1034,30 @@ export function deactivate(): Thenable<void> | undefined {
...
+const startMcpServer = async (extSettings: SettingsManager) => {
+  if (!extSettings.settings.mcpServer.enabled) {
+    console.log("MCP server is disabled");
+    return;
+  }
+
+  try {
+    console.log("Starting Ansible MCP server...");
+
+    console.log("MCP server startup logic to be implemented");
```
anusshukla, line 1046: The startMcpServer function currently only logs a placeholder message and does not create the required .mcp.json configuration, so VS Code and Cursor never register the MCP server. as a result mcp server will not be listed in vscode

shatakshiiii: That placeholder is intentional for now. About registering the MCP server, it will require some base MCP server implementation. That's why this PR only focuses on adding that new setting only. All the validation will be done as a follow-up work after this and the other base MCP server https://github.com/ansible/vscode-ansible/pull/2183 gets merged.

PR #2189, opened 2025-09-25
Improved reliability of YAML diagnostics e2e test

anusshukla:
Description :- This PR makes YAML diagnostics tests more reliable by fixing timing issues between the test, the extension, and the language server. It adds longer wait times, better logging, and a check to confirm the server is ready before running tests. The tests now also edit and save the document to make sure validation is triggered. With these changes, the expected diagnostics are detected more consistently, reducing test failures.

related - https://issues.redhat.com/browse/AAP-54475

### DIFF
--- a/test/e2e/diagnostics/testYamlWithoutEE.test.ts
+++ b/test/e2e/diagnostics/testYamlWithoutEE.test.ts
@@ -6,7 +6,6 @@ import {
   testDiagnostics,
   updateSettings,
   waitForDiagnosisCompletion,
-  sleep,
 } from "../../helper";
 
 export function testDiagnosticsYAMLWithoutEE(): void {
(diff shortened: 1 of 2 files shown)


test/e2e/diagnostics/testYamlWithoutEE.test.ts
```diff
@@ -25,7 +26,96 @@ export function testDiagnosticsYAMLWithoutEE(): void {
...
+        }
+
+        if (!serverReady) {
+          console.log(
+            "Language server may not be fully ready, continuing anyway...",
+          );
+        }
+
+        console.log("Waiting additional time for full initialization...");
+        await sleep(2000); // Give more time for the extension to be ready
```
alisonlhart:
```suggestion
        if (!serverReady) {
          await sleep(2000); // Give more time for the extension to be ready
          console.log("Waiting additional time for full initialization...");

          console.log(
            "Language server may not be fully ready, continuing anyway...",
          );
        }
```
Let's only do this final sleep if the server still isn't ready, so we're not waiting again if it is ready

test/helper.ts
```diff
@@ -46,12 +50,13 @@ export async function activate(docUri: vscode.Uri): Promise<any> {
...
   } catch (e) {
     console.error("Error from activation -> ", e);
+    throw e;
   }
 }
 
 async function reinitializeAnsibleExtension(): Promise<void> {
   await vscode.languages.setTextDocumentLanguage(doc, "ansible");
-  await sleep(2000); //  Wait for server activation (reduced from 20000 to 2000)
+  await sleep(5000); //  Wait for server activation (increased from 2000 to 5000)
```
alisonlhart: Do we also need this increased wait with the modifications for checking if the server is ready in testYamlWithoutEE?

test/e2e/diagnostics/testYamlWithoutEE.test.ts
anusshukla: Changed the logic to only do the additional sleep if the server still isn't ready, making the test more efficient. Now it only waits when necessary rather than always waiting.

PR #2190, opened 2025-09-25
Specify commit SHAs for all GHAs

anusshukla:
All GitHub Actions using mutable tags have been updated to pinned commit SHAs to prevent supply chain risks. Previously, SonarQube was reporting 50 security hotspots because mutable action tags could be changed or manipulated, creating potential supply chain vulnerabilities. ShellCheck warnings in tools/test-setup.sh have also been fixed, so all SAST checks now pass.

**Outcome:** These changes remove security hotspots, ensure CI/CD compliance, and are fully backward compatible

Fixes: AAP-54393

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -106,7 +106,7 @@ jobs:
               SKIP_PODMAN: 1
               SKIP_DOCKER: 1
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           fetch-depth: 0 # we need tags for dynamic versioning
           show-progress: false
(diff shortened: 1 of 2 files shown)


anusshukla:
> You should not do this yourself. We can configure renovate to do this if we want, especially as the way reconate does it is not the same is your change. (adds comments on the same line as the sha).
> 
> I did not use SHA on actions on purpose to reduce the number of updates from renovate.

Thanks for the feedback, @ssbarnea! You're right—Renovate is better than manual SHA pinning.

Quick questions to confirm the approach:

- SonarQube hotspots: Should Renovate auto-pin digests for GitHub Actions, or do we handle these differently?
- Renovate config: Should I enable pinDigests: true for all GitHub Actions with scheduled updates, or selectively for flagged actions?
- Shell script fix: Keeping the tools/test-setup.sh ShellCheck directive is fine, right?

I can revert the workflow changes and implement Renovate once I know your preference.

.github/workflows/ci.yaml
```diff
@@ -106,15 +106,17 @@ jobs:
               SKIP_PODMAN: 1
               SKIP_DOCKER: 1
     steps:
-      - uses: actions/checkout@v5
+      - # v5.0.0
```
ssbarnea: look on how renovate stores the comment, it must be on the same line as the sha not line before or after. please be sure you use the exact same format: https://github.com/ansible/team-devtools/pull/356/files

PR #2191, opened 2025-09-29
Add Slack notifications for CI failures

anusshukla:
This PR integrates Slack alerts into the GitHub Actions workflow. When a CI job fails on the main branch, a notification is sent to the channel via a secure Slack webhook stored in GitHub Secrets.

image


.github/workflows/ci.yaml
```diff
@@ -558,3 +558,16 @@ jobs:
       - run: npm publish --access public @ansible-ansible-language-server-*.tgz
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+  notify:
```
ssbarnea, line 562: To be able to use secrets from an environment you need to add `environment: notify` and add this secret to that environment. We do not add such secrets to any jobs for security.

PR #2202, opened 2025-10-06
Register MCP server with enable/disable settings support

shatakshiiii:
### SUMMARY

- Registers MCP server using VS Code's native programmatic API.
- Adds extension settings to enable/disable MCP server functionality.
- Implements real-time configuration changes without restart/reload.
- Includes command palette entries for MCP server control.

JIRA - [AAP-53787](https://issues.redhat.com/browse/AAP-53787)
Follow-up of https://github.com/ansible/vscode-ansible/pull/2182

Output when running - **MCP: List Servers**
Screenshot 2025-10-08 at 5 28 55 PM

New entries in the command palette to **Enable and Disable MCP server** 
Screenshot 2025-10-08 at 5 31 20 PM

### DIFF
--- a/test/ui/walkthroughUiTest.ts
+++ b/test/ui/walkthroughUiTest.ts
@@ -109,7 +109,7 @@ describe("Check walkthroughs, elements and associated commands", function () {
     });
   });
 
-  it("Check empty playbook command option", async function () {
+  it.skip("Check empty playbook command option", async function () {
     await workbench.executeCommand("Ansible: Create an empty Ansible playbook");
 
     const newFileEditor = await openUntitledFile();
@@ -124,7 +124,7 @@ describe("Check walkthroughs, elements and associated commands", function () {
     await dialogBox.getDriver().wait(until.stalenessOf(dialogBox), 2000);
   });
 
-  it("Check unauthenticated playbook command option", async function () {
+  it.skip("Check unauthenticated playbook command option", async function () {
     await workbench.executeCommand(
       "Ansible: Create an empty playbook or with Lightspeed (if authenticated)",
     );


test/unit/mcp/mcpSettings.test.ts
```diff
@@ -0,0 +1,123 @@
+require("assert");
```
anusshukla: do we need this as we are importing assert from chai

package.json
```diff
@@ -205,6 +205,10 @@
       {
         "command": "ansible.mcpServer.enabled",
         "title": "Enable Ansible Development Tools MCP Server"
+      },
+      {
+        "command": "ansible.mcpServer.disabled",
```
anusshukla: The command is named disabled but should be disable for consistency with the enabled command.

src/utils/mcpProvider.ts
```diff
@@ -0,0 +1,174 @@
...
+  public refresh(): void {
+    this.didChangeEmitter.fire();
+  }
+
+  /**
+   * Dispose of the provider
+   */
+  public dispose(): void {
+    this.didChangeEmitter.dispose();
+  }
```
anusshukla, line 173: can we also dispose mcpprovider otherwise there will be risk of memory leak

test/unit/mcp/mcpSettings.test.ts
shatakshiiii: yes that was not required, removed it!

src/utils/mcpProvider.ts
shatakshiiii: Done, I added context.subscriptions.push(mcpProvider); on line 685, right after the existing disposable registration. Now when the extension deactivates, VS Code will automatically call both-
1. mcpProviderDisposable.dispose() - to unregister the provider
2. mcpProvider.dispose() - to properly dispose the didChangeEmitter and prevent memory leaks

package.json
shatakshiiii: done

src/extension.ts
```diff
@@ -677,12 +708,11 @@ export async function activate(context: ExtensionContext): Promise<void> {
...
 
-        // Start the MCP server
-        await startMcpServer(extSettings);
+        // Refresh the MCP provider to register the server
+        mcpProvider.refresh();
 
-        // Show success message
         vscode.window.showInformationMessage(
-          "Ansible MCP Server has been enabled and started successfully.",
+          "Ansible MCP Server has been enabled successfully and is now available for AI assistants.",
```
anusshukla, line 715: Image

It look like when the user runs the enable/disable command, the success message appears twice — once from the command handler (around line 713) and again from the configuration change handler (around line 1133).

shatakshiiii: @anusshukla will be addressing that in a follow-up. I have something similar to address, will bunch that together.

PR #2208, opened 2025-10-07
Add kill exits for hanging ui tests (extest)

alisonlhart:
Fixes a strange issue with extester UI tests hanging and exiting with `143` after timeout.

This adds a check for finished UI tests, waits 5 seconds for the test to exit normally, and kills the process if it has not exited. 

It also adds logic to exclude passing tests exited with KILL from the failure list, so the tests don't fail needlessly.

Examples of tests passing but reporting failure from timeout:
https://github.com/ansible/vscode-ansible/actions/runs/18311059203/job/52139507108?pr=2205#step:20:684

### DIFF
--- a/tools/test-launcher.sh
+++ b/tools/test-launcher.sh
@@ -191,13 +191,42 @@ if [[ "${TEST_TYPE}" == "ui" ]]; then
                 start_server
             fi
             refresh_settings "${test_file}" "${TEST_ID}"
-            timeout --preserve-status 120s npm exec -- extest run-tests "${COVERAGE_ARG}" \
-                --mocha_config test/ui/.mocharc.js \
-                -s out/test-resources \
-                -e out/ext \
-                --code_settings out/settings.json \
-                -c "${CODE_VERSION}" \
-                "${test_file}" || echo "${TEST_ID}" >> out/log/.failed
+            timeout --preserve-status 120s bash -c "
+                npm exec -- extest run-tests '${COVERAGE_ARG}' \
+                    --mocha_config test/ui/.mocharc.js \
+                    -s out/test-resources \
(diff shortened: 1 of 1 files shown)


ssbarnea (changes requested):
Have you tried `timeout --help`? All this bash magic can be replaced by `--kill-after=`

ssbarnea (changes requested):
- no link to unresolved extest bug related to hangup
- current version of extest from out project is heavily outdated, not mention of attempt to update it
- current implementation looks too complex and fragile IMHO

ssbarnea (changes requested):
Processing the logs is not very reliable but another way to ensure that extest run finished its job regardless of the test results is to check for the presence of the junit report file that we tell it to produce.

`out/junit/ui/${TEST_ID}-test-results.xml` file should be present and this can be used to decide if we want to call the execution a failure or a pass.

alisonlhart:
@ssbarnea `--kill-after` won't work in this case, because we need to trigger the kill once the test is done, not after an arbitrary time limit. Waiting 120 seconds to trigger a kill after tests completed but are just hanging is too long. Also, `--kill-after` gives a non-zero exit, same as `timeout` by default, which will cause the tests to still report a failure similar to the example I gave.

I also did try upgrading the extester dependency, but that resulted in other unrelated UI failures since there have been many changes since we upgraded it. 

Using the junit file that already exists is a fine approach, we'd just change the check to be in that file instead.

I'd like to get something merged to fix this ASAP, even if the underlying issue is extester, so that we unblock CI. Even if we revert this change later when there's a fix or recommendation, we need a quick workaround to prevent CI from needlessly failing.

PR #2209, opened 2025-10-08
chore: update extest from 8.14.1 to 8.18.0

ssbarnea:
Related: #2208

### DIFF
--- a/test/ui/walkthroughUiTest.ts
+++ b/test/ui/walkthroughUiTest.ts
@@ -19,8 +19,6 @@ const openUntitledFile = async () => {
       try {
         const editorView = new EditorView();
         const titles = await editorView.getOpenEditorTitles();
-
-        // Find any untitled document
         const untitledTitle = titles.find((title) =>
           title.startsWith("Untitled"),
         );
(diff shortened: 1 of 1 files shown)


test/ui/walkthroughUiTest.ts
```diff
@@ -67,19 +67,75 @@ describe("Check walkthroughs, elements and associated commands", function () {
     ],
   ];
 
-  walkthroughs.forEach(([walkthroughName, steps]) => {
+  walkthroughs.forEach(([walkthroughName, steps], index) => {
     it(`Open the ${walkthroughName} walkthrough and check elements`, async function () {
-      const commandInput = await workbench.openCommandPrompt();
+      // First test gets extra time since VS Code walkthrough system needs to initialize
+      const isFirstTest = index === 0;
+      this.timeout(isFirstTest ? 120000 : 90000);
```
alisonlhart: This timeout is too high - if it's taking a long time to initialize, then we should expectedly time out. We can have at most a 1 minutes timeout (10000ms) IMO if we want to see if this helps temporarily.

test/ui/walkthroughUiTest.ts
```diff
@@ -101,15 +157,13 @@ describe("Check walkthroughs, elements and associated commands", function () {
...
       const stepText = fullStepText.split("\n")[0];
       expect(steps, "No walkthrough step").to.include(stepText);
-
-      await editorView.closeEditor("Walkthrough: Ansible");
     });
   });
 
-  it.skip("Check empty playbook command option", async function () {
+  it("Check empty playbook command option", async function () {
+    this.timeout(60000);
```
alisonlhart: Same for these timeouts - too high, 6 minutes is far too long a timeout. If we need a 10s or so timeout, I don't think we should go higher than that.

test/ui/walkthroughUiTest.ts
```diff
@@ -124,7 +178,9 @@ describe("Check walkthroughs, elements and associated commands", function () {
     await dialogBox.getDriver().wait(until.stalenessOf(dialogBox), 2000);
   });
 
-  it.skip("Check unauthenticated playbook command option", async function () {
+  it("Check unauthenticated playbook command option", async function () {
+    this.timeout(60000);
```
alisonlhart: Same as above.

test/ui/walkthroughUiTest.ts
```diff
@@ -67,19 +67,75 @@ describe("Check walkthroughs, elements and associated commands", function () {
...
+              }
+            }
+            return false;
+          } catch (e) {
+            console.log("Error in walkthrough detection:", e);
+            return false;
+          }
         },
-        message: "Timed out waiting for walkthrough tab to open",
+        timeout: isFirstTest ? 75000 : 45000,
```
alisonlhart: This timeout is different than the above, is this intended?

PR #2219, opened 2025-10-22
feat(mcp): add support for ansible-dev-environment

anusshukla:
This PR adds three new ADE tools to help users inspect and manage their environment:

- ade_check_adt — checks if ADT is available
image
- ade_environment_info — shows installed tools, versions, and venv root (venv disabled by default)
image
- ade_setup_environment — explains how to activate/deactivate the virtual environment
image

Related - [AAP-53062](https://issues.redhat.com/browse/AAP-53062)

### DIFF
--- a/packages/ansible-mcp-server/test/handlers.ts
+++ b/packages/ansible-mcp-server/test/handlers.ts
@@ -297,43 +297,4 @@ describe("MCP Handlers", () => {
       expect(result.isError).toBe(true);
     });
   });
-
-  // describe("ansible_lint handler", () => {
-  //   it("should handle file input without options", async () => {
-  //     const handler = createAnsibleLintHandler();
-  //     const result = await handler({ file: "playbook.yml" });
-
-  //     // This will fail in test environment since ansible-lint isn't available
-  //     // but we can test the structure
-  //     expect(result.content).toBeDefined();
-  //     expect(result.content.length).toBeGreaterThan(0);
-  //   });
-
-  //   it("should handle file input with options", async () => {
-  //     const handler = createAnsibleLintHandler();
-  //     const result = await handler({
-  //       file: "playbook.yml",
-  //       extraArgs: ["--skip-list", "no-changed-when"],
-  //     });
-
-  //     expect(result.content).toBeDefined();
-  //     expect(result.content.length).toBeGreaterThan(0);
-  //   });
-
-  //   it("should handle file input with multiple options", async () => {
-  //     const handler = createAnsibleLintHandler();
-  //     const result = await handler({
-  //       file: "playbook.yml",
-  //       extraArgs: [
-  //         "--skip-list",
-  //         "no-changed-when",
-  //         "--exclude",
-  //         "roles/legacy/",
-  //       ],
-  //     });
-
-  //     expect(result.content).toBeDefined();
-  //     expect(result.content.length).toBeGreaterThan(0);
-  //   });
-  // });
 });


shatakshiiii:
IMO the tool `ade_check_adt` that checks and installs ADT should be a separate MCP tool (i.e. not within ADE). It can create a confusion that this ADT check/installation is required for ADE purposes only when the fact is we want to make sure that our environment has all ADT packages.

@alisonlhart What do you think ^ ?

anusshukla:
> IMO the tool `ade_check_adt` that checks and installs ADT should be a separate MCP tool (i.e. not within ADE). It can create a confusion that this ADT check/installation is required for ADE purposes only when the fact is we want to make sure that our environment has all ADT packages.
> 
> @alisonlhart What do you think ^ ?

@shatakshiiii you’re correct that placing ade_check_adt under ADE could give the impression that ADT is only required for ADE.
The reason I named and placed it this way is that ADE tools actually depend on ADT to function properly. Including the check under ADE ensures that users won’t run into failures if ADT is missing.To make this clearer I can also add a note in the tool description/docs stating that this check is specifically for ADE tools, while ADT itself is a broader environment dependency.

alisonlhart:
Looks great @anusshukla ! 
It sounds like we have majority agreement that the ade_check_adt tool should be moved into its own tool instead of being under ADE (via [this thread](https://redhat-internal.slack.com/archives/C01NQV614EA/p1761737450491009)).
But that can be done in a separate ticket since it's out of scope for this feature.

packages/ansible-mcp-server/src/server.ts
```diff
@@ -86,13 +109,159 @@ export function createAnsibleMcpServer(workspaceRoot: string) {
...
+          "Install ansible-dev-tools package",
+          "Verify ADT installation status",
+          "Fix missing ADT installation",
+          "Ensure Ansible development tools are available",
+          "Install missing ansible-lint, ansible-navigator, or ansible-builder",
+        ],
+      },
     },
-    createDebugEnvHandler(workspaceRoot),
+    createADECheckADTHandler(),
```
alisonlhart:
```suggestion
    createADTCheckEnvHandler(),
```

packages/ansible-mcp-server/test/handlers.ts
```diff
@@ -1,10 +1,20 @@
-import { describe, it, expect } from "vitest";
+import { describe, it, expect, vi, beforeEach } from "vitest";
 import {
   createZenOfAnsibleHandler,
-  createDebugEnvHandler,
+  createADEEnvironmentInfoHandler,
+  createADESetupEnvironmentHandler,
+  createADECheckADTHandler,
```
alisonlhart:
```suggestion
  createADTCheckEnvHandler,
```

packages/ansible-mcp-server/test/handlers.ts
```diff
@@ -72,6 +82,222 @@ describe("MCP Handlers", () => {
...
+    it("should check and install ADT successfully", async () => {
+      const { checkAndInstallADT } = await import("../src/tools/adeTools.js");
+
+      vi.mocked(checkAndInstallADT).mockResolvedValue({
+        success: true,
+        output: "ADT is installed",
+        error: undefined,
+      });
+
+      const handler = createADECheckADTHandler();
```
alisonlhart:
```suggestion
      const handler = createADTCheckEnvHandler();
```

packages/ansible-mcp-server/test/handlers.ts
```diff
@@ -72,6 +82,222 @@ describe("MCP Handlers", () => {
...
+    it("should handle ADT check failures", async () => {
+      const { checkAndInstallADT } = await import("../src/tools/adeTools.js");
+
+      vi.mocked(checkAndInstallADT).mockResolvedValue({
+        success: false,
+        output: "Failed to install ADT",
+        error: "Installation failed",
+      });
+
+      const handler = createADECheckADTHandler();
```
alisonlhart:
```suggestion
      const handler = createADTCheckEnvHandler();
```

packages/ansible-mcp-server/test/handlers.ts
```diff
@@ -72,6 +82,222 @@ describe("MCP Handlers", () => {
...
+    });
+
+    it("should handle exceptions during ADT check", async () => {
+      const { checkAndInstallADT } = await import("../src/tools/adeTools.js");
+
+      vi.mocked(checkAndInstallADT).mockRejectedValue(
+        new Error("Check exception"),
+      );
+
+      const handler = createADECheckADTHandler();
```
alisonlhart:
```suggestion
      const handler = createADTCheckEnvHandler();
```

packages/ansible-mcp-server/test/handlers.ts
```diff
@@ -72,6 +82,222 @@ describe("MCP Handlers", () => {
...
+      );
+      expect(result.isError).toBe(true);
+    });
+
+    it("should handle non-Error exceptions during ADT check", async () => {
+      const { checkAndInstallADT } = await import("../src/tools/adeTools.js");
+
+      vi.mocked(checkAndInstallADT).mockRejectedValue("String exception");
+
+      const handler = createADECheckADTHandler();
```
alisonlhart:
```suggestion
      const handler = createADTCheckEnvHandler();
```

packages/ansible-mcp-server/src/handlers.ts
```diff
@@ -102,3 +92,99 @@ export function createListToolsHandler(getToolNames: () => string[]) {
...
+            text: `Error setting up development environment: ${errorMessage}`,
+          },
+        ],
+        isError: true,
+      };
+    }
+  };
+}
+
+export function createADECheckADTHandler() {
```
alisonlhart:
```suggestion
export function createADTCheckEnvHandler() {
```

packages/ansible-mcp-server/src/server.ts
```diff
@@ -10,7 +10,9 @@ import {
   createZenOfAnsibleHandler,
   createListToolsHandler,
   createAnsibleLintHandler,
-  createDebugEnvHandler,
+  createADEEnvironmentInfoHandler,
+  createADESetupEnvironmentHandler,
+  createADECheckADTHandler,
```
alisonlhart:
```suggestion
  createADTCheckEnvHandler,
```

packages/ansible-mcp-server/src/tools/adeTools.ts
```diff
@@ -0,0 +1,601 @@
...
+    }
+  }
+
+  return await executeCommand("pip", args, workspaceRoot);
+}
+
+/**
+ * Clean up conflicting Ansible packages
+ */
+export async function cleanupConflictingPackages(): Promise<ADECommandResult> {
```
alisonlhart: I'm a little hesitant to have this as a function we offer. We may want to just check and suggest solutions if we find older versions.

packages/ansible-mcp-server/src/tools/adeTools.ts
```diff
@@ -0,0 +1,601 @@
...
+    return {
+      success: true,
+      output: "✅ ansible-lint is working properly",
+    };
+  } else {
+    // Try to fix by upgrading ansible-core and reinstalling ansible-lint
+    const results: string[] = [];
+
+    // First, upgrade ansible-core
+    results.push("Upgrading ansible-core...");
```
alisonlhart: same for this one, not sure if we want to take action for upgrading deps rather than making suggestions. Could get very tricky for different types of platforms/package managers.

alisonlhart (commented):
Can we get @cidrblock 's thoughts on this too?

packages/ansible-mcp-server/src/tools/adeTools.ts
```diff
@@ -0,0 +1,601 @@
+import { spawn } from "node:child_process";
+import fs from "node:fs/promises";
+import path from "node:path";
+
```
alisonlhart, line 4: Can we add a description header to this file to match the creator / lint tool files? 
Like: https://github.com/ansible/vscode-ansible/pull/2246/files#diff-096c414ac86de89aee27f9d36e85e63416f499a9b7abe6c4d7bef1f138bd8bc8L4-R11

packages/ansible-mcp-server/src/tools/adeTools.ts
anusshukla: Refactored function to be diagnostic-only

packages/ansible-mcp-server/src/tools/adeTools.ts
anusshukla: Refactored functions to be diagnostic only

shatakshiiii (approved):

PR #2245, opened 2025-10-28
feat(mcp): add create_playbook and create_collection commands

ssbarnea:
Fixes: https://issues.redhat.com/browse/AAP-55770

### DIFF
--- a/packages/ansible-mcp-server/src/server.ts
+++ b/packages/ansible-mcp-server/src/server.ts
@@ -19,6 +19,7 @@ import {
   formatDependencyError,
   type Dependency,
 } from "./dependencyChecker.js";
+import { createInitHandler } from "./tools/creator.js";
 
 export function createAnsibleMcpServer(workspaceRoot: string) {
   const server = new McpServer({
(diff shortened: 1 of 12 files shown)


packages/ansible-mcp-server/src/creator.ts
```diff
@@ -0,0 +1,191 @@
...
+            text: "Ensure 'ansible-creator' is installed and on PATH\n",
+          },
+          {
+            type: "text" as const,
+            text: "Install with: pip install ansible-creator\n",
+          },
+        ],
+        isError: true,
+      };
+    }
```
anusshukla: i think these 2 function almost share the same login can we take out the common changes and make seperate handlee

packages/ansible-mcp-server/src/server.ts
```diff
@@ -85,6 +86,28 @@ export function createAnsibleMcpServer(workspaceRoot: string) {
...
+          .optional()
+          .describe(
+            "Optional path where the playbook should be created. Defaults to current directory.",
+          ),
+      },
+    },
+    createPlaybookHandler(),
+    [], // No dependencies
+  );
+
```
anusshukla, line 251: no registration for collection?🤔

alisonlhart:
+1 on breaking the handler out into its own tool file, and same for tests. It would be good to have the same for the other tools.

alisonlhart:
@shatakshiiii @anusshukla Any concerns with this change going in first and aligning other PRs from there? Would that make sense?

PR #2269, opened 2025-11-04
feat(mcp): Adding the ansible-navigator tool

anusshukla:
Added the ansible-navigator tool

image

related - [AAP-53066](https://issues.redhat.com/browse/AAP-53066)


alisonlhart:
LGTM, looks like a test needs to be updated, though.

PR #2271, opened 2025-11-05
Fixed ci issues and removed preflight from need

anusshukla:
Changes pre-flight to run in parallel since it is causing 5+ min of wait time before other jobs run.


alisonlhart:
CI issues are known and we are addressing them in other PRs: #2270 #2265

PR #2275, opened 2025-11-05
chore(e2e): Add preflight to check job and workflow cancel on failure

alisonlhart:
Draft. 
Todo: skip test suite that already ran in preflight.


anusshukla (approved):

ssbarnea (dismissed):
Don't do this, modify container to build stuff inside it not to install it at run time.

PR #2288, opened 2025-11-11
fix(build): ensure current use of local vsix file during e2e testing

ssbarnea:
Related: AAP-54472


alisonlhart:
@ssbarnea It looks like this still isn't catching the activation issue?

ganeshrn (approved):

PR #2314, opened 2025-11-24
fix: remove LS warning about disabling copilot

ssbarnea:


ganeshrn (approved):
Based on the offline discussion we can remove this warning message

PR #2329, opened 2025-11-28
feat: Google SDK base changes for BYOLLM [ACA-4539]

PyContriver:
Introduce foundational updates to the Google SDK integration to support BYOLLM (Bring Your Own Large Language Model). These changes prepare the codebase for greater model flexibility and improved compatibility with custom LLMs built on Google’s ecosystem.This update establishes the groundwork for future BYOLLM capabilities but does not include the full implementation. Additional configuration modules, model-handling logic, and runtime components will be introduced in subsequent PRs.

Manual Test Report: https://docs.google.com/document/d/1gu73G_wuRPC2xqN-0l038lXbxO-_Hf4aRy9p5sk4284/edit?usp=sharing

Jira Ref: [ACA-4539](https://issues.redhat.com/browse/ACA-4539)
Fixes: ACA-4539

PR #2342, opened 2025-12-03
feat: add BYOLLM webviews

kaushiki069:
Add the further configuration components and webview related changes to support BYOLLM (Bring Your Own Large Language Model) for google provider. It also entails addition of telemetry changes, unit tests and UI tests around the new changes being proposed, updates to the existing UI tests owing to the changes in the webviews and vsix file limit increase.

JIRA ref: [ACA-4540](https://issues.redhat.com/browse/ACA-4540), [ACA-4561](https://issues.redhat.com/browse/ACA-4561), [ACA-4543](https://issues.redhat.com/browse/ACA-4543), [ACA-4531](https://issues.redhat.com/browse/ACA-4531)
Manual test report: [Document link](https://docs.google.com/document/d/1QIXm8OoFpnwaPwYGKA2vQ2g8wjs2smH34eHa-TSvxdI/edit?usp=sharing)
Related: https://github.com/ansible/vscode-ansible/pull/2329

### DIFF
--- a/src/features/lightspeed/api.ts
+++ b/src/features/lightspeed/api.ts
@@ -448,19 +448,13 @@ export class LightSpeedAPI {
         connected: true,
         modelInfo: {
           name: "WCA",
-          capabilities: [
-            "completion",
-            "chat",
-            "generation",
-            "contentmatching",
-          ],
+          capabilities: ["completion", "chat", "generation", "contentmatching"],
         },
       };
     } catch (error) {
       return {
         connected: false,
-        error:
-          error instanceof Error ? error.message : "WCA connection failed",
+        error: error instanceof Error ? error.message : "WCA connection failed",
       };
     }
   }


src/features/lightspeed/commands/providerCommands.ts
```diff
@@ -0,0 +1,292 @@
...
+        label: provider.displayName,
+        description: provider.description,
+        detail: provider.type,
+        provider: provider,
+      }));
+
+      const selectedProvider = await vscode.window.showQuickPick(
+        providerItems,
+        {
+          placeHolder: "Select a LLM LLM provider",
```
anusshukla: Select a LLM provider

src/extension.ts
```diff
@@ -818,6 +902,34 @@ export async function activate(context: ExtensionContext): Promise<void> {
...
+                provider: provider,
+                ...param, // Include all original fields
+                model:
+                  extSettings.settings.lightSpeedService.modelName || undefined,
+              },
+            );
+          } catch (error) {
+            console.log(
+              `[Lightspeed Role Feedback] Telemetry not sent: ${error}`,
+            );
```
anusshukla: can we show the error 
console.error(`[Lightspeed Feedback] Telemetry failed: ${error}`, error);

tanwigeetika1618: Yes we can. Done!

package.json
```diff
@@ -415,39 +431,73 @@
...
+            "order": 2,
+            "scope": "resource",
+            "type": "string"
+          },
+          "ansible.lightspeed.provider": {
+            "default": "wca",
+            "enum": [
+              "wca",
+              "google"
+            ],
```
anusshukla, line 471: are we supporting only wca and google but in documentation it is written that we are supporting (`wca`, `openai`, `google`, `custom`)

kaushiki069: Thank you for pointing it out @anusshukla . Yes currently we only support wca and google. The documentation has beein updated in this PR to reflect the same.

package.json
```diff
@@ -415,39 +431,73 @@
       },
       {
         "properties": {
+          "ansible.lightspeed.apiEndpoint": {
```
alisonlhart, line 434: Is there reasoning why this was changed from URL setting?

src/extension.ts
```diff
@@ -377,6 +388,42 @@ export async function activate(context: ExtensionContext): Promise<void> {
...
+        const config = workspace.getConfiguration("ansible.lightspeed");
+        const provider = config.get<string>("provider");
+
+        // Determine the configuration target (workspace takes precedence over global)
+        const providerInspect = config.inspect<string>("provider");
+        let configTarget: vscode.ConfigurationTarget | undefined;
+
+        if (providerInspect?.workspaceValue !== undefined) {
+          configTarget = vscode.ConfigurationTarget.Workspace;
+        } else if (providerInspect?.workspaceFolderValue !== undefined) {
```
ganeshrn: folder setting takes precedence over workspace right? If that's the case this check should come first.

test/ui/lightspeedUiTestPlaybookGenWebviewPart2Test.ts
```diff
@@ -71,6 +71,7 @@ describe("playbook generation features work", function () {
         );
       },
       message: "Timed out waiting for Continue button",
+      timeout: 30000,
```
ssbarnea, line 74: Why we need such a huge value here? This seems very bad for test execution performance.

tools/test-launcher.sh
```diff
@@ -72,6 +76,23 @@ function start_server() {
     export TEST_LIGHTSPEED_URL
 }
 
+function start_llm_provider_server() {
```
ssbarnea: This is no longer an acceptable way to start a mock server, especially as we already have WIP to replace the LS mock server start from the bash launcher and use mocha hooks to start it.

Please avoid adding new logic to test-launcher.sh and make use of mocha hooks to prepare  stuff like side-loaded servers, changing settings or cleaning up at the end of the tests.

src/settings.ts
```diff
@@ -44,10 +44,22 @@ export class SettingsManager {
       lightSpeedService: {
         enabled: lightSpeedSettings.get("enabled", true),
         provider: lightSpeedSettings.get("provider", "wca"),
-        URL: lightSpeedSettings.get("URL", "https://c.ai.ansible.redhat.com"),
-        apiEndpoint: lightSpeedSettings.get("apiEndpoint", ""),
-        modelName: lightSpeedSettings.get("modelName", undefined),
-        model: lightSpeedSettings.get("modelIdOverride", undefined),
+        apiEndpoint: (() => {
```
alisonlhart, line 47: Same question as previous.

ssbarnea (changes requested):

src/features/lightspeed/commands/providerCommands.ts
```diff
@@ -0,0 +1,292 @@
...
+
+      // Set default endpoint if available and not already set
+      if (
+        selectedProvider.provider.defaultEndpoint &&
+        !lightspeedConfig.get("apiEndpoint")
+      ) {
+        await lightspeedConfig.update(
+          "apiEndpoint",
+          selectedProvider.provider.defaultEndpoint,
+          vscode.ConfigurationTarget.Workspace,
```
ganeshrn: This updates the workspace config but what if the config it for folder or global settings. Should a check to inspect config type should be added here `const providerInspect = config.inspect<string>("provider");`

package.json
kaushiki069: @alisonlhart with addition of google provider (and further more such providers to come in the future), according to us apiEndpoint seems more semantically aligned and self-explanatory. It improves the readability and communicates the purpose of the field. While the term 'URL' can be generic in nature, the term 'apiEndpoint' reduces ambiguity by specifying the intent clearly. Please let me know your thoughts on this!

src/settings.ts
kaushiki069: This has been answered in the above [coversation](https://github.com/ansible/vscode-ansible/pull/2342#discussion_r2619862744). Please let us know your thoughts on the same.

src/features/lightspeed/commands/providerCommands.ts
kaushiki069: @ganeshrn This has been addressed in the [commit](https://github.com/ansible/vscode-ansible/pull/2342/commits/f6ab3db1e6acde0648187f15965b46c12c01dd40). Please let us know your thoughts on it.

src/extension.ts
kaushiki069: @ganeshrn This has been addressed in the [commit](https://github.com/ansible/vscode-ansible/pull/2342/commits/f6ab3db1e6acde0648187f15965b46c12c01dd40).  Please let us know your thoughts on it.

test/ui/lightspeedUiTestPlaybookGenWebviewPart2Test.ts
kaushiki069: @ssbarnea actually this was causing the tests to fail due to timeout issues. Hence, we had to resort to a 30 second timeout here. Please let us know your thoughts on this!

package.json
alisonlhart: The concern is mostly around backwards compatibility for the setting, totally agree with your reasoning though. I think it's better to go ahead and rename now as you have rather than carrying the old setting, but just wanted to verify.

tools/test-launcher.sh
tanwigeetika1618: Hey @ssbarnea  Have addressed your comment here in this [commit](https://github.com/ansible/vscode-ansible/pull/2342/commits/7f0e82bf5af34ff1e837e0c3649deccab9c42e19)! 
Refactored to use Mocha hooks:

- Created `test/ui/mockLightspeedLLMProviderServer/serverManager.ts` to manage server lifecycle
- Tests now start/stop the mock server in `before()`/`after()` hooks
- Removed `start_llm_provider_server()`, `stop_llm_provider_server()`, and all `MOCK_LLM_PROVIDER_API` logic from `test-launcher.sh`

The LLM provider tests are now self-contained, following the same pattern that must be worked upon for the LS mock server.

test/ui/lightspeedUiTestPlaybookGenWebviewPart2Test.ts
tanwigeetika1618: You’re right, @ssbarnea ..  a 30s timeout is definitely on the higher side and not ideal for test performance.

As @kaushiki069  mentioned, the higher timeout was initially needed to avoid CI-related flakiness. Building on that, I tried reducing the timeout to 15 seconds specifically for the LLM provider tests. Since these tests use a local mock server with instant responses (no real API calls), 15s should be sufficient while still keeping some buffer for CI slowness.

This way, we retain test stability while also improving overall execution performance. Happy to adjust further if you feel a lower value would be safe.

tools/test-launcher.sh
tanwigeetika1618: Please lemme know your thoughts on this!

src/features/lightspeed/lightspeedUser.ts
```diff
@@ -87,13 +87,16 @@ export class LightspeedUser {
...
+    const provider = this._settingsManager.settings.lightSpeedService.provider;
     const lightspeedUri = getBaseUri(this._settingsManager);
     this._logger.info(
-      `[ansible-lightspeed-user] Initializing LightspeedUser for Lightspeed URI ${lightspeedUri} and extension host ${this._extensionHost}`,
-    );
-    this._logger.info(
-      `[ansible-lightspeed-user] Red Hat authentication extension is${vscode.extensions.getExtension("redhat.vscode-redhat-account") ? "" : " not"} installed.`,
+      `[ansible-lightspeed-user] Initializing LightspeedUser with provider: ${provider}, URI: ${lightspeedUri || "(none)"}, extension host: ${this._extensionHost}`,
     );
+    if (provider === "wca") {
```
ganeshrn: This if condition log is not required as lighspeed auth login doesn't rely on Red Hat authentication extension.

src/features/lightspeed/providerManager.ts
```diff
@@ -0,0 +1,327 @@
...
+  /**
+   * Test provider connection
+   */
+  async testProviderConnection(
+    providerType: "wca" | "llmprovider",
+  ): Promise<ProviderStatus> {
+    if (providerType === "wca") {
+      try {
+        // Test WCA connection by making a simple API call
+        const testParams = {
```
ganeshrn: This can be moved to LightSpeedAPI class by adding getStatus method

src/features/lightspeed/providers/factory.ts
```diff
@@ -30,14 +30,20 @@ export class LLMProviderFactory implements ProviderFactory {
...
         if (!config.apiKey) {
           throw new Error(
             "API Key is required for Google Gemini. Please set 'ansible.lightspeed.apiKey' in your settings.",
           );
         }
-        // Google doesn't support custom endpoints - uses SDK
-        if (config.apiEndpoint && config.apiEndpoint !== GOOGLE_API_ENDPOINT) {
+        // Allow localhost URLs for testing, block other custom endpoints
+        const isLocalhostUrl =
+          config.apiEndpoint?.startsWith("http://localhost");
```
ganeshrn, line 41: Why is this localhost handling required? Also do we need to check the value of GOOGLE_API_ENDPOINT as the value is handled by the google SDK? In the apiEndpoint setting description can we add a description mentioning this value is applicable only for WCA provider for now.

kaushiki069: "Why is this localhost handling required?"
@ganeshrn w.r.t. the above lines, this has been added for UI test mocking. This helps in routing the request to the LLM provider mock server when the UI tests run.
cc: @tanwigeetika1618 
_"Also do we need to check the value of GOOGLE_API_ENDPOINT as the value is handled by the google SDK? In the apiEndpoint setting description can we add a description mentioning this value is applicable only for WCA provider for now."_
@ganeshrn w.r.t. the above lines, we want to ensure handling edge cases in case of custom endpoints , also we think this will help ensure maintainability and consistency for llm providers which would be added in the future. We have added "For Google, the endpoint is automatically set and cannot be changed." to the apiEndpoint setting description in package.json

src/features/lightspeed/providerManager.ts
kaushiki069: @ganeshrn This has been addressed here [commit](https://github.com/ansible/vscode-ansible/pull/2342/commits/d80296c215d5379c31510494dd4e17dde0d0cb4a). Please let us know your thoughts on the same.

src/features/lightspeed/lightspeedUser.ts
kaushiki069: @ganeshrn This has been addressed here [commit](https://github.com/ansible/vscode-ansible/pull/2342/commits/d80296c215d5379c31510494dd4e17dde0d0cb4a). Please let us know your thoughts on the same.

src/features/lightspeed/providers/factory.ts
PyContriver: @ganeshrn Whatever @kaushiki069 stated above is correct, in cases where user tries adding a custom endpoint the extension should be able to flag the issue.

PR #2345, opened 2025-12-03
Fix: Pin Zod version to match MCP SDK dependency

shatakshiiii:
### Problem
After renovatebot upgraded Zod to version 4.x in [PR](https://github.com/ansible/vscode-ansible/pull/2294), MCP server tools stopped working correctly. The tools were failing to receive and parse input arguments properly.

### Root Cause Analysis
The issue stems from a **version incompatibility** between Zod and the MCP SDK:

1. **Renovatebot upgraded Zod** to version 4.x (which introduced breaking API changes)
2. **MCP SDK dependency**: The `@modelcontextprotocol/sdk@^1.17.3` package uses Zod `3.23.8` as a peer dependency

### Solution
Pin Zod to version `3.23.8` to match the MCP SDK's dependency version using npm `overrides`. This ensures:

- Compatibility with the MCP SDK's internal Zod usage
- Consistent Zod version across the entire dependency tree

### Future Considerations (IMP)
We should align our Zod version with the MCP SDK's dependency. When the MCP SDK upgrades its Zod dependency, we can then upgrade ours accordingly. Consider adding a renovatebot rule to prevent automatic major version upgrades for Zod until the MCP SDK supports it.

### DIFF
--- a/packages/ansible-mcp-server/package.json
+++ b/packages/ansible-mcp-server/package.json
@@ -5,7 +5,7 @@
     "ajv": "^8.17.1",
     "ajv-formats": "^3.0.1",
     "yaml": "^2.8.1",
-    "zod": "^4.1.13"
+    "zod": "^3.23.8"
   },
   "description": "Ansible Development Tools MCP server with linting, workspace access, and expert prompts",
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


alisonlhart:
Definitely think we should add the renovate rule!

PR #2363, opened 2025-12-08
fix(mcp): Resolve server activation issues in packaged extension

shatakshiiii:
### Summary -

- The changes in mcpProvider.ts and extension.ts deals with activation issue.
- Helper and webpack.config.ts changes were required due to `ERR_MODULE_NOT_FOUND` error that occurred because the MCP server was compiled with tsc but dependencies weren't bundled, so the runtime couldn't find modules.
- Now we bundle the MCP server and its dependencies with Webpack (same approach as the Language Server).
- This approach keeps the vsix size around our limit as we do not rely on node_modules at runtime
#### JIRA: AAP-59320

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ xunit.xml
 !.yarn/releases
 !.yarn/sdks
 !.yarn/versions
+resources/contentCreator/createDevcontainer/.devcontainer
+resources/contentCreator/createDevfile/devfile-template.txt
--- a/uv.lock
+++ b/uv.lock
@@ -26,8 +26,8 @@ name = "ansible-compat"
 version = "25.12.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "ansible-core", version = "2.19.4", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.12'" },
-    { name = "ansible-core", version = "2.20.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.12'" },
+    { name = "ansible-core", version = "2.19.5", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.12'" },
+    { name = "ansible-core", version = "2.20.1", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.12'" },
     { name = "jsonschema" },
     { name = "packaging" },
     { name = "pyyaml" },
@@ -40,7 +40,7 @@ wheels = [
 
 [[package]]
 name = "ansible-core"
-version = "2.19.4"
+version = "2.19.5"
 source = { registry = "https://pypi.org/simple" }
 resolution-markers = [
     "python_full_version < '3.12'",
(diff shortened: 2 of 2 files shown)


src/utils/mcpProvider.ts
```diff
@@ -5,13 +5,21 @@ import * as fs from "fs";
 export class AnsibleMcpServerProvider {
   private static readonly MCP_SERVER_NAME =
     "Ansible Developer Tools MCP Server";
-  private static readonly CLI_PATH =
+  // Development path (relative to project root)
+  private static readonly CLI_PATH_DEV =
     "packages/ansible-mcp-server/out/server/src/cli.js";
+  // Packaged extension path (relative to extension path)
+  private static readonly CLI_PATH_PACKAGED = "out/mcp/cli.js";
```
alisonlhart: Can remove comments, IMO var name is self-explanatory

src/utils/mcpProvider.ts
```diff
@@ -144,6 +149,39 @@ export class AnsibleMcpServerProvider {
...
+        console.log(`Found MCP server CLI at development path: ${devPath}`);
+        return devPath;
+      }
+    }
+
+    console.error(
+      `MCP server CLI not found at either location:\n  - Packaged: ${packagedPath}\n  - Development: ${devPath || "N/A"}`,
+    );
+    return undefined;
+  }
```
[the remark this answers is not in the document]
alisonlhart: Definitely needs a test to ensure usage of both paths are being tested in CI.

shatakshiiii:
> By reading the title of the PR I assume that currently the mcp server in extension is broken as it fails to instantiate correctly.
> 
> Considering that this change does not touch any tests, I would assume that is fails to include any regression measures.
> 
> IMHO, any bugfix must start with a regression tests, even before we have the fix, we need reproducers or we endup breaking two other places when we try to patch one.

I am going to add tests for these changes (that's why PR is in draft)

tools/helper
```diff
@@ -142,8 +142,12 @@ def cli() -> None:
         # --no-dependencies and --no-yarn needed due to https://github.com/microsoft/vscode-vsce/issues/439
         run("yarn run vite-build")
         # Build and stage the MCP server into out/mcp so it is included in the VSIX
+        # Webpack bundles the MCP server code into out/mcp/cli.js
+        # We only need to copy the resource data files that are loaded at runtime
+        # The bundled code uses __dirname which will be out/mcp/, so resources should be at out/mcp/data/
         run("task build")
-        run("mkdir -p out/mcp && cp -r packages/ansible-mcp-server/out/server/src/* out/mcp/")
+        run("mkdir -p out/mcp/data")
+        run("cp -r packages/ansible-mcp-server/src/resources/data/* out/mcp/data/ 2>/dev/null || true")
```
ssbarnea, line 150: We should never have to copy files during the build process, it is a bad practice. I am not asking you do it in this change but see if you can ditch this operation. We should build stuff directly to their target location, no copy.

shatakshiiii: I will check if there is a better way

PR #2382, opened 2025-12-12
fix: disable thumbsUpDown button When Telemetry Is Disabled

Abhishekism9450:
Jira: [ACA-4531](https://issues.redhat.com/browse/ACA-4531)
Title: Display a Pop-Up Warning When Telemetry Is Disabled
Description: When telemetry is disabled for LLM providers (non-WCA), the thumbs-up/down feedback buttons are disabled with a hover tooltip explaining how to enable telemetry. WCA provider feedback always works since it uses the Lightspeed API.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -847,18 +847,6 @@ export async function activate(context: ExtensionContext): Promise<void> {
         const provider = extSettings.settings.lightSpeedService.provider;
         // For LLM providers, send telemetry via Segment instead of WCA API
         if (provider && provider !== "wca") {
-          // Check if telemetry is enabled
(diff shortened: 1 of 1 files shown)


ganeshrn:
> TBH i find the need for this popup a design flaw. If telemetry is disabled we should not show the buttons in the first place or make them disabled. We could also use the tooltip on these buttons to explain why they might be disabled.

The pop-up will be shown only  if user is trying to send feedback on the webviews and they have explicitly turned off the telemetry setting. If we don't provide any sufficiently visible feedback user will end up thinking that they have provided the feedback successfully.

ssbarnea (approved):

PR #2386, opened 2025-12-15
chore: use node 25 during build and testing

ssbarnea:
Use node 25 as this adds --experimental-transform-types which would
allow us to run tests files without transpile.


shatakshiiii:
codecov failure seems unrelated

PR #2393, opened 2025-12-17
fix: ade_devloper_env tool fix

anusshukla:
Related: AAP-60688

The ade_setup_environment tool was returning empty output when
Python was not available:

1. The tool tried to create a venv with an unavailable Python version
2. `python3.11 -m venv` failed with ENOENT
3. The error was not captured or reported
4. The user saw an empty output box with no explanation

### DIFF
--- a/packages/ansible-mcp-server/src/handlers.ts
+++ b/packages/ansible-mcp-server/src/handlers.ts
@@ -33,9 +33,9 @@ export function createZenOfAnsibleHandler() {
 }
 
 export function createAgentsGuidelinesHandler() {
-  return async () => {
+  return async (args?: { topic?: string }) => {
     try {
-      const guidelines = await getAgentsGuidelines();
+      const guidelines = await getAgentsGuidelines(args?.topic);
       return {
         content: [
           {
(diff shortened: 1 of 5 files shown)


ssbarnea (changes requested):
One change needed: please change description to ensure that jira recognize this PR and displays it in the UI. Once this happens I will approve the PR.

Do not manually link from JIRA side, just ensure that PR description and commit message use required syntax for JIRA to recognize it. Usually it takes less than 5min from when you make the change till jira will show it.

ssbarnea (commented):
Coverage is going down, we are almost done :D

packages/ansible-mcp-server/test/tools/adeIntegration.test.ts
```diff
@@ -55,7 +55,7 @@ describe("ADE Tools Integration", () => {
       expect(result).toBeDefined();
       expect(result.content).toBeDefined();
       expect(Array.isArray(result.content)).toBe(true);
-    });
+    }, 45000); // 45 second timeout
```
alisonlhart: Do we need this - and if we do, can we test this in another way?

alisonlhart (approved):

Issue #900, opened 2023-05-18
`redhat.ansible` extension must not modify the workspace settings (with `ansible.python.interpreterPath`).


deadnews:
### Summary

`redhat.ansible` extension must not modify the workspace settings (with `ansible.python.interpreterPath`).

Example: https://github.com/DeadNews/deadnews-template-ansible/pull/28/commits/8b537fbb3cb20b2e58f7061fa59026f371c83a6a

`${workspaceFolder}/.vscode/settings.json` сan be used on different devices by different people. This config is not suitable for mandatory storage of `ansible.python.interpreterPath` setting. Moreover, `redhat.ansible` changes the workspace settings whenever the python interpreter changes. Even if this is in a regular python repository and has nothing to do with ansible.

`ms-python.python` does not need to modify the workspace settings (`${workspaceFolder}/.vscode/settings.json`) to select and remember an interpreter for the current workspace.

Refs: https://github.com/ansible/vscode-ansible/issues/775, https://github.com/ansible/vscode-ansible/discussions/898

### Extension version

v2.1

### VS Code version

1.77

### Ansible Version

```shell
ansible [core 2.15.0]
  config file = /home/user/git/deadnews-template-ansible/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/.cache/pypoetry/virtualenvs/deadnews-template-ansible-EQAYLr04-py3.10/lib/python3.10/site-packages/ansible
  ansible collection location = /home/user/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/user/.cache/pypoetry/virtualenvs/deadnews-template-ansible-EQAYLr04-py3.10/bin/ansible
  python version = 3.10.10 (main, Mar  5 2023, 22:26:53) [GCC 12.2.1 20230201] (/home/user/.cache/pypoetry/virtualenvs/deadnews-template-ansible-EQAYLr04-py3.10/bin/python)
  jinja version = 3.1.2
  libyaml = True
```

### OS / Environment

Linux

cidrblock:
@DeadNews TY for this, and the quick feedback.

We'll figure out what can be done here.....................

pluto00987:
This is a direct result of #871 and is problematic indeed

To elaborate on what OP stated, it's doing two wrong things IMO:

[...]

PR #2411, opened 2026-01-05
fix: respect relative interpreter path settings

anusshukla:
### Summary
Prevents the extension from overwriting user-configured relative or variable-based Python interpreter paths with absolute paths.

### Changes
- Detect when users have explicitly set `ansible.python.interpreterPath`
- Skip auto-detection if a user-defined path already exists
- Show a choice dialog when manually changing interpreter if a relative path is configured
image

### Before
User sets `${workspaceFolder}/.venv/bin/python` → Extension overwrites to `/home/user/project/.venv/bin/python`

### After  
User sets `${workspaceFolder}/.venv/bin/python` → Setting is preserved

Fixes: #900


ssbarnea (approved):

alisonlhart:
Overriding minor codecov decrease.

PR #2419, opened 2026-01-08
fix: remove unused exports and dead code

ssbarnea:
Removes all unused code reported by knip and eslint and reconfigures these to prevent accidental introducing of code that is not used.

Related: #2418


src/definitions/constants.ts
```diff
@@ -48,8 +48,6 @@ export const ANSIBLE_CREATOR_VERSION_MIN = "24.10.1";
 
 export const ANSIBLE_CREATOR_COLLECTION_VERSION_MIN = "24.7.1";
 
-export const ANSIBLE_CREATOR_EE_VERSION_MIN = "24.12.1";
```
alisonlhart, line 51: This is a bug that it isn't being used. If we remove this, we need an issue opened to track it and remember to readd and fix in the webview.

ssbarnea: Added https://issues.redhat.com/browse/AAP-62256

PR #2424, opened 2026-01-09
fix: remove use of lodash library

ssbarnea:
This dependency did not receive any update in 5 years and it did lost most of its original purpose.


alisonlhart:
Similar thoughts to the copilot feedback - would need more time to go through and individually evaluate to see if the functions are 1-1 with those from lodash, but seems like Copilot has caught some.

PR #2428, opened 2026-01-12
fix: ADT command execution in non-ansible workspace folder

shatakshiiii:
Related: AAP-60914

### Fix: ADT installation executes in wrong directory with multiple workspace folders

**Problem** -  When VSCode has multiple workspace folders (e.g., one Ansible repository and multiple Kubernetes repositories), the adt_check_env tool executed commands in the wrong directory instead of the Ansible workspace.

**Root Cause** - The checkADTInstalled() and checkAndInstallADT() functions were not receiving the workspaceRoot parameter, so they defaulted to using process.cwd(), which could be any of the workspace folders, not necessarily the Ansible one.

**Solution** - Updated the code to pass workspaceRoot through the call chain so all commands execute in the correct workspace directory.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -232,6 +232,7 @@ tasks:
     generates:
       - out/junit/ui/.passed
     cmds:
+      - rm -f out/junit/ui/${TEST_PREFIX}-*-test-results.xml
       - '{{.XVFB}}./tools/test-launcher.sh {{.CLI_ARGS}}'
     interactive: true
   ui-smoke:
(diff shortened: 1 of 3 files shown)


alisonlhart:
Can we add a test for the multiple-workspaces scenario? Or if that's not workable with our tests, having a screenshot of this scenario passing locally in the description would be great.

alisonlhart:
Also, I think the best way to get the PR to auto-link to Jira is by adding the ticket ID in the comment. I've been doing it like this: "commit-message (AAP-XXX)" and it links immediately

alisonlhart (approved):

PR #2438, opened 2026-01-14
fix: implement selenium based UI testing

goneri:
Remove the extester tests for Lightspeed, replace them with lightweight Selenium based tests, and unit-tests (https://github.com/ansible/vscode-ansible/pull/2450).

Related: https://github.com/ansible/vscode-ansible/issues/2461


Taskfile.yml
```diff
@@ -259,6 +259,17 @@ tasks:
       TEST_PREFIX: "ui-oldest"
     deps:
       - ui
+  ui-selenium:
+    desc: Run UI tests (Selenium)
+    deps:
+      - setup
+      - build
+      # - package
```
goneri: we need the package, and since package already depends on build and setup, we can keep it.

PR #2505, opened 2026-01-28
chore: consolidate test config into a single vitest.config.ts

ssbarnea:
- Implements vitest workspace configuration with named projects ("ext", "vue", "mcp", "als")
- Updates vitest and related packages to version 4.0.18
- Adds biome as a dev dependency and configures it for local pre-commit hooks
- Updates test names for better clarity and adds validation for vitest list command
- From now on, `task unit` will run all 4 unit test projects and produce just one report.
- Coverage requirements should bump themselves when needed
- Ensures we check the coverage level
- Temporary make use of coverage pragma comments to hide differences in coverage between platforms
- Auto-bumps coverage level when it does increase


packages/ansible-language-server/src/services/ansibleConfig.ts
```diff
@@ -90,6 +90,7 @@ export class AnsibleConfig {
         ...parsePythonStringArray(pythonPathResult.stdout),
       );
     } catch (error) {
+      /* v8 ignore next 3 */
```
alisonlhart, line 93: What's the need for so many inline ignores? Is it because Selenium tests aren't reporting coverage so these appear uncovered?

package.json
```diff
@@ -1078,8 +1078,8 @@
     "@types/yargs": "^17.0.35",
     "@typescript-eslint/eslint-plugin": "^8.53.1",
     "@typescript-eslint/parser": "^8.53.1",
-    "@vitest/coverage-v8": "^4.0.17",
-    "@vitest/ui": "^4.0.17",
+    "@vitest/coverage-v8": "^v4.1.0-beta.2",
```
alisonlhart, line 1081: Using beta packages - can we align on 4.0.18 or is there some beta functionality we need?

ssbarnea: there is one bug fix i needed that was not in release. we will get rid of befa quite soon.

packages/ansible-language-server/src/services/ansibleConfig.ts
ssbarnea: is different issue which is not yet sorted but I will address later.

PR #2525, opened 2026-02-02
chore: selenium: use the settings.json

goneri:
Rely on a static `settings.json` file.

### DIFF
--- a/.config/dictionary.txt
+++ b/.config/dictionary.txt
@@ -63,9 +63,11 @@ equalto
 eslintcache
 extest
 extester
+fixturenames
 fqcn
 funcargs
 genai
+getfixturevalue
 haaaad
 highlightjs
 hookwrapper
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@
 .idea
 .nyc_output
 .pytest_cache
+*.pyc
+*.pyo
 .task
 .vscode-test
 CHANGELOG.html
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,6 +18,7 @@ dev = [
   "mypy>=1.17.1",
   "pip>=22.3.1",
   "prek>=0.3.1",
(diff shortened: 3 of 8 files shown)


ssbarnea (commented):
Do this instead: 

      - VSX_FILE=${PWD}/$(ls -1 ansible-*.vsix) && podman run --pull=missing --replace --detach --rm --shm-size=10g --name selenium-vscode -p 4444:4444 -p 5999:5999 --volume $VSX_FILE:/ansible-latest.vsix quay.io/tshinhar/selenium-vscode-multi:latest
      # settings.json gets changed during tests so we do copy it
      # we use same location for altered settings.json for all test suites
      - mkdir -p out/test-resources/settings/User/ out/ui-selenium/logs out/ui-selenium/coder-logs
      - cp -f test/testFixtures/settings.json out/test-resources/settings/User/settings.json
      - >
        VSX_FILE=${PWD}/$(ls -1 ansible-*.vsix) &&
        podman run
        --pull=missing
        --replace
        --detach
        --rm
        --shm-size=10g
        --name selenium-vscode
        -p 4444:4444
        -p 5999:5999
        --volume $VSX_FILE:/ansible-latest.vsix:ro
        --volume $PWD/out/test-resources/settings/User/settings.json:/home/selenium/.local/share/code-server/User/settings.json:rw
        --volume $PWD/out/ui-selenium/logs:/home/selenium/.local/share/code-server/logs:rw
        --volume $PWD/out/ui-selenium/coder-logs:/home/selenium/.loc



Basically we should have only one root `settings.json` file used for testing that we copy to the same location each time. For selenium we would mount the file from this location.

I hope that selenium tests settings.are not incompatible with our default settings. We can always reconfigure later with pytest but it would be a good idea to have the same 'reference' default settings.json to start with, especially as we might want to either migrate tests from one framework to another or just test behavior in another one.

PR #2563, opened 2026-02-11
feat: integrate with Python Environments extension

anusshukla:
This PR addresses issue #900 where the extension was automatically modifying ansible.python.interpreterPath to absolute paths, breaking portability for teams sharing .vscode/settings.json.
New Architecture:
PythonEnvironmentService: Singleton service that wraps ms-python.vscode-python-envs API for environment retrieval and selection
TerminalService: Creates terminals with automatic Python environment activation using the Python Environments extension's createTerminal() method
Prompts users to enable python.useEnvironmentsExtension setting if the API is not

image


ssbarnea (changes requested):
I added some fixes to your change like exposing the required dependency so users will know to install it (and also contributors.

Still, Sonar reports that your code does not provide the require level of 80% coverage for it to be merged and few other things. 

Tests are passing, now is only about adding some more unit tests.

ssbarnea (changes requested):
Please keep a change in draft until all CI jobs are passing, including Sonar quality gate. This change still reports not meeting coverage requirements and other problems with the code on sonar.

ssbarnea (approved):

Zocker1999NET:
Sorry for commenting on an older PR. I want to ask an honest question to understand your design decision: Why was the Ansible extension integrated with the "Python Environments" extension and not just integrated with the environment selector?

Because as far as I understand the "Python Environments" extension, it is there for managing environments, not for switching between those, as this seems to be implemented somewhere else.

And when reading through #900, I expected this extension to be "simply" integrated to respect that selector, which, I hope, should be possible without a hard requirement on "Python Environments".

And I can confirm that the Python environment selection is either implemented in the `ms-python` extension or in VSCode itself as I have that selector without having "Python Environments" installed.

For context: I'm a user of this Ansible extension, but I do not (want to) use the "[Python Environments](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-python-envs)" extension because I prefer to manage my Python environments either via nix devShells or by hand using a requirements.txt.

PR #2573, opened 2026-02-12
feat: refactor LLM explorer side panel

Abhishekism9450:
This PR migrates the existing LLM explorer side panel view into a new webview page. 
This removes fetching existing configuration management for LLM provider from setting.json. 
It adds the support to store the provider configuration into VSCode globalState and secrets for persistance . 

Jira: [ACA-4979](https://issues.redhat.com/browse/ACA-4979)
ANSTRAT: [ANSTRAT-1828](https://issues.redhat.com/browse/ANSTRAT-1828)
Figma:  [UX](https://www.figma.com/proto/Uzf3aMNZ4SCH4FvDxSvj0Y/VSCode-Extention?page-id=6130%3A751&node-id=6482-4156&viewport=225%2C96%2C0.06&t=nWRAanH2xdkIqfyp-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=6130%3A1203)
Depends-On: https://github.com/ansible/vscode-ansible/pull/2600

### DIFF
--- a/package.json
+++ b/package.json
@@ -216,6 +216,10 @@
       {
         "command": "ansible.lightspeed.switchProvider",
         "title": "Ansible Lightspeed: Switch Provider"
+      },
+      {
+        "command": "ansible.lightspeed.openLlmProviderSettings",
+        "title": "Ansible Lightspeed: Open LLM Provider Settings"
       }
     ],
     "configuration": [
@@ -429,19 +433,6 @@
       },
       {
         "properties": {
-          "ansible.lightspeed.apiEndpoint": {
-            "default": "",
-            "markdownDescription": "API endpoint URL. **Only configurable for WCA provider.** For Google, the endpoint is automatically set and cannot be changed. Default for WCA: https://c.ai.ansible.redhat.com",
-            "order": 4,
-            "scope": "resource",
-            "type": "string"
-          },
-          "ansible.lightspeed.apiKey": {
-            "markdownDescription": "API key for Google provider. Not used for WCA (uses OAuth2 authentication). **Important**: This will be stored in VS Code settings.",
-            "order": 3,
-            "scope": "resource",
-            "type": "string"
-          },
           "ansible.lightspeed.enabled": {
             "default": true,
             "markdownDescription": "Enable Ansible Lightspeed.",
@@ -449,47 +440,66 @@
             "scope": "resource",
             "type": "boolean"
           },
-          "ansible.lightspeed.modelName": {
-            "markdownDescription": "Model name/ID to use. **Optional for Google** (uses smart defaults).",
-            "order": 2,
-            "scope": "resource",
-            "type": "string"
-          },
-          "ansible.lightspeed.provider": {
-            "default": "wca",
-            "enum": [
-              "wca",
-              "google"
-            ],
-            "enumDescriptions": [
-              "Red Hat Ansible Lightspeed with IBM watsonx Code Assistant",
-              "Google Gemini - Direct access to Google Gemini models (use your Google API key)"
-            ],
-            "markdownDescription": "Select the AI provider for Ansible Lightspeed suggestions and generation.",
-            "order": 1,
-            "scope": "resource",
-            "type": "string"
-          },
           "ansible.lightspeed.suggestions.enabled": {
             "default": true,
             "markdownDescription": "Enable inline suggestions. Note: Currently only supported with WCA provider.",
-            "order": 7,
+            "order": 1,
             "scope": "resource",
             "type": "boolean"
           },
           "ansible.lightspeed.suggestions.waitWindow": {
             "default": 0,
             "markdownDescription": "Delay (in msecs) prior to sending an inline suggestion request.",
-            "order": 8,
+            "order": 2,
             "scope": "resource",
             "type": "number"
           },
           "ansible.lightspeed.timeout": {
-            "default": 3000,
+            "default": 30000,
             "markdownDescription": "Request timeout in milliseconds for API calls.",
-            "order": 5,
+            "order": 3,
             "scope": "resource",
             "type": "number"
+          },
+          "ansible.lightspeed.provider": {
+            "type": "string",
+            "enum": [
+              "wca",
+              "google"
+            ],
+            "default": "wca",
+            "markdownDescription": "LLM provider to use.",
+            "deprecated": true,
+            "deprecationMessage": "This setting is deprecated and will be removed in a future release. Use the LLM Provider Settings panel instead (Ansible Lightspeed: Open LLM Provider Settings).",
+            "order": 4,
+            "scope": "resource"
+          },
+          "ansible.lightspeed.apiEndpoint": {
+            "type": "string",
+            "default": "",
+            "markdownDescription": "API endpoint URL for the selected provider.",
+            "deprecated": true,
+            "deprecationMessage": "This setting is deprecated and will be removed in a future release. Use the LLM Provider Settings panel instead.",
+            "order": 5,
+            "scope": "resource"
+          },
+          "ansible.lightspeed.modelName": {
+            "type": "string",
+            "default": "",
+            "markdownDescription": "Model name/ID to use.",
+            "deprecated": true,
+            "deprecationMessage": "This setting is deprecated and will be removed in a future release. Use the LLM Provider Settings panel instead.",
+            "order": 6,
+            "scope": "resource"
+          },
+          "ansible.lightspeed.apiKey": {
+            "type": "string",
+            "default": "",
+            "markdownDescription": "API key for authentication.",
+            "deprecated": true,
+            "deprecationMessage": "This setting is deprecated and will be removed in a future release. Use the LLM Provider Settings panel instead for secure API key storage.",
+            "order": 7,
+            "scope": "resource"
           }
         },
         "title": "Ansible Lightspeed"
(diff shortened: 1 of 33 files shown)


webviews/LlmProviderApp.vue
```diff
@@ -0,0 +1,689 @@
...
+  transition: opacity 0.2s ease, transform 0.2s ease;
+  pointer-events: none;
+}
+
+.save-indicator.visible {
+  opacity: 1;
+  transform: translateY(0);
+}
+
+/* Spinner animation */
```
goneri: Can you maybe prepare a dedicated Vue component for the spin and move the CSS inside. This way you you can reduce the complexity of this App.

Also we have a dedicated spinner for Lightspeed, https://github.com/ansible/vscode-ansible/blob/main/webviews/lightspeed/src/RoleGenApp.vue#L123 it would be great for consistency if we can use the same.

webviews/LlmProviderApp.vue
```diff
@@ -0,0 +1,689 @@
+<script setup lang="ts">
+import '@vscode/codicons/dist/codicon.css';
```
goneri, line 2: The App is fairly complicated and you can ask Claude to break it up in smaller reusable Vue components.

```
(...)
  ┌────────────────────────┬───────┬──────────────────────────┐                                                                                                                                                     
  │       Component        │ Lines │         Benefit          │                                                                                                                                                     
  ├────────────────────────┼───────┼──────────────────────────┤                                                                                                                                                     
  │ ConfigField.vue        │ ~50   │ Reusable for any form    │                                                                                                                                                     
  ├────────────────────────┼───────┼──────────────────────────┤                                                                                                                                                     
  │ StatusBadge.vue        │ ~30   │ Consistent badge styling │                                                                                                                                                     
  ├────────────────────────┼───────┼──────────────────────────┤                                                                                                                                                     
  │ ActionButton.vue       │ ~40   │ Unified button behavior  │                                                                                                                                                     
  ├────────────────────────┼───────┼──────────────────────────┤                                                                                                                                                     
  │ ProviderConfigForm.vue │ ~70   │ Isolated form logic      │                                                                                                                                                     
  ├────────────────────────┼───────┼──────────────────────────┤                                                                                                                                                     
  │ ProviderHeader.vue     │ ~60   │ Isolated header logic    │                                                                                                                                                     
  ├────────────────────────┼───────┼──────────────────────────┤                                                                                                                                                     
  │ ProviderRow.vue        │ ~40   │ Composition container    │                                                                                                                                                     
  ├────────────────────────┼───────┼──────────────────────────┤                                                                                                                                                     
  │ ToastNotification.vue  │ ~40   │ Shared across webviews   │                                                                                                                                                     
  ├────────────────────────┼───────┼──────────────────────────┤                                                                                                                                                     
  │ useProviderSettings.ts │ ~100  │ Testable business logic  │                                                                                                                                                     
  ├────────────────────────┼───────┼──────────────────────────┤                                                                                                                                                     
  │ LlmProviderApp.vue     │ ~150  │ Down from 690            │                                                                                                                                                     
  └────────────────────────┴───────┴──────────────────────────┘                                                                                                                                                     
  This reduces the main app by ~80% and creates reusable components for future webviews. 
```

webviews/QuickLinksApp.vue
```diff
@@ -8,13 +8,24 @@ const systemReadinessIcon = ref('');
 const systemReadinessDescription = ref('');
```
goneri, line 8: It would be helpful to do a similar refactoring/breakup here.

package.json
```diff
@@ -429,65 +433,31 @@
...
       {
         "properties": {
-          "ansible.lightspeed.apiEndpoint": {
-            "default": "",
-            "markdownDescription": "API endpoint URL. **Only configurable for WCA provider.** For Google, the endpoint is automatically set and cannot be changed. Default for WCA: https://c.ai.ansible.redhat.com",
-            "order": 4,
-            "scope": "resource",
-            "type": "string"
-          },
-          "ansible.lightspeed.apiKey": {
```
goneri, line 439: How do you plan to handle the migration to the new system?

webviews/LlmProviderApp.vue
Abhishekism9450: I split the App into following majorly:

- `ProviderConfigForm.vue` - The Edit form. when a user clicks on Edit, a form appears with fields ( endpoint, key, model, ). This has now its own component. 
- `SpinnerLoading.vue` - The loader spinner. 
- `ToastNotification.vue` - The settings saved popup. 
- `llmProviderState.ts` - All the logic lives here like loading providers, saving configs,..

webviews/LlmProviderApp.vue
Abhishekism9450: sure. I have added a separate Vue component for spin i.e. `SpinnerLoading.vue`.

package.json
Abhishekism9450: Hi @goneri , I am thinking of 2 ways to handle this.

- Add a migration function that runs on extension activation. It would use inspect() to read user-configured LLM settings from settings.json and copy them into the new globalState/secrets storage. A globalState flag can be used to ensure it runs only once.

- Add a note in the docs informing users about the storage change and asking them to re-configure their LLM settings through the new UI. 

@ganeshrn , what are your thoughts on this ?

goneri: In this case, you may want to keep the key, but just flag it as `deprecated: true` with a `deprecationMessage` message, and:
- ignore the legacy key if it's undef
- import the legacy key if it's defined and the new config key is undef.
- ignore the values are the same both side

Abhishekism9450: @goneri , I have few follow-up questions, 
1. Do we only need to keep the `api_key` and `provider` in the settings and mark them as deprecated ? as all other fields have default values(we can remove them).
2. Do we plan to put a deprecation message with the timeline/ target version ?
3. what if both have different values ? do we need to keep the new values and set it in globalState ?

ganeshrn: As discussed in todays extension sync call I think we can deprecate the settings and remove it after 2 releases,

shatakshiiii (approved):

ssbarnea (changes requested):
Blocked until we do see LS tests running on CI via https://github.com/ansible/vscode-ansible/pull/2600 or similar fix.

src/features/lightspeed/vue/views/llmProviderMessageHandlers.ts
```diff
@@ -0,0 +1,372 @@
...
+        ? undefined
+        : "Authentication failed or was cancelled.";
+
+      await this.handleConnectionResult(
+        providerType,
+        displayName,
+        isAuth,
+        error,
+      );
+    }, 2000);
```
anusshukla: The 2-second timeout for checking OAuth completion seems a bit short for a standard authentication flow. In the current implementation, the OAuth logic in `lightSpeedOAuthProvider.ts `uses `LIGHTSPEED_SERVICE_LOGIN_TIMEOUT = 120000 `  to give users sufficient time to open the browser, log in, complete any MFA steps, and be redirected back to VS Code


Additionally, instead of hardcoding the 2000 value, would it be possible to define a constant and reference it here? This would improve readability and make future adjustments easier.

package.json
```diff
@@ -429,67 +433,73 @@
...
             "default": 0,
             "markdownDescription": "Delay (in msecs) prior to sending an inline suggestion request.",
-            "order": 8,
+            "order": 2,
             "scope": "resource",
             "type": "number"
           },
           "ansible.lightspeed.timeout": {
-            "default": 3000,
+            "default": 30000,
```
anusshukla, line 458: Is this change intentional? If so, please document the rationale in the PR description. Increasing the default timeout to 30s may delay failure detection for unreachable endpoints and impact user experience.

Also, should this be noted in the release notes? Currently, 10s the highest timeout we use, could you clarify the reason for increasing it to 30s?

src/features/lightspeed/vue/views/llmProviderMessageHandlers.ts
anusshukla: Also use same constant in test file

package.json
Abhishekism9450: The old 3000 default in package.json was inconsistent. Since settings.ts reads the value as lightSpeedSettings.get("timeout", 30000), the 30s fallback was already being used in practice. This change aligns the declared default with the actual runtime behavior. No functional change for existing users.

src/features/lightspeed/vue/views/llmProviderMessageHandlers.ts
Abhishekism9450: Good catch. The setTimeout(2000) was unnecessary since executeCommand(LIGHTSPEED_AUTH_REQUEST) already awaits the full OAuth flow (which uses LIGHTSPEED_SERVICE_LOGIN_TIMEOUT = 120000 internally). Removed the setTimeout entirely and now check isAuthenticated() directly after the awaited command returns. Updated the corresponding tests to drop fake timers as well.

anusshukla (approved):

ssbarnea (changes requested):
This change does not have any minimal acceptance tests included (e2e or ui/selenium), it only includes unit tests. Due to its nature and size it would require at lest on e2e/ui test that validates the presence of this into the final build.

The importance of this is exacerbated by the fact that at this moment we do not have any LS related testing running on ui. See https://github.com/ansible/vscode-ansible/issues/2611

To be clear, I am not asking about a full testing using UI (smoke) testing, only minimal, just to check that it is there. I would also be acceptable to modify existing test to include this extra check for it as long we see the test passing and we can run it manually using `pytest -k test_foo`.

PS. I also see SonarCloud gate reporting that it does not meet 80% coverage, luckily it seems to only one file and very close to reach the bar.

ssbarnea (approved):

PR #2605, opened 2026-02-27
   chore(selenium): add MCP server smoke tests

raksrh:
## Summary

Adds Selenium smoke tests for the Ansible Development Tools MCP Server to ensure the feature can be enabled, started, and used from the UI.

fixes: AAP-64488

## What's covered

- **Enable via command**: Run `>ansible.mcpServer.enabled` and assert success notification.
- **Disable**: Run `>ansible.mcpServer.disable` and assert disable notification.
- **Enable via settings**: With `ansible.mcpServer.enabled: true` in settings, run enable command and assert already-enabled notification.
- **Usable flow**: Enable MCP server, open Ansible Development Tools view, open welcome page, and assert the MCP Server (AI) option is shown.

Tests use the existing `@pytest.mark.vscode` Selenium setup and do not exercise full AI/Language Model integration.

### DIFF
--- a/test/selenium/ui/test_mcp_server.py
+++ b/test/selenium/ui/test_mcp_server.py
@@ -1,0 +1,134 @@
+"""Smoke tests for Ansible Development Tools MCP Server.
+
+Tests cover: enable MCP server (via command or settings), verify MCP server
+starts (success notification), and verify MCP server is usable (welcome page
+shows MCP section). Full AI/Language Model integration is not exercised here.
+"""
+
+# pylint: disable=E0401, W0613, R0801
+import time
+from typing import Any
+
+import pytest
+from selenium.webdriver.common.action_chains import ActionChains
+from selenium.webdriver.common.keys import Keys
+
+from test.selenium.utils.ui_utils import (
+    ensure_vscode_ready,
+    find_element_across_iframes,
+    vscode_run_command,
+)
+
+
+def _assert_enable_notification(driver: Any) -> None:
+    """Assert that a notification indicates MCP Server is enabled (or already enabled)."""
+    notification = find_element_across_iframes(
+        driver,
+        "//*[contains(., 'MCP Server') and contains(., 'enabled')]",
+        retries=10,
+    )
+    assert notification is not None, (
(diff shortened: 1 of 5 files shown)


test/selenium/ui/test_mcp_server.py
```diff
@@ -0,0 +1,145 @@
...
+    text = notification.text or notification.get_attribute("textContent") or ""
+    assert "MCP Server" in text, (
+        f"Expected MCP disable notification to mention 'MCP Server', got: {text!r}"
+    )
+    assert "disabled" in text, (
+        f"Expected MCP disable notification to mention 'disabled', got: {text!r}"
+    )
+
+
+def _ensure_vscode_ready(driver: Any) -> None:
```
alisonlhart: Note: We should move this to the utils file for reuse in other tests. Not something to be modified now due to the time sensitivity of this, just a note for later.

test/selenium/ui/test_mcp_server.py
```diff
@@ -0,0 +1,145 @@
...
+    driver, _ = browser_setup
+    _ensure_vscode_ready(driver)
+
+    vscode_run_command(driver, ">ansible.mcpServer.enabled")
+    time.sleep(2)
+    _assert_enable_notification(driver)
+
+
+@pytest.mark.vscode
+def test_mcp_server_usable(
```
alisonlhart, line 106: So this check is a little different than what we need to verify. The reason this check won't work is the MCP section will always show up in the welcome page even if the server is disabled. 

What we need is - after the server is enabled, use the command palette to list available MCP servers, and check if the "Ansible Development Tools MCP Server" is present in the list. If it is, then packaging was successful.

test/selenium/ui/test_mcp_server.py
raksrh: Okay. sure

test/selenium/ui/test_mcp_server.py
raksrh: ohh yeah. I'll make the changes and push.

ssbarnea (approved):

PR #2625, opened 2026-03-10
fix: add type to workflow_dispatch input and clean up publish conditions

raksrh:
## Summary

Fixes the publish job being skipped during releases by using more reliable tag detection and properly typing the workflow_dispatch input.

## Changes

- **Use `startsWith(github.ref, 'refs/tags/')` instead of `github.ref_type == 'tag'`** for more reliable tag detection
- Added `type: boolean` to the `publish` workflow_dispatch input (was missing, causing GitHub to send `inputs: null`)
- Simplified `github.event_name == 'release'` condition (removed redundant `&& github.event.action == 'published'`)
- Removed redundant `== 'true'` string checks (input is now properly typed as boolean)
- Applied changes consistently across publish, publish-npm, and Sonar continue-on-error conditions

## Testing

Will test with v26.3.2 tag after merge.

Fixes the issue where publish job was being skipped even though all conditions should evaluate to true.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,7 +24,8 @@ on:
       publish:
         description: "Publish a pre-release"
         required: false
-        default: "false"
+        type: boolean
+        default: false
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.ref }}-${{ github.event.pull_request.number || github.sha }}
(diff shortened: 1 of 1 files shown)


shatakshiiii (approved):

.github/workflows/ci.yaml
```diff
@@ -24,7 +24,8 @@ on:
       publish:
         description: "Publish a pre-release"
         required: false
-        default: "false"
+        type: boolean
```
ssbarnea, line 27: there is a genuine reason why I did not use booleans for these as github translates all arguments to environment variables and how boolean are converted to environment variables is not very clear. By using strings we ensure that we know what these environment variables will container.

raksrh: Without `type: boolean`, GitHub sends inputs: null instead of `{"publish": "false"}`, which breaks manual workflow dispatch. Tested in run #22856144126.

PR #2652, opened 2026-03-16
fix: remove Node DEP0190 and ls stderr from unit test output

rockygeekz:
## Summary
Stops `ls` stderr from appearing in unit test output without changing behavior (return values and error handling unchanged).

## Changes
- **ALS `executionEnvironment.ts`** – `isPluginInPath`: run container `ls` via `spawnSync` with `stdio: ['ignore','pipe','ignore']` so stderr is not forwarded; still return false when exit code !== 0 or stdout is empty.
## Testing
- `task unit` / `task unit -f` runs without DEP0190 warnings and without `ls: cannot access ...` in output.
- `task lint` and `task test`.

fixes: AAP-65959  
related: AAP-64395

### DIFF
--- a/packages/ansible-language-server/src/services/executionEnvironment.ts
+++ b/packages/ansible-language-server/src/services/executionEnvironment.ts
@@ -508,15 +508,27 @@ export class ExecutionEnvironment {
   ): boolean {
     /* v8 ignore next 21 */
     const completeSearchPath = path.join(searchPath, pluginFolderPath);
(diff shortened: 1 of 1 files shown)


ssbarnea (changes requested):
You missed the whole point of the previous change, we MUST use `shell: true` everywhere because without it we would fail to find some of the tools (example if they were installed using mise).

Once `shell` becomes a requirement, we cannot escape the DEP0190.

PR #2675, opened 2026-03-20
AAP-69248: WCA OAuth ignores custom Lightspeed URL configuration

hasys:
refactor: simplify getBaseUri by removing WCA fallback logic

Address PR review feedback by fully adopting fail-fast strategy:

1. Remove WCA-specific fallback to WCA_API_ENDPOINT_DEFAULT
   - Trust llmProviderSettings.get() to return defaults via provider factory
   - Add explicit error when apiEndpoint is not configured (defensive)

2. Remove fallback to legacy settings schema
   - Only use llmProviderSettings after migration
   - Check llmProviderSettings initialization first (fail early)

3. Update all tests to match new behavior
   - Tests now expect provider factory defaults
   - Add new tests for empty endpoint error scenarios
   - Remove dependency on old settings structure in mocks

This simplification:
- Removes 15+ lines of redundant code
- Eliminates technical debt from legacy settings
- Provides clearer error messages for misconfiguration
- Follows fail-fast principle for better debugging

All 1544 tests passing, lint checks pass.

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

Fixed WCA OAuth to respect custom Lightspeed URL configuration by refactoring `getBaseUri` from synchronous to asynchronous, removing WCA-specific fallbacks and legacy settings dependencies. The function now explicitly requires `llmProviderSettings` and throws clear errors when the API endpoint is unconfigured. This ensures OAuth and code generation work with custom certificates and custom Lightspeed URLs.

- Made `getBaseUri` asynchronous and removed WCA-specific fallback to legacy settings
- Updated `LightSpeedAPI.lightspeedPost` to await `getBaseUri` 
- Updated OAuth provider authorization and token endpoints to await `getBaseUri`
- Converted `logAuthProviderDebugHints` to async and removed synchronous blocking calls in constructor
- Updated Lightspeed endpoint calls in `lightspeedUser.ts` to await `getBaseUri`
- Added explicit error handling for missing `llmProviderSettings` and empty API endpoints
- Enhanced error logging in token request flows with additional network error metadata
- Added comprehensive unit tests for `getBaseUri` covering custom endpoints, normalization, and error scenarios
- Added unit tests for `getFetch` behavior with custom certificates and Electron integration
- Added "Ollama" to spellcheck dictionary

related: #AAP-69248


src/features/lightspeed/api.ts
```diff
@@ -46,14 +46,25 @@ const UNKNOWN_ERROR: string = "An unknown error occurred.";
...
+    const electron = (globalThis as any).require?.('electron') ?? require('electron');
+    const electronFetch = electron?.net?.fetch;
+    if (electronFetch && typeof electronFetch === 'function') {
+      console.debug('[getFetch] Using electron.net.fetch');
       return electronFetch;
     }
-  } catch {
-    // electron not available (e.g., in tests or web environment)
+  } catch (error) {
+    console.debug('[getFetch] Electron not available, falling back to globalThis.fetch:', error);
```
goneri: This may be a bit too verbose, I'm afraid you will flood the console in a web code-server environment.

src/features/lightspeed/api.ts
```diff
@@ -46,14 +46,25 @@ const UNKNOWN_ERROR: string = "An unknown error occurred.";
...
   }
   // Fallback to global fetch
+  console.debug('[getFetch] Using globalThis.fetch (Node.js undici)');
```
goneri: same.

rockygeekz (changes requested):
also the CI failure is because the cspell hook auto-added "Ollama" to the dictionary but it wasn't committed. This should pass the CI.

src/features/lightspeed/utils/webUtils.ts
```diff
@@ -50,22 +50,48 @@ export function calculateTokenExpiryTime(expiresIn: number) {
...
+    } catch (error) {
+      console.debug(
+        "[getBaseUri] Failed to get provider-specific endpoint, falling back to legacy settings",
+        error,
+      );
+    }
+  }
+
+  // Fall back to legacy settings if provider-specific setting is not available
+  if (!baseUri) {
```
goneri: This should never happens and this may actually cause us to miss some unexpected problems with the logic upper.

https://github.com/ansible/vscode-ansible/blob/main/src/features/lightspeed/llmProviderSettings.ts#L24-L75

src/features/lightspeed/utils/webUtils.ts
```diff
@@ -50,22 +50,48 @@ export function calculateTokenExpiryTime(expiresIn: number) {
...
+    baseUri = settingsManager.settings.lightSpeedService.apiEndpoint.trim();
+  }
+
+  // For WCA: use custom endpoint if configured, otherwise use default
   if (provider === "wca") {
```
goneri: Same here. You should be able to remove the whole block.

src/features/lightspeed/utils/webUtils.ts
```diff
@@ -50,22 +50,48 @@ export function calculateTokenExpiryTime(expiresIn: number) {
...
-): string {
+): Promise<string> {
   // Check the provider from globalState via lightSpeedManager if available
   const provider =
     providerOverride ||
     lightSpeedManager?.llmProviderSettings?.getProvider() ||
     settingsManager.settings.lightSpeedService.provider;
 
-  // For WCA, always use the default WCA endpoint
+  // Try to get endpoint from new provider-specific settings first
```
goneri: We may want to raise an exception or an error if `provider` is undefined or an empty string.

goneri (changes requested):

src/features/lightspeed/lightspeedUser.ts
```diff
@@ -81,13 +81,13 @@ export class LightspeedUser {
           ? ExtensionHost.Local
           : ExtensionHost.Remote
         : ExtensionHost.WebWorker;
-    this.logAuthProviderDebugHints();
+    this.logAuthProviderDebugHints(); // fire-and-forget async call
```
goneri: This is not really

test/unit/lightspeed/utils/webUtils.test.ts
```diff
@@ -0,0 +1,421 @@
...
+  } | null;
+};
+
+describe("webUtils - getBaseUri", () => {
+  let mockSettingsManager: SettingsManager;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+
+    // Create a minimal mock settings manager
```
goneri, line 35: This should triggers the migration to the new settings schema, See src/features/lightspeed/llmProviderSettings.ts:24-75 via migrateFromSettingsJson().

test/unit/lightspeed/utils/webUtils.test.ts
```diff
@@ -0,0 +1,421 @@
...
+
+// Type for mocked lightSpeedManager to avoid 'any' usage
+type MockedLightSpeedManager = {
+  llmProviderSettings: {
+    getProvider: ReturnType<typeof vi.fn>;
+    get: ReturnType<typeof vi.fn>;
+  } | null;
+};
+
+describe("webUtils - getBaseUri", () => {
```
goneri, line 29: IMO, we should only assume the configuration is based on `llmProviderSettings`. The extension handles the migration to the new schema already. If for some reason we need to fallback on the old system, it means that we've got a problem with this logic.

The old system will anyway be deprecated soon.

goneri (changes requested):

src/features/lightspeed/lightspeedUser.ts
```diff
@@ -140,7 +140,7 @@ export class LightspeedUser {
 
     try {
       const response = await fetch(
-        `${getBaseUri(this._settingsManager)}${LIGHTSPEED_ME_AUTH_URL}`,
+        `${await getBaseUri(this._settingsManager)}${LIGHTSPEED_ME_AUTH_URL}`,
```
de1987, line 151: just be careful when using `await`, this call can block the Ansible Plugin from performing another actions, which might led to bad user navigation experience

de1987: when it's `async`, other operations will not block the other operations from running

goneri: The code is fine. `await` won't block the other async functions. It's just a way to say, "I want the actual result of the function, not the promise".

src/features/lightspeed/api.ts
```diff
@@ -46,12 +46,22 @@ const UNKNOWN_ERROR: string = "An unknown error occurred.";
 
 export function getFetch() {
   try {
-    const electronFetch = require("electron")?.net?.fetch;
-    if (electronFetch) {
+    // Check if NODE_EXTRA_CA_CERTS is set - if so, use Node.js fetch
+    // because electron.net.fetch doesn't honor custom CA certificates
+    if (process.env.NODE_EXTRA_CA_CERTS) {
```
goneri: This is a rather unexpected/dangerous change IMO and this will likely break the proxy support. I would prefer to see it in a different PR.

hasys: Without it, it is not possible to use VS Code plugin with custom certificates. Those are just ignored entirely and there is no other way to configure them.

This variable is default for node.js https://nodejs.org/api/cli.html#node-extra-ca-certsfile

hasys: To use it, user should explicitly set it up like this:
```
NODE_EXTRA_CA_CERTS="/tmp/ingress-operator-ca-correct.pem" /Applications/Visual\ Studio\ Code.app/Contents/Resources/app/bin/code .
```

goneri: Yes, I understand this, but:
- this goes beyond the scope of this PR, which is to focus on the OAuth2 authentification,
- VSCode injects the proxy configuration in the Electron network stack (AFAIR), if we silently switch to Node's `fetch` implementation, we will lose it,
- and this kind of changes are really really hard to test.

rockygeekz (approved):

ssbarnea (dismissed):
Please ensure that the PR title and body is following our rules. See AGENTS.md file, also for internal work, be sure you include JIRA ticket in body using JIRA required syntax (smart commits).

src/features/lightspeed/utils/webUtils.ts
```diff
@@ -50,22 +47,37 @@ export function calculateTokenExpiryTime(expiresIn: number) {
...
+
+  // Get endpoint from provider-specific settings
+  // Migration from legacy settings happens during extension activation via migrateFromSettingsJson()
+  // Provider factory ensures default endpoints are returned when not configured
+  const baseUri = (
+    await lightSpeedManager.llmProviderSettings.get(provider, "apiEndpoint")
+  ).trim();
 
-  // For WCA, always use the default WCA endpoint
-  if (provider === "wca") {
```
goneri, line 64: We should set this value during the migration if the original is empty.

goneri: Nevermind, it's already the case.

goneri (approved):

PR #2695, opened 2026-03-27
feat: use local HTTP server for OAuth callback - AAP-70006

goneri:
Replace reliance on the `vscode://` URI scheme handler (which is not
always functional at OS level) with a local HTTP server on a random
port to receive the OAuth redirect. This behaviour is only enabled
for the Lightspeed / SAAS endpoint.

## Changes

- **Local loopback OAuth callback**: start an ephemeral HTTP server on
  `127.0.0.1:<random-port>/callback` and use it as the `redirect_uri`
  when the base URI matches SAAS hosts. The `vscode://` URI handler is
  used for all other endpoints.
- **Redirect URI threading**: pass the actual `redirectUri` through the
  call chain (`handleUriForCode` → `requestOAuthAccountFromCode`)
  instead of storing it in a mutable field, avoiding cross-request
  race conditions.
- **Sign-out on config change**: sign out the current OAuth session when
  the provider or API endpoint setting changes, so stale tokens are
  not reused.
- **Server listen error handling**: properly reject the listen promise
  on startup failure instead of hanging indefinitely.
- **Callback page improvements**: use a neutral "Authorization received"
  message (token exchange hasn't completed yet), escape HTML in error
  messages to prevent reflected markup, and return HTTP 400 on error.
- **ESLint config**: allow underscore-prefixed unused vars/args
  (`argsIgnorePattern: "^_"`, `varsIgnorePattern: "^_"`) and remove
  now-redundant `eslint-disable-next-line` comments.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/features/lightspeed/lightSpeedOAuthProvider.ts
+++ b/src/features/lightspeed/lightSpeedOAuthProvider.ts
@@ -318,8 +318,8 @@ export class LightSpeedAuthenticationProvider
       const escapeHtml = (value: string) =>
         value.replace(/[&<>"']/g, (char) => entities[char] ?? char);
 
-      res.writeHead(200, { "Content-Type": "text/html" });
       if (code) {
+        res.writeHead(200, { "Content-Type": "text/html" });
         res.end(
           "<html><body><h1>Authorization received</h1><p>Return to your editor to finish signing in.</p></body></html>",
         );
@@ -327,6 +327,7 @@ export class LightSpeedAuthenticationProvider
       } else {
         const errorMsg =
           error || "No authorization code received from the server";
+        res.writeHead(400, { "Content-Type": "text/html" });
         res.end(
           `<html><body><h1>Authentication failed</h1><p>${escapeHtml(errorMsg)}</p></body></html>`,
         );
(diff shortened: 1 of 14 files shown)


src/features/lightspeed/lightSpeedOAuthProvider.ts
```diff
@@ -276,34 +280,152 @@ export class LightSpeedAuthenticationProvider
...
+        res.end(
+          "<html><body><h1>Authorization received</h1><p>Return to your editor to finish signing in.</p></body></html>",
+        );
+        resolveCode(code);
+      } else {
+        const errorMsg =
+          error || "No authorization code received from the server";
+        res.end(
+          `<html><body><h1>Authentication failed</h1><p>${escapeHtml(errorMsg)}</p></body></html>`,
+        );
```
de1987, line 333: with an error, shouldn't the `writeHead` be http status error, such as `4xx` or `5xx`?
wdyt?

goneri: done, thanks.

de1987 (approved):

goneri:
We prefer to keep the statu-quo and we may revisit this later.

PR #2696, opened 2026-03-27
fix: sign out OAuth session when provider or API endpoint changes

goneri:
When the active provider or WCA API endpoint is changed through the
LLM Provider panel or provider commands, all existing OAuth sessions
are removed (equivalent to clicking "Sign Out" in VSCode), forcing a
fresh login against the new configuration.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>


shatakshiiii:
Hi @goneri, just checking if you'd be continuing work on this PR ?

PR #2721, opened 2026-04-07
fix: use persistent container for execution environment commands

lalten:
## Description

The execution environment (EE) implementation previously spawned a new ephemeral `docker run --rm` container for every single background diagnostic and linting command. This caused significant overhead as each invocation paid the full cost of container creation (namespace, overlayFS, veth pair setup) even though all commands ran against the same image and volume mounts.

This PR replaces ephemeral containers with a single persistent background container managed via `docker exec` / `podman exec`:

- A long-lived container (`sleep infinity`) is started once during `initialize()` and reused for all subsequent commands via `docker exec`.
- Container name is derived from a SHA-256 hash of the workspace URI (`als_persistent_<hash12>`), preventing collisions and enabling reliable cleanup.
- `ensurePersistentContainerHealthy()` performs debounced health checks (5s interval) and auto-restarts dead containers transparently.
- `dispose()` cleans up the persistent container on workspace removal, config change, or server shutdown. Wired into `WorkspaceManager` and `AnsibleLanguageService` shutdown handler.
- `getCachedCommand` / `setCachedCommand` / `clearCommandCache` API for callers to avoid redundant `docker exec` invocations (public, no callers yet).
- Bug fix: `copyPluginDocFiles` had a latent race condition — `docker cp` calls via `asyncExec` were fire-and-forget inside `forEach`. Converted to `for...of` with `await`.

fix: use persistent container for execution environment commands. Execution now reuses a single workspace-scoped long-lived Docker/Podman container (named als_persistent_<sha256 hash12>) via exec instead of spawning/removing a container per command. Persistent health checks auto-restart dead containers, workspace/server disposal performs cleanup, and a command cache reduces redundant exec overhead.

- Start and reuse the persistent container during initialization (deterministic naming + health monitoring/restart)
- Add command-result caching APIs (`getCachedCommand`, `setCachedCommand`, `clearCommandCache`)
- Dispose execution environments on workspace removal/config changes and during language service shutdown
- Fix `copyPluginDocFiles` race by switching `docker cp` from fire-and-forget `forEach` to awaited sequential `for...of`
  
Related: `#2994`
Related: `#2720`


sathyapramod:
CI is failing because `pnpm-lock.yaml` is out of sync with `packages/ansible-language-server/package.json`.

The lockfile still references the old specifiers (`^10.0.1` for `vscode-languageserver` and `^3.18.1` for `vscode-languageserver-protocol`) while `package.json` now declares `^9.0.1` and `^3.17.5`.

**Fix:** Run `pnpm install` to regenerate the lockfile, then commit the updated `pnpm-lock.yaml`:

```bash
pnpm install
git add pnpm-lock.yaml
git commit -m "chore: regenerate pnpm-lock.yaml"
git push
```

sathyapramod:
## Code Review: fix: use persistent container for execution environment commands

### Verdict: NEEDS_CHANGES

### Scores
| Lens | Score | Findings |
|---|---|---|
| Functionality | 1/10 | 3 Critical, 2 Major, 2 Minor |
| Security | 6/10 | 3 Major, 2 Minor, 1 Nit |
| Quality | 7.5/10 | 5 Minor |
| **Overall** | **4.8/10** | |

### Findings

#### Critical

1. **[Functionality]** **[executionEnvironment.ts:~719]** Stale variable reference `copyArgv` causes `ReferenceError` crash. The `copyPluginDocFiles` method builds a `copyCommand` string but the log statement still references the removed variable `copyArgv` and `containerEngine` (also out of scope).
   - **Evidence**:
     ```typescript
     const copyCommand = `${this._container_engine} cp ${shellQuote(`${containerName}:${srcPath}`)} ${shellQuote(destPathFolder)}`;
     this.connection.console.log(
       `Copying plugins from container to local cache path ${containerEngine} ${copyArgv.join(" ")}`,
     );
     ```
   - **Confidence**: HIGH
   - **Fix**: Replace the log line with:
     ```typescript
     this.connection.console.log(
       `Copying plugins from container to local cache path: ${copyCommand}`,
     );
     ```
   - **Points**: 2

2. **[Functionality]** **[executionEnvironment.ts:~359]** `settingsVolumeMounts` pushed without `-v` flag prefix. The old code used `push("-v", volumeMountSpec)` for each mount. The new code spreads the array directly, so docker/podman will interpret specs like `/host/path:/container/path` as positional arguments — silently breaking all user-configured volume mounts.
   - **Evidence**:
     ```typescript
     if (this.settingsVolumeMounts && this.settingsVolumeMounts.length > 0) {
       containerCommand.push(...this.settingsVolumeMounts);
     }
     ```
   - **Confidence**: HIGH
   - **Fix**:
     ```typescript
     if (this.settingsVolumeMounts && this.settingsVolumeMounts.length > 0) {
       for (const mount of this.settingsVolumeMounts) {
         containerCommand.push("-v", mount);
       }
     }
     ```
   - **Points**: 2

3. **[Functionality]** **[package.json + pnpm-lock.yaml]** Lockfile specifier mismatch: `package.json` declares `vscode-languageserver: "^9.0.1"` but `pnpm-lock.yaml` still shows `specifier: ^10.0.1, version: 10.0.1`. This causes `pnpm install --frozen-lockfile` to fail in CI and is likely the root cause of the CI failures.
   - **Evidence**:
     ```json
     "vscode-languageserver": "^9.0.1",
     "vscode-languageserver-protocol": "^3.17.5",
     ```
     While lockfile retains `version: 10.0.1`
   - **Confidence**: HIGH
   - **Fix**: Run `pnpm install` to regenerate the lockfile, or revert the version change if unintentional.
   - **Points**: 2

#### Major

4. **[Security]** **[executionEnvironment.ts:~398]** Architectural regression: container startup uses `child_process.execSync(command)` with a concatenated string. `execSync` with a string argument invokes `/bin/sh -c`, enabling shell interpretation. The old code used `spawnSync`/`spawnSyncWithResult` with argument arrays and `shell: false`, which was safe by construction. This change means every value in `containerCommand` must be correctly quoted — and the quoting is applied inconsistently.
   - **Evidence**:
     ```typescript
     const command = containerCommand.join(" ");
     child_process.execSync(command, { encoding: "utf-8" });
     ```
   - **Confidence**: HIGH
   - **Fix**: Use `child_process.spawnSync(containerCommand[0], containerCommand.slice(1), { encoding: "utf-8", shell: false })` and remove `shellQuote` calls from arguments (they become literal strings with `shell: false`).
   - **Points**: 1

5. **[Security]** **[executionEnvironment.ts:~254-276]** `execInContainer()` appends the `command` parameter directly without quoting. The `--workdir` argument is `shellQuote`-d, but the command string (which contains user-influenced values like workspace paths and executable arguments) is appended raw. The resulting string is executed via shell.
   - **Evidence**:
     ```typescript
     containerCommand.push("--workdir", shellQuote(workspaceFolderPath));
     containerCommand.push(...ansibleEnvArgs());
     containerCommand.push(this._persistentContainerName);
     containerCommand.push(command);  // NOT quoted
     return containerCommand.join(" ");
     ```
   - **Confidence**: HIGH
   - **Fix**: Either use `spawnSync` with `shell: false` (see #4), or wrap the command: `containerCommand.push("sh", "-c", shellQuote(command));`
   - **Points**: 1

6. **[Security]** **[executionEnvironment.ts:~380]** `settingsContainerOptions` parsed with naive `.split(" ")` instead of `parseContainerOptions()`. This breaks options containing spaces (e.g., `--label "my label"`) and, since tokens are re-joined and shell-interpreted, could enable injection via crafted options.
   - **Evidence**:
     ```typescript
     const containerOptions = this.settingsContainerOptions.split(" ");
     ```
   - **Confidence**: HIGH
   - **Fix**: Restore `parseContainerOptions()`:
     ```typescript
     const containerOptions = parseContainerOptions(this.settingsContainerOptions);
     ```
   - **Points**: 1

7. **[Functionality]** **[executionEnvironment.ts:~189-195]** `cacheBase` computed from `XDG_CACHE_HOME` but then ignored — `hostCacheBasePath` hardcodes `process.env.HOME/.cache`. Users with `XDG_CACHE_HOME` set will have cache written to the wrong location.
   - **Evidence**:
     ```typescript
     const cacheBase =
       process.env.XDG_CACHE_HOME ||
       `${process.env.HOME || os.homedir()}/.cache`;
     const hostCacheBasePath = path.resolve(
       `${process.env.HOME}/.cache/ansible-language-server/${imageSafeName}/${this._container_image_id}`,
     );
     ```
   - **Confidence**: HIGH
   - **Fix**:
     ```typescript
     const hostCacheBasePath = path.resolve(
       `${cacheBase}/ansible-language-server/${imageSafeName}/${this._container_image_id}`,
     );
     ```
   - **Points**: 1

8. **[Functionality]** **[workspaceManager.ts:~164-182]** Debounce removal on `onConfigurationChanged` enables race conditions. The old 500ms debounce prevented rapid-fire config changes from causing thrashing. The new code directly clears services, allowing concurrent `startPersistentContainer()` calls racing on the same deterministic container name.
   - **Evidence**:
     ```typescript
     this.documentSettings.onConfigurationChanged(
       this.workspaceFolder.uri,
       () => {
         this._executionEnvironment = undefined;
         this._ansibleConfig = undefined;
         this._docsLibrary = undefined;
         this._ansibleInventory = undefined;
       },
     );
     ```
   - **Confidence**: MEDIUM
   - **Fix**: Re-introduce debouncing or add a guard against concurrent `startPersistentContainer()` calls.
   - **Points**: 1

#### Minor

9. **[Security]** **[executionEnvironment.ts:~71-73]** `shellQuote` does not reject null bytes, which can cause argument truncation in C-based shell implementations.
   - **Evidence**:
     ```typescript
     function shellQuote(s: string): string {
       return `'${s.replace(/'/g, "'\\''")}'`;
     }
     ```
   - **Confidence**: MEDIUM
   - **Fix**: Add null byte rejection: `if (s.includes('\0')) throw new Error('Null bytes not allowed');`
   - **Points**: 0.5

10. **[Security]** **[executionEnvironment.ts:~707-708]** Path traversal fix is incomplete — strips leading separators but does not block `../` sequences. `path.join(base, "../../etc/passwd")` resolves outside the base directory.
    - **Evidence**:
      ```typescript
      const safeSrcPath = srcPath.replace(/^[/\\]+/, "");
      const destPath = path.join(hostPluginDocCacheBasePath, safeSrcPath);
      ```
    - **Confidence**: MEDIUM
    - **Fix**: Verify resolved path is within base:
      ```typescript
      const destPath = path.resolve(hostPluginDocCacheBasePath, safeSrcPath);
      if (!destPath.startsWith(path.resolve(hostPluginDocCacheBasePath) + path.sep)) {
        this.connection.console.warn(`Skipping path traversal attempt: ${srcPath}`);
        continue;
      }
      ```
    - **Points**: 0.5

11. **[Functionality]** **[executionEnvironment.ts:~79-80]** `ansibleEnvArgs()` couples shell-quoting with argument generation. `-e` values like `ANSIBLE_FOO='bar'` with literal quotes are correct when going through a shell, but would break if args are ever used with `spawnSync` `shell: false`.
    - **Evidence**: `args.push("-e", \`${key}=${shellQuote(value ?? "")}\`);`
    - **Confidence**: MEDIUM
    - **Fix**: If switching to `shell: false` (per finding #4), remove `shellQuote` from this function.
    - **Points**: 0.5

12. **[Functionality]** **[executionEnvironment.ts:~398-405]** `startPersistentContainer` blocks the event loop with `execSync` plus a `spawnSync` health check with 10s timeout — worst case >10s blocking for a language server.
    - **Evidence**:
      ```typescript
      child_process.execSync(command, { encoding: "utf-8" });
      const healthCheck = child_process.spawnSync(
        this._container_engine,
        ["exec", this._persistentContainerName, "echo", "ok"],
        { encoding: "utf-8", timeout: 10000, shell: false },
      );
      ```
    - **Confidence**: HIGH
    - **Fix**: Add a timeout to `execSync` and consider async alternatives.
    - **Points**: 0.5

13. **[Quality]** **[executionEnvironment.ts:~529]** `ensurePersistentContainerHealthy` ignores `startPersistentContainer()` return value on restart.
    - **Evidence**: `this.startPersistentContainer(); // return value ignored`
    - **Confidence**: LOW
    - **Fix**: Log the failure: `if (!this.startPersistentContainer()) { ... }`
    - **Points**: 0.5

14. **[Quality]** **[executionEnvironment.ts]** `_commandCache` Map grows indefinitely with no TTL, size limit, or eviction policy.
    - **Evidence**: `private _commandCache: Map<string, { stdout: string; stderr: string }> = new Map();`
    - **Confidence**: MEDIUM
    - **Fix**: Add maximum cache size or LRU eviction.
    - **Points**: 0.5

15. **[Quality]** **[executionEnvironment.ts:~622-632]** Dead code: `cleanUpContainer` has duplicate container lookup — once inline (unused `allContainers` variable) and once via `getContainerIds`.
    - **Evidence**: Inline `allContainers` computed but never used for removal; `getContainerIds` duplicates the work.
    - **Confidence**: HIGH
    - **Fix**: Remove the unused inline block or replace `getContainerIds` with it.
    - **Points**: 0.5

16. **[Quality]** **[workspaceManager.ts:~103-112]** `disposeExecutionEnvironment()` called fire-and-forget (`void`) before `folderContexts.delete()`. The async dispose is deferred to the next microtask, but the delete happens synchronously first.
    - **Evidence**:
      ```typescript
      void context.disposeExecutionEnvironment();
      this.folderContexts.delete(removedUri);
      ```
    - **Confidence**: MEDIUM
    - **Fix**: `await` the disposal or chain the delete.
    - **Points**: 0.5

17. **[Quality]** **[executionEnvironment.test.ts]** Missing test coverage for `startPersistentContainer` reuse/image-mismatch paths. Only the stub-returns-false path is tested.
    - **Evidence**: Tests only stub `startPersistentContainer` to return false; no test exercises the actual implementation.
    - **Confidence**: HIGH
    - **Fix**: Add tests mocking `doesContainerNameExist`, `isContainerRunning`, `getContainerImage`.
    - **Points**: 0.5

#### Nit

18. **[Security]** **[executionEnvironment.ts:~75-83]** `ansibleEnvArgs` key is not validated — env var names could theoretically contain shell metacharacters on non-POSIX platforms.
    - **Evidence**: `if (key.startsWith("ANSIBLE_")) { args.push("-e", ...); }`
    - **Confidence**: LOW
    - **Fix**: Add regex validation: `if (/^[A-Z_][A-Z0-9_]*$/.test(key)) { ... }`
    - **Points**: 0

### Needs Human Judgment

- **vscode-languageserver downgrade ^10→^9**: Is this intentional? The lockfile wasn't updated. If intentional, this is a significant API change that needs validation. If accidental, revert.
- **Debounce removal**: The comment says `startPersistentContainer()` handles cleanup via deterministic naming. A human should verify this holds under rapid config change scenarios.
- **`execSync` vs `spawnSync` architecture**: The shift to shell-interpreted strings is the root cause of multiple security findings. The author should decide whether to revert to `spawnSync` with `shell: false` (recommended) or systematically audit all quoting.

### Verification Results
| Check | Command | Output | Result |
|-------|---------|--------|--------|
| CI pipeline status | (from PR metadata) | 2 FAILED (package, build windows) | FAIL |
| Stale reference check | (API-only — skipped) | N/A | SKIPPED |
| Codebase grep verification | (API-only — skipped) | N/A | SKIPPED |

### Observations

1. **CI is failing** — the `package` and `build (windows)` jobs failed. The lockfile/package.json version mismatch (Finding #3) is the likely root cause.
2. **3 general comments on the PR** with 0 unresolved review comments — existing discussion was not blocking.
3. **The core idea is sound** — replacing ephemeral containers with a persistent one is a significant performance improvement. The implementation needs corrections in shell safety and a few behavioral regressions.
4. **Good additions**: deterministic container naming via SHA-256, health check with auto-restart, path traversal fix in `copyPluginDocFiles`, proper shutdown/dispose lifecycle.

### Incidental Findings (out of scope — not scored)
None.

### Limitations

This review was performed by an AI agent. It does not understand
business context, domain intent, organizational constraints, or
deployment environment specifics. LOW-confidence findings and items
in "Needs Human Judgment" require human verification. This review
is a first pass, not a final approval.

**Partial review (API-only mode):** This review was performed
without a local checkout of the repository. The following
verification steps were skipped:
- Codebase searches for stale references from deleted files
- grep/lint verification commands from project checklists
- Reading files outside the diff for caller/consumer context
- Prior review iteration detection

Findings are based solely on the diff. Issues requiring
codebase-wide context may be missed.

### Validation Summary
| Lens | Proposed | Removed | Converted | Surviving |
|------|----------|---------|-----------|-----------|
| Functionality | 9 | 2 | 0 | 7 |
| Security | 7 | 1 | 0 | 6 |
| Quality | 5 | 0 | 0 | 5 |
| **Total** | **21** | **3** | **0** | **18** |

### Path to 10/10

Each item is an agent-ready fix prompt — paste into your coding agent to apply:

1. **[Functionality] executionEnvironment.ts:~719** (+2 points)
   > In `copyPluginDocFiles`, the log statement references `copyArgv` and `containerEngine` which were removed. Replace with `copyCommand`.

2. **[Functionality] executionEnvironment.ts:~359** (+2 points)
   > In `startPersistentContainer`, `settingsVolumeMounts` are pushed without `-v` prefix. Change to `for (const mount of this.settingsVolumeMounts) { containerCommand.push("-v", mount); }`.

3. **[Functionality] package.json + pnpm-lock.yaml** (+2 points)
   > Run `pnpm install` to sync the lockfile with the package.json version changes, or revert the vscode-languageserver downgrade.

4. **[Security] executionEnvironment.ts:~398** (+1 point)
   > Replace `execSync(command)` with `spawnSync(containerCommand[0], containerCommand.slice(1), { encoding: "utf-8", shell: false })`. Remove `shellQuote` from arguments that are passed as array elements.

5. **[Functionality] executionEnvironment.ts:~189** (+1 point)
   > Use the computed `cacheBase` variable in `hostCacheBasePath` instead of hardcoding `process.env.HOME/.cache`.

### Final Score
**Overall: 4.8/10 — NEEDS_CHANGES**

---
*Generated with AI assistance: Claude Code / Opus 4.6 (Anthropic)*

lalten:
@sathyapramod thank you for having a look. My interpretation of this massive wall of text is that you're generally ok with the intent of these changes and only care about exactly these findings. Is that right? If you'd like to discuss the general behavior changes, let's do that before fixing all of these.

ssbarnea:
@lalten While I love your PR because this was on my mind from the day we added support for execution environments, I realise that the size of change combined with https://github.com/ansible/vscode-ansible/issues/2994 makes me think that dropping it would be best action, as the new extension will no longer need any of these as the containers will be persistent.

sathyapramod:
Thanks for the contribution @lalten! As @ssbarnea noted, the new extension on the next branch takes a different approach where containers are persistent by design, making this change unnecessary. Closing in favor of the new architecture.

PR #2730, opened 2026-04-13
fix: complete build isolation - no HOME pollution, no files outside out/

rockygeekz:
## Summary

Prevents builds and tests from creating files outside `out/` or polluting
the user HOME directory, completing the remaining work for AAP-65204.

### HOME isolation for unit test suites

- Set `process.env.HOME` to isolated directories under `out/` in globalSetup
  for ALS and extension unit tests
- Added `TMP`/`TMPDIR` env vars to unit, ALS, and e2e tasks in Taskfile

### Relocated build artifacts to `out/`

- Vite dev server markers (`.vite-dev-server-url`, `.vite-dev.pid`)
- `resources/contentCreator/` symlinks created by `uv.sh`
- MCP server tests now use system tmpdir instead of writing next to source

### Cleanup

- Removed `__dirname` derivations in favor of `import.meta.dirname` and
  shared `FIXTURES_BASE_PATH` export
- Removed 9 `.gitignore` entries for files no longer created at root
- Removed stale `exclude:` entries from Taskfile sources

related: AAP-65204
depends-on: #2726

Isolates builds/tests by redirecting temporary files and artifacts into out/, preventing HOME pollution and writes to the repo root; tests, tools, Taskfiles, Vite, and CI were updated to use isolated dirs and perform targeted cleanup.

- Override HOME/USERPROFILE and set ANSIBLE_HOME in globalSetup (unit + ALS)
- Add TMP/TMPDIR envs and create out/* tmp dirs for unit, ALS, e2e tasks
- Move Vite marker/PID files and content-creator artifacts to out/resources/contentCreator/
- Use system tmpdir or per-run mkdtemp for MCP and other tests; switch to sync IO and recursive rm
- Replace __dirname/fileURLToPath with import.meta.dirname and export FIXTURES_BASE_PATH
- Remove obsolete .gitignore entries and stale Taskfile excludes
- Update CI container storage to workspace-scoped paths and add cleanup with podman/buildah unshare fallbacks

related: AAP-65204


packages/ansible-language-server/test/globalSetup.ts
```diff
@@ -66,6 +67,37 @@ function execWithTimeout(
...
 
 export async function setup() {
+  // Clean up any leftover containers from previous runs first (before setting HOME)
+  // This prevents permission errors when CI tries to clean the workspace
+  try {
+    if (!SKIP_PODMAN) {
+      console.info(
+        "Cleaning up any leftover podman containers from previous runs...",
+      );
+      spawnSync("podman", ["system", "reset", "--force"], { stdio: "inherit" });
```
goneri: I like the PR, but this is change in particular a bit like bringing an atomic nuke in a bar brawl :-). Are you 100% sure this only clean the `testHome` directory?
What if the user has different configuration for `CONTAINER_HOST`, `XDG_CONFIG_HOME`, `XDG_DATA_HOME`, `XDG_RUNTIME_DIR`, etc?

Overall, I'm not sure WHY this had to be done?

packages/ansible-language-server/test/globalSetup.ts
```diff
@@ -196,5 +228,25 @@ export async function setup() {
...
 
 export async function teardown() {
-  // Cleanup if needed
+  // Clean up podman/docker containers and storage to avoid permission issues in CI
+  // The isolated HOME stores containers in out/als/tmp/home/.local/share/containers/
+  // which creates root-owned files that CI can't clean up without this
+  try {
+    if (!SKIP_PODMAN) {
+      console.info("Cleaning up podman containers and storage...");
+      spawnSync("podman", ["system", "reset", "--force"], { stdio: "inherit" });
```
goneri: Yes, same.

tools/uv.sh
```diff
@@ -6,10 +6,10 @@ DIR="$(dirname "$(realpath "$0")")"
...
-mkdir -p resources/contentCreator/createDevfile
-ln -fs "$creator_resources_path/common/devfile/devfile.yaml.j2" "resources/contentCreator/createDevfile/devfile-template.txt"
-ln -fs "$creator_resources_path/common/devcontainer/.devcontainer" "resources/contentCreator/createDevcontainer/"
+mkdir -p out/resources/contentCreator/createDevcontainer
+mkdir -p out/resources/contentCreator/createDevfile
+ln -fs "$creator_resources_path/common/devfile/devfile.yaml.j2" "out/resources/contentCreator/createDevfile/devfile-template.txt"
+ln -fs "$creator_resources_path/common/devcontainer/.devcontainer" "out/resources/contentCreator/createDevcontainer/"
 
 log notice "Using $(python3 --version) from $(uv run which python3)"
 if [[ "$(which python3)" != ${VIRTUAL_ENV}/bin/python3 ]]; then
```
goneri, line 15: `$VIRTUAL_ENV` also points outside of the source directory:

`./.config/mise.toml:VIRTUAL_ENV = "{{ xdg_cache_home }}/.local/share/virtualenvs/vsa"`

goneri (changes requested):

.github/workflows/ci.yaml
```diff
@@ -437,6 +448,10 @@ jobs:
           find out -name '*"*' -exec rm -r {} \; || true
           find out -name '*:*' -exec rm -r {} \; || true
           rm -rf dist
+          # Remove container overlay files to avoid gitleaks false positives
+          sudo rm -rf out/als/tmp/home/.local/share/containers/ || true
```
goneri: Does this prevent the CI from caching the images? Maybe we can set-up some symlink instead to continue to safe the images are at place that will be restored across builds?

packages/ansible-language-server/test/globalSetup.ts
rockygeekz: Yeah you are right. This is aggressive and I was trying to clean up leftover containers from failed test runs and I have removed those aggressive changes if CI fails I'll handle it by cleanup or something.

packages/ansible-language-server/test/globalSetup.ts
rockygeekz: Fixed :+1:

tools/uv.sh
rockygeekz: Yeah, `VIRTUAL_ENV` does point outside the repo, but that's intentional. It's managed by mise (not by our build system), and moving it inside out/ would break mise's venv management and force reinstalling all Python deps on every clean. The goal of this PR is to prevent build artifacts and temp test files from polluting `HOME/repo root` - not to isolate the Python environment itself. The venv is already properly isolated per-project by mise/uv.

.github/workflows/ci.yaml
rockygeekz: and yeah about caching. However, we can't use symlinks here because:
1. It would defeat the isolation purpose - we'd be writing to user's actual container storage
2. Container images are already cached at system level by podman/docker (in `/var/lib/containers/` or `~/.local/share/containers/` on the runner), not in the workspace 
3. We're only removing the overlay files to avoid gitleaks false positives (they contain strings that look like API keys) The actual image layers are still cached at system level, we're just cleaning up the overlay storage used during test runs.

rockygeekz:
@coderabbitai review

rockygeekz:
@coderabbitai resume

packages/ansible-language-server/test/globalSetup.ts
```diff
@@ -66,6 +67,44 @@ function execWithTimeout(
 }
 
 export async function setup() {
+  // Use shared HOME for all tests to prevent writing to user's home directory
+  const originalHome = process.env.HOME || process.env.USERPROFILE || "";
+  const sharedHome = path.resolve(__dirname, "../../../out/home");
```
goneri: Can you use `import.meta.dirname` (ESM) here instead of `__dirname` like for the rest of the PR?

packages/ansible-language-server/test/globalSetup.ts
```diff
@@ -196,5 +235,9 @@ export async function setup() {
 }
 
 export async function teardown() {
-  // Cleanup if needed
```
goneri, line 199: These comments don't bring any value IMO. Can you just remove the `teardown()` function?

goneri (approved):

PR #2747, opened 2026-04-16
fix(ci): bump e2e download timeout to 60s

cidrblock:
## Summary

- Patches `@vscode/test-cli@0.0.12` to pass the full options object (including `timeout`) to `downloadAndUnzipVSCode` instead of positional args that drop the timeout
- Sets `download.timeout` to 60s in `.vscode-test.mjs` as a belt-and-suspenders defense

## Root cause

`@vscode/test-cli`'s `installDependentExtensions()` in `desktop.mjs` calls:

```js
electron.downloadAndUnzipVSCode(opts.version, opts.platform, opts.reporter);
```

This uses positional args, so `opts.timeout` (from `baseCliOptions()`) is never forwarded. `@vscode/test-electron` defaults to 15s, which is too short on CI runners under load.

The patch changes the call to `electron.downloadAndUnzipVSCode(opts)`, which passes the object form and preserves `timeout`.

Upstream bug: microsoft/vscode-test-cli#106

## Test plan

- [ ] CI `e2e` step passes on Linux without `@vscode/test-electron request timeout out after 15000ms`
- [ ] No regressions on macOS/Windows e2e jobs

### DIFF
--- a/.vscode-test.mjs
+++ b/.vscode-test.mjs
@@ -10,6 +10,7 @@ export default defineConfig({
   files: "test/e2e/**/*.test.ts",
   extensionDevelopmentPath: ".", // package.json location
   workspaceFolder: "test/testFixtures",
+  download: { timeout: 60_000 },
   launchArgs: [
     // cannot rely on vscode-test ability to install extensions because it does
     // implicitly use --force and the server can respond with 503 errors.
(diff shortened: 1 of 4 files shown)


cidrblock:
Addressed both review comments:

- Regenerated the patch file with real blob hashes (`b901849..50b5bcc`) computed via `git hash-object` against the original and patched `desktop.mjs`. The placeholder `index 1234567..abcdefg` line is gone.
- Also fixed the commit message to use `related:` instead of `closes:` (it's an upstream issue, not ours).

cidrblock:
e2e fix, see summary

PR #2773, opened 2026-04-27
fix: resolve devfile and devcontainer templates from ansible_creator package

anusshukla:
Embedded devfile and devcontainer templates directly in the extension code so these webviews work without ansible-creator installed. Removed the out/resources/ symlink dependency entirely. This addresses the review feedback that these features cannot depend on the Python package.
related: #2762

Embedded devfile and devcontainer templates directly into the extension code to eliminate dependency on the ansible-creator Python package. Static template maps are now inline within the WebviewMessageHandlers class, with template variable substitution and file generation handled internally without requiring external resources via extensionUri.

- Embedded static DEVCONTAINER_TEMPLATES map in WebviewMessageHandlers class for inline template substitution and .json file generation
- Embedded static DEVFILE_TEMPLATE in WebviewMessageHandlers class for inline template substitution and YAML file generation
- Removed extensionUri parameter from runDevcontainerCreateProcess and runDevfileCreateProcess method signatures
- Updated test suite to validate embedded template behavior instead of external filesystem template dependencies
- Updated examples/devfile.yaml metadata identifier

related: #2762

### DIFF
--- a/src/features/lightspeed/vue/views/webviewMessageHandlers.ts
+++ b/src/features/lightspeed/vue/views/webviewMessageHandlers.ts
@@ -260,7 +260,6 @@ export class WebviewMessageHandlers {
   private async handleInitCreateDevcontainer(
     message: any,
     webview: vscode.Webview,
-    _context: vscode.ExtensionContext,
   ) {
     const payload = message.payload as DevcontainerFormInterface;
     await this.runDevcontainerCreateProcess(payload, webview);
@@ -269,7 +268,6 @@ export class WebviewMessageHandlers {
   private async handleInitCreateDevfile(
     message: any,
     webview: vscode.Webview,
-    _context: vscode.ExtensionContext,
   ) {
     const payload = message.payload as DevfileFormInterface;
     await this.runDevfileCreateProcess(payload, webview);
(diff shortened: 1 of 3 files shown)


alisonlhart (changes requested):
@anusshukla These webviews are intended to work without having ansible-creator installed, so they cannot be derived from the installed package.

anusshukla:
> @anusshukla These webviews are intended to work without having ansible-creator installed, so they cannot be derived from the installed package.

@alisonlhart made the changes

src/features/lightspeed/vue/views/webviewMessageHandlers.ts
```diff
@@ -726,11 +718,110 @@ export class WebviewMessageHandlers {
     return DevcontainerRecommendedExtensions.RECOMMENDED_EXTENSIONS;
   }
 
+  private static readonly DEVCONTAINER_TEMPLATES: Record<string, string> = {
```
alisonlhart, line 721: We should have this defined in a file so we can easily check it against the ansible-creator file and verify the contents in general. I'm not sure if that check was removed...

alisonlhart:
@anusshukla I found the fix for this! In #2757 the files were identified as build artifacts, but they are supposed to be accessible in the final vsix. I fixed this in my change #2763 . We can close this PR since it is resolved there.

PR #2777, opened 2026-04-30
fix(mcp): copy resource data files to build output directory

shatakshiiii:
# Fix: MCP server resource files missing from VSIX

## The Problem

The MCP server has 4 data files that tools need at runtime:

- `agents.md` — best practices guidelines used by `ansible_content_best_practices`
- `ee-rules.md` — execution environment rules used by `define_and_build_execution_env`
- `execution-environment-schema.json` — JSON schema for EE validation
- `execution-environment-sample.yml` — sample EE config

These files live at `packages/ansible-mcp-server/src/resources/data/`.

**How the code finds them:** When a tool like `define_and_build_execution_env` runs, it calls `resolveResourcePath("ee-rules.md", import.meta.url)` in `src/utils/resourcePath.ts`. This function figures out which directory the running code lives in (using `__dirname` for CJS or `import.meta.url` for ESM), then appends `/data/ee-rules.md` to that directory.

**What happens in the packaged extension:** The build tool (tsup) bundles all TypeScript into `dist/cli.cjs`. When this bundled file runs, `__dirname` resolves to `dist/`. So the code looks for `dist/data/ee-rules.md`.

**But `dist/data/` didn't exist.** tsup only compiles `.ts` files — it has no built-in mechanism to copy non-code assets like `.md` or `.json` files. The data files stayed in `src/resources/data/` and were never copied to the build output.

The `.vscodeignore` had a line `!packages/ansible-mcp-server/src/resources/data/**/*` which did ship the source data files inside the VSIX — but it didn't matter because the runtime code never looks in `src/resources/data/`. It only looks relative to where the compiled code runs from (`dist/`).

## The Fix

**1. `packages/ansible-mcp-server/tsup.config.ts`** — Added an `onSuccess` callback that runs after each tsup build and copies the data files to the right location:

- Production build (`dist/`): copies to `dist/data/` — because the bundled `cli.cjs` has `__dirname = dist/`
- Dev build (`lib/`): copies to `lib/resources/data/` — because non-bundled files like `lib/resources/eeSchema.js` have `__dirname = lib/resources/`

The `onSuccess` hook runs after tsup finishes (and after `clean: true` wipes the output dir), so the copied files survive in the final output.

**2. `.vscodeignore`** — Removed the `!packages/ansible-mcp-server/src/resources/data/**/*` line. Since data files are now properly in `dist/data/`, the existing `!packages/ansible-mcp-server/dist/**/*` line already includes them. This avoids shipping duplicate copies of the data files in the VSIX.

**3. `packages/ansible-mcp-server/package.json`** — Removed `"./src/resources/data/**/*"` from the `files` array. Same reasoning — `"dist"` and `"lib"` already cover the data files for npm publishing now.

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -25,7 +25,6 @@
 !packages/ansible-language-server/dist/cli.cjs
 !packages/ansible-mcp-server/dist/**/*
 !packages/ansible-mcp-server/package.json
-!packages/ansible-mcp-server/src/resources/data/**/*
 !package.json
 !package.nls.json
 !syntaxes/**/*
--- a/packages/ansible-mcp-server/package.json
+++ b/packages/ansible-mcp-server/package.json
@@ -23,8 +23,7 @@
   },
   "files": [
     "lib",
-    "dist",
-    "./src/resources/data/**/*"
+    "dist"
   ],
   "keywords": [
     "mcp",
--- a/packages/ansible-mcp-server/tsup.config.ts
+++ b/packages/ansible-mcp-server/tsup.config.ts
@@ -1,7 +1,9 @@
 import type { Options } from "tsup";
+import { cpSync } from "node:fs";
 import { builtinModules } from "node:module";
 
 const env = process.env.NODE_ENV;
+const outDir = env === "production" ? "dist" : "lib";
 
 export const tsup: Options = {
   clean: true,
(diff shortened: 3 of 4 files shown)


alisonlhart (approved):
LGTM, tested locally and is working!

PR #2848, opened 2026-06-09
fix: @typescript-eslint/restrict-template-expressions

ssbarnea:
Enables the `@typescript-eslint/restrict-template-expressions` ESLint rule to enforce type-safe template string formatting. Updates throughout the codebase to comply by explicitly converting values to strings, safely extracting error messages with instanceof checks, and replacing unsafe type casts with proper string coercion.

- Enable `@typescript-eslint/restrict-template-expressions` rule at error level in eslint.config.mjs
- Convert error values to strings using `error instanceof Error ? error.message : String(error)` pattern across error handling paths
- Add explicit String() conversions in template expressions where values may not be strings
- Replace `Object.keys(...as object)` with safer `getEnumerableKeys()` helper in completionProviderUtils.ts
- Update array logging to use `.join(", ")` instead of relying on default array-to-string conversion
- Improve type safety in webviewUtils.ts by treating message data as unknown and adding helpers to coerce to strings
- Add escapeHtml() helper in contentMatchesWebview.ts to prevent HTML injection
- Strengthen validation in ansibleContext.ts and outlineGenerator.ts to explicitly check objects before accessing properties
- Update formatAnsibleMetaData.ts to defensively normalize nested objects using optional chaining
- Make WebviewMessage.payload optional and add payload validation in message handlers

Related: AAP-66052

### DIFF
--- a/src/features/utils/formatAnsibleMetaData.ts
+++ b/src/features/utils/formatAnsibleMetaData.ts
@@ -3,6 +3,11 @@ import { MarkdownString, workspace } from "vscode";
 import * as os from "os";
 import * as path from "path";
 
+const asRecord = (value: unknown): Record<string, unknown> =>
+  value !== null && typeof value === "object" && !Array.isArray(value)
+    ? (value as Record<string, unknown>)
+    : {};
+
 export function formatAnsibleMetaData(ansibleMetaData: any) {
   let mdString = "";
   let ansiblePresent = true;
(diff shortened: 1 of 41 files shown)


src/features/utils/formatAnsibleMetaData.ts
```diff
@@ -76,18 +81,18 @@ export function formatAnsibleMetaData(ansibleMetaData: any) {
...
         if (value != null && String(value).trim().toLowerCase() !== "nil") {
-          mdString += ` <span ${WARNING_STYLE}>${value}</span>`;
+          mdString += ` <span ${WARNING_STYLE}>${String(value)}</span>`;
         }
         return;
       }
       mdString += `\n   - ${key}: `;
-      const value = valueObj[key];
-      if (typeof value === "object") {
+      const value = valueObj[key] as any;
```
sudhirverma, line 94: can we avoid using any as type

ssbarnea: We have linters such eslint to notify us about these but there are cases where this cannot be avoided. Also this is outside the narrow scope of this change: address one specific eslint issue, and considering that eslint passed we are fine. Also note that all the changes were made by Claude.

AFAIK use of any is already checked by eslint but if is not it will be activated and fixed as part of a follow-up. The jira item specifies on eslint lint fixed per PR (progressive).

sudhirverma (approved):

ssbarnea:
@sudhirverma Due to the nature of the change I am also inclined to ignore sonar quality gate because is again unrelated to this 'bulk' change. Addressing cyclomatic complexity should be done outside this PR to keep it simple and easy to review. Usually we don't want to change the logic when we address a rule.

PR #2895, opened 2026-06-18
Integrate Abbenay as LLM provider with status bar icon

shatakshiiii:
## Summary

This PR integrates the [Abbenay](https://github.com/redhat-developer/abbenay) LLM provider extension into the Ansible VS Code extension, replacing all references to the deprecated "Open LLM Provider" (`openllm`).

### What changed

- **Abbenay as recommended extension** — Added `redhat.abbenay-provider` to `extensionRecommendations` in `package.json`, so VS Code suggests it to users alongside the Ansible extension. It is intentionally a *recommendation* rather than a hard dependency since Abbenay is not yet published on all platforms (e.g. Windows).

- **Status bar icon for LLM configuration** — Added an Ansible logo icon (`$(ansible-logo)`) to the VS Code status bar. Clicking it activates the Abbenay extension and opens its LLM provider configuration webview (`abbenay.configureProvider`), where users can set up their preferred LLM providers and API keys. If Abbenay is not installed, a warning message is shown instead.

- **Chat provider setting updated** — The `ansibleEnvironments.llm.chatProvider` setting now offers `vscode` and `abbenay` as options (replacing `openllm`). When `abbenay` is selected, AI prompts focus the Abbenay chat sidebar and copy the prompt to clipboard.

- **All `openllm` references replaced** — Updated enum values, descriptions, markdown documentation, and error messages across `package.json`, `src/extension.ts`, `src/services/LlmService.ts`, and `README.md` to reference Abbenay instead of Open LLM Provider.

### Test plan

- [ ] Verify the Ansible logo icon appears in the status bar
- [ ] Click the icon with Abbenay installed — should open LLM provider config webview
- [ ] Click the icon without Abbenay installed — should show a warning message
- [ ] Set `llm.chatProvider` to `abbenay` and trigger an AI prompt — should focus Abbenay chat and copy prompt to clipboard
- [ ] Set `llm.chatProvider` to `vscode` — should use VS Code's built-in chat as before
- [ ] CI passes on all platforms (Windows, WSL, Linux)

JIRA - https://redhat.atlassian.net/browse/AAP-78572

### DIFF
--- a/package.json
+++ b/package.json
@@ -617,6 +617,12 @@
         "icon": "$(info)"
       },
       {
+        "command": "ansibleEnvironments.configureLlmProvider",
+        "title": "Configure LLM Provider",
+        "category": "Ansible Environments",
+        "icon": "$(ansible-logo)"
+      },
+      {
         "command": "ansible.statusBar.pythonClick",
         "title": "Python Environment Details",
         "category": "Ansible"
--- a/package.json
+++ b/package.json
@@ -1056,11 +1056,11 @@
           "default": "vscode",
           "enum": [
             "vscode",
-            "openllm"
+            "abbenay"
           ],
           "enumDescriptions": [
             "Use VS Code's built-in chat (GitHub Copilot Chat)",
-            "Use Open LLM Provider's chat panel"
+            "Use Abbenay's chat panel"
           ],
           "markdownDescription": "Which chat UI to use for AI features.",
           "scope": "resource",
@@ -1074,7 +1074,7 @@
         "ansibleEnvironments.llm.provider": {
           "type": "string",
           "default": "",
-          "markdownDescription": "*(Advanced)* LLM provider/vendor (e.g., `copilot`, `open-llm`). Leave empty for auto-selection.",
+          "markdownDescription": "*(Advanced)* LLM provider/vendor (e.g., `copilot`, `abbenay`). Leave empty for auto-selection.",
           "scope": "resource",
           "order": 3
         },
--- a/package.json
+++ b/package.json
@@ -30,7 +30,8 @@
   },
   "extensionDependencies": [
     "ms-python.python",
-    "redhat.vscode-yaml"
+    "redhat.vscode-yaml",
+    "redhat.abbenay-provider"
   ],
   "extensionRecommendations": [
     "ms-python.vscode-python-envs"
--- a/package.json
+++ b/package.json
@@ -30,11 +30,11 @@
   },
   "extensionDependencies": [
     "ms-python.python",
-    "redhat.vscode-yaml",
-    "redhat.abbenay-provider"
+    "redhat.vscode-yaml"
   ],
   "extensionRecommendations": [
-    "ms-python.vscode-python-envs"
+    "ms-python.vscode-python-envs",
+    "redhat.abbenay-provider"
   ],
   "contributes": {
     "icons": {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1409,6 +1409,33 @@ export function activate(context: vscode.ExtensionContext) {
         },
     );
 
+    // Register Abbenay LLM provider command and status bar item
+    const openLlmProviderCommand = vscode.commands.registerCommand(
+        'ansibleEnvironments.openLlmProvider',
+        async () => {
+            const ext = vscode.extensions.getExtension('redhat.abbenay-provider');
+            if (!ext) {
+                vscode.window.showWarningMessage(
+                    'Abbenay extension not found. Install it to configure LLM providers.',
+                );
+                return;
+            }
+            if (!ext.isActive) {
+                await ext.activate();
+            }
+            await vscode.commands.executeCommand('abbenay.configureProvider');
+        },
+    );
+
+    const ansibleStatusBarItem = vscode.window.createStatusBarItem(
+        vscode.StatusBarAlignment.Right,
+        100,
+    );
+    ansibleStatusBarItem.text = '$(ansible-logo)';
+    ansibleStatusBarItem.tooltip = 'Configure LLM Provider';
+    ansibleStatusBarItem.command = 'ansibleEnvironments.openLlmProvider';
+    ansibleStatusBarItem.show();
+
     context.subscriptions.push(
         refreshCommand,
         installCommand,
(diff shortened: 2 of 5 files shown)


src/extension.ts
```diff
@@ -1421,6 +1409,33 @@ export function activate(context: vscode.ExtensionContext) {
...
+    );
+
+    const ansibleStatusBarItem = vscode.window.createStatusBarItem(
+        vscode.StatusBarAlignment.Right,
+        100,
+    );
+    ansibleStatusBarItem.text = '$(ansible-logo)';
+    ansibleStatusBarItem.tooltip = 'Configure LLM Provider';
+    ansibleStatusBarItem.command = 'ansibleEnvironments.openLlmProvider';
+    ansibleStatusBarItem.show();
```
rockygeekz, line 1437: The status bar item (`ansibleStatusBarItem.show()`) is called unconditionally — users without Abbenay installed will see the icon but get a "not found" warning when they click it. Should we gate it behind a check like `if (vscode.extensions.getExtension('redhat.abbenay-provider'))` before showing?

package.json
```diff
@@ -1056,11 +1057,11 @@
           "default": "vscode",
           "enum": [
             "vscode",
-            "openllm"
+            "abbenay"
```
Hrithik-Gavankar, line 1066: **Breaking change without migration path**

  The chat provider enum changed from `["vscode", "openllm"]` to `["vscode", "abbenay"]`, which breaks existing user configurations.
  
  **Impact:**
  Users who have `"ansibleEnvironments.llm.chatProvider": "openllm"` in their settings.json will experience silent failure—the setting becomes invalid and VS Code falls back to "vscode" without warning.
  
  **Recommendation:**
  Add migration logic in the `activate()` function to automatically update existing "openllm" values to "abbenay":
  
  ```typescript
  const config = vscode.workspace.getConfiguration('ansibleEnvironments');
  const chatProvider = config.get<string>('llm.chatProvider');
  if (chatProvider === 'openllm') {
      await config.update('llm.chatProvider', 'abbenay', vscode.ConfigurationTarget.Global);
      vscode.window.showInformationMessage(
          'Updated chat provider from deprecated "openllm" to "abbenay".'
      );
  }
  
  This ensures a smooth upgrade experience for existing users.

src/extension.ts
```diff
@@ -1421,6 +1409,33 @@ export function activate(context: vscode.ExtensionContext) {
         },
     );
 
+    // Register Abbenay LLM provider command and status bar item
+    const openLlmProviderCommand = vscode.commands.registerCommand(
```
Hrithik-Gavankar: **Missing command declaration + naming inconsistency**
  
  Two issues with this command registration:

  **1. Missing from package.json manifest**
  The command `ansibleEnvironments.openLlmProvider` is registered here but missing from `contributes.commands` in package.json. VS Code requires all commands to be declared in the manifest for Command Palette
  visibility and proper validation.
  
  **Impact:** Status bar icon appears but clicking it may fail silently.
  
  **2. Command naming doesn't follow conventions**
  - Other LLM commands use `selectLlmModel`, `showLlmStatus` (with "Llm" casing, not "LLM")
  - The tooltip says "**Configure** LLM Provider" but the command says "**open**"
  - Implementation calls `abbenay.configureProvider` (configure action, not open)
  
  **Recommended fix:**
  1. Rename to `configureLlmProvider` for consistency and accuracy
  2. Add to package.json `contributes.commands`:
  ```json
  {
    "command": "ansibleEnvironments.configureLlmProvider",
    "category": "Ansible",
    "title": "Configure LLM Provider"
  }
  
  Also update the command reference in the status bar item (line ~1437) and subscriptions (line ~1492).

Hrithik-Gavankar (changes requested):

package.json
ganeshrn: SInce openllm was not released, we can ignore this migration path.

ganeshrn (commented):
Any specific reason to keep this an optional dependency?

src/extension.ts
shatakshiiii: @rockygeekz Good point! There are two valid approaches here:

Option A — Always show (current behavior) The icon is always visible regardless of whether Abbenay is installed. Clicking it without Abbenay shows a "not found" warning with guidance to install it. This serves as a discovery mechanism — users see the Ansible logo in the status bar and learn about the Abbenay LLM provider integration. This is similar to how some extensions show UI elements that prompt users to install companion extensions.

Option B — Gate behind extension check Only show the icon when vscode.extensions.getExtension('redhat.abbenay-provider') returns a result. This provides a cleaner UX for users who don't have or don't want Abbenay, avoiding a potentially confusing icon that leads to a warning. The downside is reduced discoverability.

Keeping this open for discussion — would appreciate input from other reviewers on which UX direction we prefer.

shatakshiiii:
> Any specific reason to keep this an optional dependency?

Yes — Abbenay is not yet published on the VS Code Marketplace for all platforms (e.g. Windows). So making it an extensionDependency (hard dependency) will:

1. Break extension activation on Windows — VS Code refuses to activate an extension if any extensionDependency can't be installed. This would make the entire Ansible extension non-functional on Windows.

1. Break CI — The @vscode/test-cli integration test runner automatically installs all extensionDependencies before running tests. On Windows runners, this fails with: "The 'redhat.abbenay-provider' extension is not available in Visual Studio Code for the Windows 64 bit platform." (We hit this exact issue during CI — see commits fc61dd3 and a410ad6.)

The code in extension.ts already handles the missing-extension case gracefully — if Abbenay isn't installed, clicking the status bar icon shows a warning message instead of crashing. Once Abbenay is published on all platforms, we can promote it to extensionDependencies.

src/extension.ts
shatakshiiii: done

Hrithik-Gavankar (approved):
LGTM!

ganeshrn:
Adding a follow-up comment here since the PR is already merged.

Regarding the concern `Break extension activation on Windows — VS Code refuses to activate an extension if any extensionDependency can't be installed. This would make the entire Ansible extension non-functional on Windows.`

This concern isn't valid in the Ansible extension's case as the Ansible extension is supported on Windows WSL environments and not on plain Windows, so this does not apply.

Issue #2792, opened 2026-05-11
Python environment discovery is degraded (PET binary missing). This commonly occurs in OpenVSX-based editors (Dev Spaces, VSCodium).


tertek:
### Summary

**Using VS Codium**

When opening VS Codium or restarting extenstions hosts, the following Warning will be shown:

<img width="473" height="147" alt="Image" src="https://github.com/user-attachments/assets/1a4c2faa-6ee2-4a82-8848-8b054a219993" />

More, links to: https://github.com/microsoft/vscode-python/issues/25820

### Extension version

26.4.6

### VS Code version

on: 1.116.02821 Commit: 221e0a382c0be3a673a4e4cab0601344a0b3de3a Date: 2026-04-15T00:28:13Z Electron: 39.8.7 ElectronBuildId: undefined Chromium: 142.0.7444.265 Node.js: 22.22.1 V8: 14.2.231.22-electron.0 OS: Linux x64 6.17.0-23-generic

### Ansible Version

```shell
ansible [core 2.20.5]
```

### OS / Environment

Ubuntu 24.04

### Relevant log output

```shell

```

PR #2899, opened 2026-06-18
fix: suppress repeated PET warning using globalState and Don't show again

gy-gl:
## Summary

On OpenVSX-based editors (VSCodium, Dev Spaces) the PET binary is absent from the `ms-python.vscode-python-envs` universal build, so the warning fired on every extension host restart. The `_petWarningShown` flag was in-memory and reset each time the extension host restarted.

- Pass `ExtensionContext` into `PythonEnvironmentService.initialize()` so the service can read/write `globalState`
- Replace the unconditional once-per-process suppression with an explicit **Don't show again** button that persists the decision to `globalState`; users who dismiss without clicking continue to see the warning on subsequent restarts until they actively choose to suppress it
- Add tests covering the new `globalState` path: persist on click, no persist on dismiss, suppress when already set

The root cause of #2792 is upstream - OpenVSX distributes a universal VSIX for `ms-python.vscode-python-envs` that omits the platform-specific `pet` binary. The extension already handles this correctly by falling back to `ms-python.python`. This change addresses the UX symptom: the repeated warning popup on every restart.

Relates to #2792

Merge branch 'main' into gy/fix-pet-warning-globalstate: Persist PET-missing Python environment warning suppression across extension-host restarts by passing ExtensionContext into PythonEnvironmentService.initialize(). Persist warning suppression across restarts via ExtensionContext.globalState; “Don’t show again” writes ansible.petWarningShown, dismiss leaves it unset. Tests verify persistence, dismiss, and pre-set suppression.
- Pass context to pythonEnvService.initialize
- Persist/skip PET warning via globalState
- Add unit tests for persistence/dismiss/pre-set

Related: `#2792`

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -109,7 +109,7 @@ export async function activate(context: ExtensionContext): Promise<void> {
 
   // Initialize Python Environment Service early
   const pythonEnvService = PythonEnvironmentService.getInstance();
-  await pythonEnvService.initialize();
+  await pythonEnvService.initialize(context);
   context.subscriptions.push(pythonEnvService);
 
   // Initialize Terminal Service
--- a/src/services/PythonEnvironmentService.ts
+++ b/src/services/PythonEnvironmentService.ts
@@ -33,6 +33,7 @@ export class PythonEnvironmentService implements vscode.Disposable {
   private _pythonExtApi: PythonExtension | undefined;
   private _initialized: boolean = false;
   private _petWarningShown: boolean = false;
+  private _context: vscode.ExtensionContext | undefined;
   private _disposables: vscode.Disposable[] = [];
 
   private _onDidChangeEnvironment =
@@ -53,11 +54,17 @@ export class PythonEnvironmentService implements vscode.Disposable {
    * Initialize the service. Tries the Environments extension first; if PET
    * is missing falls back to the main Python extension's environments API.
    */
-  public async initialize(): Promise<boolean> {
+  public async initialize(
+    context?: vscode.ExtensionContext,
+  ): Promise<boolean> {
     if (this._initialized) {
       return (
         this._pythonEnvApi !== undefined || this._pythonExtApi !== undefined
       );
+    }
+
+    if (context) {
+      this._context = context;
     }
 
     console.log(
@@ -238,9 +245,15 @@ export class PythonEnvironmentService implements vscode.Disposable {
   // ---------------------------------------------------------------------------
 
   private _showPetWarning(): void {
-    if (this._petWarningShown) {
+    const alreadyShown = this._context
+      ? this._context.globalState.get<boolean>("ansible.petWarningShown") ===
+        true
+      : this._petWarningShown;
+
+    if (alreadyShown) {
       return;
     }
+
     this._petWarningShown = true;
 
     const useEnvsSetting = vscode.workspace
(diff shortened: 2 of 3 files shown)


ssbarnea (approved):

ssbarnea (changes requested):
- Ensure that added code is covered (coderabit requires >90% on new code)
- Ensure that all commits are signed, if not we cannot merged it (you might want to squash them too).

Once you do these we will be able to merge this great addition. I will maker the PR as draft until all CI is reporting green, helps with reviewing queue.

gy-gl:
Thanks @ssbarnea. I believe the testing threshold is met and the commits are now all signed.

ssbarnea (dismissed):

ssbarnea:
@gygitlab please try to address tests that are failing and comments from coderabbit.

gy-gl:
Addressed the code review feedback:

- **Duplicate-warning guard**: Split the combined `alreadyShown` check into two separate early returns - the in-memory `_petWarningShown` flag is checked first, then `globalState`. Previously, when a context was present the in-memory flag was ignored, meaning rapid concurrent calls could show duplicate dialogs before the user persisted suppression.
- **Floating promise**: Added `void` prefix to the `showWarningMessage` call to make the intent explicit.

Regarding the `test (macos)` wdio failure (`commands.spec.ts` - "Playbook content should not be empty"): this test exercises `ansible.create-empty-playbook` which has no interaction with the code changed in this PR. The same test passed on the `main` branch run that ran concurrently, so this appears to be a timing-sensitive flake rather than a regression introduced here.

shvenkat-rh:
@gy-gl can you please address the codecov and sonar cloud issues

gy-gl:
Added three tests to raise patch coverage:

- In-session deduplication: resets `_initialized` on the live instance and calls `initialize()` again to exercise the `_petWarningShown` early-return branch that only fires when the warning has already been shown in the same session.
- "Disable Environments Extension" selection: mocks the dialog returning that button and verifies the `useEnvironmentsExtension` config update is called.
- "Reload Now" follow-up: verifies `workbench.action.reloadWindow` is executed when the user confirms the reload prompt after disabling the setting.

These cover the `} else if (selection === "Disable Environments Extension")` branch and the in-memory `_petWarningShown` guard that were missing from the patch.

gy-gl:
Pushed additional test coverage to address the SonarCloud quality gate (was failing on new-code coverage: 57.3% vs the 90% threshold, plus one minor `S5906` finding):

- Added tests exercising the catch blocks in `getEnvironmentFromPythonExt` and `getEnvironmentsFromPythonExt` (previously untested error paths)
- Switched a generic `.length` comparison to `toHaveLength()` per the SonarCloud suggestion

Latest CI run shows failures unrelated to this PR's diff:
- `test (macos)` / `test (wsl)`: both fail on `ENOTDIR: not a directory ... als-cli-docs-fail-.../not-a-dir/settings.md` inside `packages/ansible-language-server/test/cli.test.ts`, plus separate infra issues (Chrome/Electron failing to spawn on macOS, a WSL permissions error, a dbus spawn error) - none of these touch `PythonEnvironmentService.ts` or the test file we changed.
- `test (linux)`: lint now passes; `task build / package` was cancelled as a side effect of the other jobs failing in the same run, not a failure of its own.

I don't have permission to re-run CI on this repo - could someone with access kick off a re-run? Everything in our diff (lint, our unit tests, coverage on the changed code) is green.

shvenkat-rh:
Looks like codecov is still failing

gy-gl:
It appears codecov is failing because `test (macos)` hits a real Electron spawn error on that runner, which trips `fail-fast: true` and cancels `test (linux)` mid-run before it can upload coverage - not an actual coverage regression. WSL and everything in our diff passed. Could you re-run, or check why macOS can't spawn Electron?

shvenkat-rh:
Both the tests (macos and linux) are passing now

shvenkat-rh:
Closing the PR since there was no response

gy-gl:
@shvenkat-rh can you clarify what was needing a response? My understanding was that the CI was passing and this was ready to be merged?

PR #2925, opened 2026-06-23
chore(ci): replace self-hosted WSL runner with GitHub-hosted

sudhirverma:
## Summary

Migrate the WSL CI test job from self-hosted WSL runners to GitHub-managed `windows-latest`, removing self-hosted-specific cleanup/guard logic. This simplifies WSL activation, bootstraps WSL toolchains with cached setup, installs dependencies inside WSL, and copies artifacts back to Windows for reporting.

- Replace self-hosted/wsl-runner with `windows-latest` using a `wsl-bash` shell
- Remove self-hosted-dependent checkout `clean:` behavior and Podman overlay/workspace cleanup steps
- Simplify WSL activation to trigger when the matrix name contains `wsl`; extend WSL `additional-packages` with `*-t64` libs and adjust `WSLENV`
- Add cached WSL bootstrap/tooling steps, persist PATH via `.bashrc`, copy the repo into native WSL at `$HOME/workspace`, and run `pnpm install` inside WSL
- Copy WSL test outputs back to `$GITHUB_WORKSPACE` and broaden log/report upload conditions to Windows `wsl` matrices
- Expand `.config/dictionary.txt` `lib*` entries and add `zstd`
- Mark WSL job `continue-on-error` while pipeline hardens

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,7 +33,7 @@ concurrency:
 env:
   # https://docs.github.com/en/actions/learn-github-actions/environment-variables
   # https://devblogs.microsoft.com/commandline/share-environment-vars-between-wsl-and-windows/
-  WSLENV: HOSTNAME:CI:FORCE_COLOR:GITHUB_ACTION:GITHUB_ACTION_PATH/p:GITHUB_ACTION_REPOSITORY:GITHUB_WORKFLOW:GITHUB_WORKSPACE/p:GITHUB_PATH/p:GITHUB_ENV/p:VIRTUAL_ENV/p:SKIP_PODMAN:SKIP_DOCKER:NODE_OPTIONS:MISE_ENV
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -292,48 +292,18 @@ jobs:
...
             id: test-wsl
             task-name: test
-            os: wsl-runner
-            runs-on: self-hosted
+            os: windows-latest
+            shell: "wsl-bash {0}"
             env:
               SKIP_PODMAN: 1
               SKIP_DOCKER: 1
+            continue-on-error: true
```
rockygeekz: Are we allowing the CI to pass or ability to merge even this step fails? Can you explain this if this is temporary plan or any other idea?

rockygeekz (changes requested):

.github/workflows/ci.yaml
sudhirverma: removed

rockygeekz (approved):

PR #2949, opened 2026-06-26
fix: enable eslint-plugin-sonarjs plugin

ssbarnea:
Disable high-volume SonarJS rules for follow-up and resolve
straightforward violations so eslint passes cleanly.

Co-authored-by: Cursor <cursoragent@cursor.com>

- fix: enable eslint-plugin-sonarjs plugin — added SonarJS to ESLint, disabled high-volume rules for later follow-up, and fixed a few straightforward lint issues across completion, docs, tests, utilities, and watchers to keep lint passing cleanly.

_AI-assisted._

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,5 +1,5 @@
 // @ts-check
-// cspell: ignore tseslint
+// cspell: ignore tseslint sonarjs
 import eslint from "@eslint/js";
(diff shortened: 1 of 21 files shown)


ssbarnea:
We will skip the coverage variance on these because we will tackle it separately when we review how we test.

PR #3046, opened 2026-07-21
feat(core): port tox-ansible integration from main with full rewrite

Hrithik-Gavankar:
## Summary

- Rewrites the tox-ansible integration for the `next` architecture, addressing all 10 reliability bugs from the `main` branch implementation (AAP-81423)
- Adds `ToxAnsibleService` in `@ansible/developer-services` with JSON-first discovery (`tox list --ansible --gh-matrix`) and proper exit-code-based test verdicts via `CommandService`
- Registers `ToxTestController` (Test Explorer) + `ToxTaskProvider` (Run Task) + MCP tools (`tox_list_environments`, `tox_run_environment`)

## What changed

| Layer | Files |
|-------|-------|
| Types | `packages/common/src/types/toxAnsible.ts` |
| Service | `packages/services/src/ToxAnsibleService.ts` |
| Extension | `src/features/toxAnsible/{ToxTestController,ToxTaskProvider,register}.ts` |
| MCP | `packages/mcp-server/src/{tools,handlers}.ts` |
| Telemetry | `packages/common/src/types/telemetry.ts` (TOX_DISCOVER, TOX_RUN) |
| User stories | `.sdlc/user-stories.yaml` (TOX-001, TOX-002) |
| Tests | Unit tests for service, controller, task provider, MCP handlers; WDIO skeleton |
| Config | `package.json` (taskDefinitions: `ansible-tox`) |

## Key improvements over main

1. **No false passes** — exit code determines pass/fail (main used terminal fire-and-forget)
2. **No stale items** — items cleared before re-discovery
3. **No duplicates** — single CLI discovery path (main had INI parse + env loop)
4. **CommandService** — venv-aware execution (main bypassed it)
5. **Cross-platform** — uses `path.join` (main hardcoded `/bin/`)
6. **Timeout support** — 10 min default per env (main hung forever)
7. **MCP parity** — agents can discover/run tox tests (invariant 9)
8. **JSON discovery** — `--gh-matrix` with `--no-desc` fallback (main parsed INI)

## Test plan

- [x] `pnpm run compile` — clean
- [x] `pnpm run lint:eslint` — passed
- [x] `pnpm exec vitest run` — 77 files, 1287 tests, all green
- [x] `pnpm run build` — all 4 bundles produced
- [ ] Manual: open a collection workspace with tox-ansible → verify Test Explorer populates
- [ ] Manual: run a tox env from Test Explorer → verify pass/fail reporting
- [ ] Manual: "Run Task" → verify tox-ansible environments appear

Reworked tox-ansible integration to discover environments via JSON-first parsing and run tests using exit-code verdicts, with AbortSignal cancellation and a 10-minute timeout to prevent stale items, false passes, and hangs. refactor(tox-ansible): streamline entry parsing and enhance runEnvironments with AbortSignal support.

- Added `ToxAnsibleService` with JSON-first discovery (`--gh-matrix`), `--no-desc` fallback, and signal-aware `runEnvironment(s)` execution
- Implemented tox-ansible Test Explorer controller (categorized discovery, cancellable runs) and Run Task integration (`ansible-tox` task type)
- Added MCP tools to list and run tox-ansible environments with structured results and timeout handling
- Introduced shared tox types and telemetry events, plus user stories and configuration/test coverage
- Registered the new feature in extension activation and added/updated unit, MCP, and UI tests

### DIFF
--- a/packages/services/src/ToxAnsibleService.ts
+++ b/packages/services/src/ToxAnsibleService.ts
@@ -125,7 +125,7 @@ export function parseNoDescOutput(stdout: string): ToxEnvironment[] {
 }
 
 /**
- *
+ * Discovers and runs tox-ansible test environments via the tox CLI.
  */
 export class ToxAnsibleService {
     private readonly _cmd: CommandService;
--- a/packages/services/src/ToxAnsibleService.ts
+++ b/packages/services/src/ToxAnsibleService.ts
@@ -1,0 +1,323 @@
+/**
+ * ToxAnsibleService
+ *
+ * Discovers and runs tox-ansible test environments via the tox CLI.
+ * Uses CommandService for venv-aware execution (ADR-005 invariant 7).
+ *
+ * Discovery uses `tox list --ansible --gh-matrix` for structured JSON
+ * output with pre-split factors, falling back to `--no-desc` text
+ * parsing when --gh-matrix is unavailable.
+ */
+
+import * as fs from 'fs';
+import * as path from 'path';
+import { log } from '@ansible/common';
+import type {
+    ToxEnvironment,
+    ToxTestCategory,
+    ToxRunResult,
+    ToxAvailability,
+    ToxGhMatrixEntry,
+} from '@ansible/common';
+import { CommandService } from './CommandService';
+
+const KNOWN_CATEGORIES = new Set<string>(['integration', 'sanity', 'unit']);
+const TOX_LIST_TIMEOUT_MS = 30_000;
+const TOX_RUN_DEFAULT_TIMEOUT_MS = 10 * 60 * 1000; // 10 min
+
+const CONFIG_FILE_CANDIDATES = ['tox-ansible.ini', 'tox.ini'] as const;
+
+/**
+ * Match tox-ansible environment names in category-first format.
+ * Example: "integration-py3.12-devel"
+ */
+const CATEGORY_FIRST_PATTERN = /^(integration|sanity|unit)-py(\d+\.\d+)-(.+)$/;
+
+/**
+ * Match tox-ansible environment names in python-first format.
+ * Example: "py312-2.17-integration-default"
+ */
+const PYTHON_FIRST_PATTERN = /^py(\d)(\d+)-(\d+\.\d+|devel)-(integration|sanity|unit)(?:-(.+))?$/;
+
+/**
+ * Parse an environment name into category, python version, and ansible version.
+ * Tries multiple patterns and falls back to 'unknown' category.
+ * @param name - Full tox environment name to parse
+ * @returns Parsed components: category, pythonVersion, ansibleVersion
+ */
+export function parseToxEnvironmentName(name: string): {
+    category: ToxTestCategory;
+    pythonVersion: string;
+    ansibleVersion: string;
+} {
+    const catFirst = CATEGORY_FIRST_PATTERN.exec(name);
+    if (catFirst) {
+        return {
+            category: catFirst[1] as ToxTestCategory,
+            pythonVersion: catFirst[2],
+            ansibleVersion: catFirst[3],
+        };
+    }
+
+    const pyFirst = PYTHON_FIRST_PATTERN.exec(name);
+    if (pyFirst) {
+        return {
+            category: pyFirst[4] as ToxTestCategory,
+            pythonVersion: `${pyFirst[1]}.${pyFirst[2]}`,
+            ansibleVersion: pyFirst[3],
+        };
+    }
+
+    return { category: 'unknown', pythonVersion: '', ansibleVersion: '' };
+}
+
+/**
+ * Parse --gh-matrix JSON into ToxEnvironment[].
+ * Extracts category from the factors array.
+ * @param json - Raw JSON string from tox --gh-matrix output
+ * @returns Parsed tox environments, empty array on invalid input
+ */
+export function parseGhMatrixJson(json: string): ToxEnvironment[] {
+    let entries: ToxGhMatrixEntry[];
+    try {
+        entries = JSON.parse(json) as ToxGhMatrixEntry[];
+    } catch {
+        log('ToxAnsibleService: failed to parse --gh-matrix JSON');
+        return [];
+    }
+
+    if (!Array.isArray(entries)) {
+        log('ToxAnsibleService: --gh-matrix output is not an array');
+        return [];
+    }
+
+    return entries.map((entry) => {
+        const categoryFactor = entry.factors.find((f) => KNOWN_CATEGORIES.has(f));
+        const ansibleFactor = entry.factors.find(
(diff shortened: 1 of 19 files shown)


src/features/toxAnsible/ToxTestController.ts
```diff
@@ -0,0 +1,299 @@
...
+                if (excludeSet.has(item)) continue;
+                if (item.children.size > 0) {
+                    item.children.forEach((child) => {
+                        if (!excludeSet.has(child)) items.push(child);
+                    });
+                } else {
+                    items.push(item);
+                }
+            }
+        } else {
```
rockygeekz, line 354: `run.end()` is sitting outside the `try/finally`, so if `runEnvironment` throws the test run just stays stuck as "running" in the UI. Move it inside `finally` after `cancelSub.dispose()`

packages/services/src/ToxAnsibleService.ts
```diff
@@ -0,0 +1,347 @@
...
+
+        const versionOutput = result.stdout;
+        const toxAnsibleInstalled = versionOutput.includes('tox-ansible');
+        const versionMatch = /^(\S+)/.exec(versionOutput);
+
+        return {
+            toxInstalled: true,
+            toxAnsibleInstalled,
+            toxVersion: versionMatch?.[1],
+        };
```
rockygeekz, line 170: This `signal` param doesn't actually do anything right now. `CommandOptions` doesn't have a `signal` field so `runCommandArgs` never gets it, tox just runs to completion regardless. The checks after on L173/L181 only slap a "Cancelled" label on it after the process already finished. Either wire it through to the child process or just remove the param so nobody thinks cancellation actually works

src/features/toxAnsible/ToxTestController.ts
Hrithik-Gavankar: Fixed in commit 18f752f7 — run.end() is now inside the finally block alongside cancelSub.dispose(). Also extracted per-item execution into _executeItem() with its own try/catch so a thrown error marks the item as errored() instead of leaving it stuck.

packages/services/src/ToxAnsibleService.ts
Hrithik-Gavankar: Fixed in commit 18f752f7 — signal is now properly forwarded to the child process. Added signal as the 4th parameter to CommandService.runCommandArgs which passes it to execFileAsync, so the tox process is actually killed on cancellation. Also added explicit AbortError handling in the catch block.

sudhirverma (changes requested):
## Review summary

Thanks for the rewrite — the overall shape looks right (`ToxAnsibleService` + Test Explorer + Task Provider + MCP parity), and automated checks are green:

- Local: vitest **1314** tests, `tsc -b`, eslint on tox paths
- CI: linux (22/24/26), Windows, WSL unit + WDIO all passing

I still think this needs a few correctness/AC fixes before merge. Codecov also shows `ToxTestController` patch coverage ~27%.

### Must-fix (P0)

1. **Abort race can turn a successful run into cancelled**  
   In `ToxAnsibleService.runEnvironment`, after `runCommandArgs` returns, checking `signal?.aborted` and overwriting a completed success drops stdout/stderr and reports cancel. Only treat as cancelled when abort interrupted the child process.

2. **Discovery debounce can hang / drop updates**  
   In `ToxTestController.discoverTests`, `clearTimeout` orphans prior Promise resolves, and `_discovering` early-return can discard a trailing rediscovery from file watchers.

3. **Multi-root workspace cwd collision**  
   `_resolveWorkspaceDir` uses `item.id.startsWith(folder.uri.toString())` and falls back to `folders[0]`. Prefix matches like `file:///workspace` vs `file:///workspace2` can pick the wrong folder.

### Should-fix (P1)

4. **Task Provider bypasses `CommandService`**  
   `ToxTaskProvider` uses bare `ShellExecution('tox', …)` while Test Explorer / MCP resolve tox via `CommandService`. That breaks the PR’s own “venv-aware via CommandService” claim and the todo AC in `.sdlc/todos/pending/port-ansible-tox-integration.md` (discover can work while Run Task fails).

5. **Timeout looks like a normal test failure**  
   Timed-out tox runs surface as `exitCode: 1` / `success: false` with no distinct timeout signal.

6. **TOX-001 empty state unmet**  
   When tox / tox-ansible is missing, the controller only logs and leaves an empty Testing view. Story criterion asks for an empty state.

7. **Telemetry events are unused**  
   `TelemetryEvents.TOX_DISCOVER` / `TOX_RUN` are defined but never emitted.

### Nice-to-have (P2)

- Category-first env parser mangles scenario suffixes (`integration-py3.12-devel-default` → ansibleVersion `devel-default`)
- WDIO tests only assert `taskDefinitions` / extension `isActive` — they don’t cover real discovery/run for TOX-001/002
- Controller unit tests don’t cover discover/run/exclude/cancel paths

### Manual smoke still needed

PR checklist items still unchecked:
- [x] Collection workspace → Test Explorer populates
- [x] Run env from Test Explorer → pass/fail + output
- [x] Run Task → `ansible-tox` environments appear (especially with tox only in the active venv)

Happy to re-review once the P0/P1 items are addressed.

sathyapramod (changes requested):
**packages/services/src/ToxAnsibleService.ts:308**

Abort detection currently conflates the cancellation signal with a non-zero exit code:

```ts
const abortedDuringExec = signal?.aborted && r.exitCode !== 0;
```

If the process fails normally (e.g. `exitCode = 2`) and the signal is aborted after the process has already exited, this condition becomes `true`, causing the actual stderr to be replaced with `"Cancelled by user"` and masking the real failure.

Consider checking whether the process was actually cancelled by `CommandService` instead:

```ts
const abortedDuringExec = signal?.aborted && r.stderr === 'Cancelled';
```

Alternatively, `CommandService` could return a distinct exit code (e.g. `130`) for signal-terminated processes so cancellation can be identified reliably.

---

**packages/mcp-server/src/handlers.ts:2137,2198**

`workspace_dir` is passed directly to the service without validation:

```ts
const workspaceDir = args.workspace_dir as string | undefined;
// ...
const envs = await service.listEnvironments(workspaceDir);
```

Although the extension executes with the user's privileges, MCP tool inputs originate from AI assistants and may include untrusted values. As a defense-in-depth measure, consider validating that `workspace_dir` is an absolute path and does not contain traversal sequences before passing it to the service.

For example:

```ts
const resolved = path.resolve(workspaceDir);
if (workspaceDir.includes('..') || !path.isAbsolute(resolved)) {
    return mcpError({
        code: 'INVALID_PARAM',
        recoverability: 'fail',
        message: 'workspace_dir must be an absolute path without traversal sequences',
    });
}
```

This helps prevent unintended execution against arbitrary directories while preserving the intended functionality.

sudhirverma (approved):

sathyapramod (approved):
Reviewed 
- Security posture is strong.
- Codecov is good.
- Test coverage is thorough for the service layer.

PR #3060, opened 2026-07-27
feat(ls): open PluginDocPanel on go-to-def

sudhirverma:
F12/Ctrl+click on a task module resolves FQCN via the doc cache and notifies the extension host to open the panel instead of jumping to Python source.

Adds go-to-definition for Ansible task modules by resolving module FQCN from cached collections docs (short or already-qualified) and using an LSP→extension notification to open `PluginDocPanel` instead of jumping to Python source. Needed to provide accurate module documentation navigation. Original commit message: Adds go-to-definition support for task modules.

- Enable LSP `definitionProvider` in `AnsibleLanguageService` and implement `onDefinition` delegating to `doDefinition`
- Add a YAML-scoped definition provider that validates cursor position, resolves the module’s FQCN, and triggers `ansible/openPluginDoc`
- Use cached `CollectionsService` to fetch module documentation and derive the definition link
- Register an `ansible/openPluginDoc` notification handler in the extension host to open `PluginDocPanel` with `{ fqcn, pluginType }`
- Add Vitest coverage for definition resolution and notification behavior (including resolution and “no definition” cases)
- Update the SDL C TODO entry to mark the go-to-definition behavior as complete

### DIFF
--- a/.sdlc/todos/complete/add-goto-definition-provider.md
+++ b/.sdlc/todos/complete/add-goto-definition-provider.md
@@ -1,7 +1,7 @@
 ---
 title: Add go-to-definition that opens PluginDocPanel
 created: 2026-05-26
-status: pending
+status: complete
 priority: medium
 scope: ls
 ---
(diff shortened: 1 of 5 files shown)


sathyapramod (approved):
LGTM!

PR #3100, opened 2026-08-13
fix(deps): exclude ansible-core 2.17.x (CVE-2026-11332)

rockygeekz:
## Summary

- Excludes `ansible-core 2.17.*` from the dev dependency range to avoid CVE-2026-11332
- The constraint `>=2.13.0,!=2.17.*` keeps all other supported versions available
- Regenerated `uv.lock` to reflect the updated constraint

## Test plan

- [ ] CI passes with the exclusion in place
- [ ] Verify `uv lock` resolves without pulling any 2.17.x release

- Exclude `ansible-core` `2.17.x` from development dependencies because the EOL release will not receive a fix for CVE-2026-11332. Preserve support for `>=2.13.0` and regenerate `uv.lock` so resolution selects `2.16.19` on Python 3.10 or `2.18+` on Python 3.11+.

Original commit message: “ansible-core 2.17 is end-of-life and will not receive a fix for CVE-2026-11332.”


rockygeekz:
Already merged in main, closing.

PR #3165, opened 2026-09-15
fix(e2e): add retry logic to testHover for flaky CI tests

shvenkat-rh:
Add retry/polling mechanism to testHover() to handle timing-dependent flakiness in e2e tests, particularly withEE.test.ts hover tests for builtin modules.

Changes:
- Add waitForCondition() utility for generic async polling
- Update testHover() with configurable retry logic (default: 5 retries)
- Update withEE.test.ts builtin module tests with aggressive retry options (10 retries, 1000ms delay) to handle EE doc indexing timing
- Add retry options to collection module tests for stability

The builtin module hover tests were failing intermittently (~54-80% of classified failures) because docs may not be immediately available after the docsLibraryReady notification, even though the notification timing is correct.

related: scheduled CI flakiness analysis

- Adds `waitForCondition()` polling and configurable `testHover()` retries to reduce timing-related E2E test failures.
- Configures longer retries for Execution Environment documentation indexing and collection module hover tests.
- Retries hover checks until documentation is available, then reports aggregated failures.

Original commit message: fix(e2e): add retry logic to testHover for flaky CI tests

Co-authored-by: Cursor <cursoragent@cursor.com>


test/e2e/e2e.utils.ts
```diff
@@ -115,6 +115,61 @@ export async function sleep(ms: number): Promise<void> {
...
+  const {
+    timeout = 5000,
+    interval = 200,
+    description = "condition",
+  } = options ?? {};
+  const startTime = Date.now();
+
+  while (Date.now() - startTime < timeout) {
+    try {
+      if (await condition()) {
```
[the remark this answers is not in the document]
shvenkat-rh: @coderabbitai ignore

Won't fix: The current implementation is adequate for the use case. If a 
condition hangs, we want the test to eventually fail rather than silently 
timeout. The interval-based polling is predictable and debuggable.

test/e2e/e2e.utils.ts
```diff
@@ -115,6 +115,61 @@ export async function sleep(ms: number): Promise<void> {
...
+  options?: WaitForConditionOptions,
+): Promise<void> {
+  const {
+    timeout = 5000,
+    interval = 200,
+    description = "condition",
+  } = options ?? {};
+  const startTime = Date.now();
+
+  while (Date.now() - startTime < timeout) {
```
rockygeekz, line 159: The `timeout` option is only checked at the start of each loop iteration. A slow or hung `await condition()` can run past the documented max wait, and `sleep(interval)` can add more time after that. Could you enforce a single deadline (remaining time before each poll, cap the sleep, or race the condition against the remaining budget)?

test/e2e/e2e.utils.ts
```diff
@@ -115,6 +115,61 @@ export async function sleep(ms: number): Promise<void> {
...
+ * // Wait for hover results to be available
+ * await waitForCondition(
+ *   async () => {
+ *     const hovers = await vscode.commands.executeCommand("vscode.executeHoverProvider", uri, pos);
+ *     return hovers.length > 0;
+ *   },
+ *   { timeout: 10000, description: "hover results" }
+ * );
+ */
+export async function waitForCondition(
```
rockygeekz, line 148: `waitForCondition` is exported here but nothing in the PR calls it, while `testHover` implements its own retry loop. Either drop the helper until it is needed or refactor `testHover` to use it so we do not maintain two polling patterns.

test/e2e/e2e.utils.ts
```diff
@@ -115,6 +115,61 @@ export async function sleep(ms: number): Promise<void> {
...
+    description = "condition",
+  } = options ?? {};
+  const startTime = Date.now();
+
+  while (Date.now() - startTime < timeout) {
+    try {
+      if (await condition()) {
+        return;
+      }
+    } catch {
```
rockygeekz, line 164: This bare `catch` drops every error until the global timeout, so callers cannot tell "not ready yet" from a real failure. Consider keeping a `lastError` and including it in the timeout message (same idea as `testHover` below).

test/e2e/e2e.utils.ts
```diff
@@ -310,29 +365,85 @@ export async function testDiagnostics(
...
-  if (actualHover.length && expectedHover.length) {
-    expectedHover.forEach((expectedItem, i) => {
-      const actualItem = actualHover[i];
-      assert.ok(
-        (actualItem.contents[i] as vscode.MarkdownString).value.includes(
-          expectedItem.contents[i].toString(),
-        ),
-      );
-    });
+  for (let attempt = 0; attempt <= retries; attempt++) {
```
rockygeekz, line 401: When `expectedHover` is empty, this loop will retry until hover goes away. That can pass if the provider briefly returns a spurious hover and then clears. For the "should not hover" cases, consider failing immediately on the first non-empty result, or use `retries: 0` for negative assertions only.

test/e2e/e2e.utils.ts
```diff
@@ -310,29 +365,85 @@ export async function testDiagnostics(
   }
 }
 
+/**
+ * Options for testHover retry behavior
+ */
+export interface TestHoverOptions {
+  /** Number of retry attempts (default: 5) */
```
rockygeekz: The JSDoc says 5 retry attempts, but `for (let attempt = 0; attempt <= retries; attempt++)` runs `retries + 1` tries (the failure message already says that). Please align the comment or rename the option (e.g. `maxAttempts`).

test/e2e/hover/withEE.test.ts
```diff
@@ -79,46 +79,80 @@ describe("ee", function () {
...
+            },
+          ],
+          builtinRetryOptions,
+        );
       });
     });
 
     describe("Hover for module name and options present in the EE", function () {
+      // Collection module docs may also benefit from retry logic for stability
+      const collectionRetryOptions = { retries: 5, retryDelay: 500 };
```
rockygeekz, line 117: nit: these values match the new `testHover` defaults, so the constant and fourth argument are redundant unless you want them documented at the call site only.

rockygeekz (changes requested):

test/e2e/e2e.utils.ts
shvenkat-rh: Will fix - changing to "Number of retries after initial attempt" for clarity.

test/e2e/hover/withEE.test.ts
shvenkat-rh: Intentional documentation at the call site. When someone reads the test, they 
see the retry behavior explicitly rather than having to know testHover's 
defaults. I would prefer explicit over implicit 
in tests.

test/e2e/e2e.utils.ts
shvenkat-rh: Fair point on the semantics, but in practice the "should not hover" tests are 
checking for missing modules, the hover provider won't suddenly return a 
spurious result and then clear it. The retry gives time for any lazy loading to 
settle. If this becomes a real issue, we can add a `retries: 0` override, but 
I'd rather not complicate the tests preemptively.

test/e2e/e2e.utils.ts
shvenkat-rh: The semantics are intentional: waitForCondition is "poll until truthy". If the 
condition throws, it's not ready yet, same as returning false. Callers that 
need error details should use testHover-style explicit retry loops. This keeps 
the generic utility simple.

test/e2e/e2e.utils.ts
shvenkat-rh: The timeout is a soft guideline, not a hard requirement. If a condition hangs, that's a 
test bug we want to surface, not mask with aggressive deadline enforcement. 
Adding Promise.race complexity here doesn't improve debuggability , it just 
makes failures harder to trace. The current behavior is intentional.

test/e2e/e2e.utils.ts
shvenkat-rh: This is intentional infrastructure for future e2e test robustness. The plan was 
to add it now alongside testHover so both patterns exist. testHover has 
hover-specific logic (content matching, error messages) that doesn't fit a 
generic utility. Removing it means re-adding it later when other tests need 
polling. We should rather have the utility ready.

test/e2e/e2e.utils.ts
```diff
@@ -310,29 +365,85 @@ export async function testDiagnostics(
...
-          expectedItem.contents[i].toString(),
-        ),
-      );
-    });
+  for (let attempt = 0; attempt <= retries; attempt++) {
+    const actualHover: vscode.Hover[] = await vscode.commands.executeCommand(
+      "vscode.executeHoverProvider",
+      docUri,
+      position,
+    );
```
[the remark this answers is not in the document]
shvenkat-rh: The retry is targeting the specific failure mode we observed: "expected 1 hover, 
got 0" - where the command succeeds but docs aren't indexed yet. If 
executeHoverProvider rejects, that's a real failure (extension not ready, 
language server crashed, etc.) that we want to surface immediately rather than 
mask with retries. Adding this would slow down failure detection for actual bugs.

rockygeekz (approved):

